PR #45, opened 2015-01-02
Implement IO.readString_

pwoolcoc:


pwoolcoc:
would love to add a couple tests, but I will probably need to refactor it a bit so that it can take input from more than just `stdin`. Unless you have a good idea of how to provide input on `stdin` while the test suite is running?

munificent:
Oh, good point. I've got some ideas on how to handle that, so don't sweat writing test for this. If you can make the other changes, I'll land it and handle the tests.

pwoolcoc:
@munificent ok, added the changes you mentioned

munificent:
Looks great! One last thing: would you like to add yourself to the AUTHORS file?

pwoolcoc:
sure, will send a PR shortly

munificent:
This is in now! It took a little bit of effort to get it tested, but that's working now.

It's still a little rough around the edges. We'll eventually want to think more about how it should handle the hard-coded line limit, and whether the returned string should include the trailing line separator or not, but this is a great start.

:metal:

PR #48, opened 2015-01-02
First try at building separate lib + binary

pwoolcoc:
This makes it so a `make release` puts the `wren` binary in your `$PWD`, along with `libwren.a`, so that `wren` can be used more easily from other applications. (Also does similar things for `make debug`)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,3 +1,4 @@
+AR = ar rcu
 # Compiler flags.
 CFLAGS = -std=c99 -Wall -Werror
 # TODO: Add -Wextra.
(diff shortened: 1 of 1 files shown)


munificent:
Oops, closed the wrong bug!

munificent:
Thanks for hacking on this! This has been on my TODO list for a while. You'll have to spoonfeed this to me since I'm a total make novice. By:

> puts the wren binary in your $PWD,

do you mean something different from what it currently does, or just that it does the same thing for the shared lib now too?

pwoolcoc:
No, it does the same thing for the `.a` now. `wren` and `wrend` will still show up in the same place as before.

munificent:
Beautiful! Thank you very much!

This reminds me. I should really get an AUTHORS file going. Would you like to be in it? What name and email?

pwoolcoc:
Sure! Paul Woolcock <paul@woolcock.us>

PR #49, opened 2015-01-02
Fix example of catching and printing an error

skybrian:
The output didn't match the source code.

(Also, it looks like a newline was missing at the end of the file.)


munificent:
Thanks!

PR #51, opened 2015-01-02
Makefile fixes for Linux.

vadz:
The main commit is the first one, as wren doesn't build out of the box for me:
```
% make -j1
cc -c -std=c99 -Wall -Werror -Os -Iinclude -o build/release/wren_compiler.o src/wren_compiler.c
cc1: warnings being treated as errors
In file included from src/wren_compiler.h:5,
                 from src/wren_compiler.c:8:
src/wren_value.h:293: error: declaration does not declare anything
make: *** [build/release/wren_compiler.o] Error 1
```
This is due to using strict `c99` which, at least with g++ 4.4 used above, disables support for unnamed unions
as it's not in C99 (but apparently in C11, according to https://gcc.gnu.org/onlinedocs/gcc/Unnamed-Fields.html).

My fix allows to use `make CC=gcc` to avoid this, but this is still not ideal as it's still broken out of the box. Detecting
gcc from makefile is rather ugly though, so I don't know if you're willing to do it.

The two subsequent ones just clean up the makefile a little.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,9 +2,6 @@
 CFLAGS := -std=c99 -Wall -Werror
 # TODO: Add -Wextra.
 
-DEBUG_CFLAGS := -O0 -DDEBUG
-RELEASE_CFLAGS := -Os
-
 # Files.
 SOURCES := $(wildcard src/*.c)
 HEADERS := $(wildcard src/*.h)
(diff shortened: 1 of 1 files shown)


vadz:
P.S. Travis failure seems to be unrelated to my changes.

munificent:
Yeah the Travis this is my problem. :)

I really like your makefile changes except for the GNU dialect. I want Wren to be strict C99, so I went ahead and made the one anonymous union I know of have a name. Can you merge from master and let me know if that works for you?

vadz:
Yes, it compiles with `-std=c99` now, thanks. I've rebased the branch to remove that commit.

munificent:
Thanks for these changes! I just merged in another change to the Makefile (to add support for building shared libraries) and it's conflicting with this one in nasty ways.

What to you think about keeping the separate `DEBUG_CFLAGS` and `RELEASE_CFLAGS` variables? Or can you merge in the latest changes here and figure out how to get them to play nice together?

vadz:
I won't have time to do it immediately but I could redo the change later if you think it's worthwhile. Personally I like using the target-dependent variables because they're nicely propagated by gmake to all the target dependencies which is less error-prone than doing it manually -- and is also shorter. But it could be argued that separate `{DEBUG,RELEASE}_CFLAGS` is more clear because it's more explicit -- your choice.

FWIW if you want to build shared libraries portably, you really need to use libtool. And then you probably will also want to use automake.

munificent:
> But it could be argued that separate {DEBUG,RELEASE}_CFLAGS is more clear because it's more explicit -- your choice.

Yeah, I think I like being a little more explicit here.

> FWIW if you want to build shared libraries portably, you really need to use libtool. And then you probably will also want to use automake.

I'm trying to keep things as simple as possible and avoid going into metabuild systems. I figure if `ar` and a raw `Makefile` are good enough for Lua, they're probably good enough for Wren.

I think many users will just drop Wren into their project as source anyway.

I'll close this out, but I really do appreciate you putting time into this. Thanks!

PR #55, opened 2015-01-03
Don't try to run an empty line in the REPL.

gsmaverick:
This PR prevents the REPL from trying to interpret an empty line (a behaviour that is similar to most other REPLs).


munificent:
:heart:

PR #56, opened 2015-01-03
Support for hashtag comments.

mauvm:
I like using the hashtag for comments (like in bash) instead of the double slash. Purely personal preference ofcourse.

Additionally this PR allows for the use of shebangs:

```
#!/usr/bin/env wren

IO.print("Hello world!")
```


munificent:
Having support for shebangs is nice, but I'm not crazy about having a third comment form in general. Seems like overkill to me (though I get that some people like them). I'm really hesitant to sacrifice a delicious punctuation character for this. Once it's claimed by something in the grammar, it's gone forever.

One option would be to do what Dart does and allow a `#` shebang, but *only* as the first line of the program. That still makes it available for use elsewhere in the file to mean other things. (Dart uses them for symbols.)

I'm going to close this out because I don't think I want to add `#` overall, but I do really appreciate the patch (and the tests and docs!). If you're interested in another itch to scratch, I can point you to something else that needs hacking on. :)

mauvm:
After implementing both solutions (see #58) I figured this version is a lot shorter/cleaner.

And yes, I'd like to help. Full disclaimer though: I'm not very experienced in (the practical side of) writing code in C, so code review and suggestions (about memory leaks and stuff) are very welcome!

munificent:
> Full disclaimer though: I'm not very experienced in (the practical side of) writing code in C.

That's OK. That's why we have code review. :) I'm still learning a lot too. My knowledge of make and C99 is pretty bare bones right now.

PR #59, opened 2015-01-04
Fixing CRLF problem on Windows

elder-george:
On Windows newline is represented as CRLF ('\r\n').

When file is open in "r' (not "rb") mode, the `\r`s are omitted and `fread` returns number of actual bytes read which might be smaller than file size.

In this case validation in the wren code will fail.

This patch changes the way validation is done to prevent the problem without complicating lexer.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -39,7 +39,8 @@ static char* readFile(const char* path)
 
   // Read the entire file.
   size_t bytesRead = fread(buffer, sizeof(char), fileSize, file);
-  failIf(bytesRead < fileSize, 74, "Could not read file \"%s\".\n", path);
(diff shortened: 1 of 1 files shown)


munificent:
I definitely appreciate not wanting to complicate the lexer, but I think the language grammar probably should specify that `\r` is an ignored whitespace character. That way, if you open a Windows (i.e. `\r\n`) file on a Unix machine, it still does the right thing.

This was a mistake on my part. I think the right thing to do is to `fopen` with `"rb"`. That way the behavior isn't platform-specific. I can fix the lexer if you want or you're welcome to. You'll basically just be handling `\r` around line 788 in wren_compiler.c.

PR #62, opened 2015-01-04
Adding support for '\r' char to lexer.

elder-george:
Another way to fix same issue: opening script file in binary mode and handling `\r` as a new line character (both followed with `\n` as in Windows or standalone as in old Macs - just in case =))


munificent:
Thanks! If you'd like, feel free to also send me a PR to add yourself to the AUTHORS file. :)

PR #63, opened 2015-01-04
Let the user know the arity of a missing method in the error message.

gsmaverick:
Since the arity of a function is expressed with a number of spaces after the function name it can be a little be difficult to quickly see that you used an incorrect number of arguments. This PR updates the error message to include the number of arguments that the user tried to use but wasn't implemented.

### DIFF
--- a/src/wren_vm.c
+++ b/src/wren_vm.c
@@ -24,37 +24,32 @@ static void* defaultReallocate(void* memory, size_t oldSize, size_t newSize)
   return realloc(memory, newSize);
 }
 
-// Returns the canonical name of a symbol with the argument spaces stripped off.
-char* canonicalSymbol(char* symbolName)
-{
-  // Find the end of the text in the given symbol name.
-  int pos = strlen(symbolName) - 1;
-  while (symbolName[pos] == ' ') pos--;
-
-  // Create a new string that contains only the symbol name.
-  char *canonicalizedSymbol = malloc(pos + 1);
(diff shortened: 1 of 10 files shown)


gsmaverick:
@munificent updated this PR to address your feedback. I tried to encapsulate everything inside the `methodNotFound` function but the `RUNTIME_ERROR` macro is defined inside of  `runInterpreter` so I had to keep that where it was.

I think I've covered all my tracks on memory leakage now but let me know if that's not the case (I'm a relative C noob).

munificent:
I tweaked a few things, but this is landed on master now! If you're curious, you can see what I did here: https://github.com/munificent/wren/commit/efabf20fee6cac747e2bb384efa19616b499d98f

Really nice bonus, this patch significantly increased performance:

```
binary_trees - wren            ..........  3577  0.28s  108.12% relative to baseline
delta_blue - wren              ..........  7287  0.14s  105.21% relative to baseline
fib - wren                     ..........  3323  0.30s  112.91% relative to baseline
for - wren                     ..........  8811  0.11s  105.77% relative to baseline
method_call - wren             ..........  5202  0.19s  103.77% relative to baseline
```

I'm guessing it's because the error message strings used to be static local characters arrays right in the main interpret function, and those are gone now so we get a little better cache locality. Woo! :fireworks:

gsmaverick:
Good news on performance :) Those look like some good improvements, I'll have to study them a bit for my next PR.

PR #64, opened 2015-01-04
Adds the ability to use compound variable declarations.

gsmaverick:
This PR enables you to define multiple variables in one statement/line (see the updated test file for the different variations that are supported).

### DIFF
--- a/src/wren_compiler.c
+++ b/src/wren_compiler.c
@@ -2680,21 +2680,26 @@ static void classDefinition(Compiler* compiler)
 
 static void variableDefinition(Compiler* compiler)
 {
-  // TODO: Variable should not be in scope until after initializer.
-  int symbol = declareNamedVariable(compiler);
+  do {
+    ignoreNewlines(compiler);
 
-  // Compile the initializer.
-  if (match(compiler, TOKEN_EQ))
-  {
-    expression(compiler);
-  }
-  else
-  {
-    // Default initialize it to null.
-    null(compiler, false);
-  }
+    // TODO: Variable should not be in scope until after initializer.
+    int symbol = declareNamedVariable(compiler);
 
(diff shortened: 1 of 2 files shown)


munificent:
I like how this looks, but I've shied away from this for now because I worry that it may have some weird corner cases or prevent us from doing other stuff we might want later.

In particular, in [a previous language I hacked on](http://magpie-lang.org/variables.html), you could declare multiple variables in a single statement, but it did so by doing destructuring assignment, like:

``
var a, b = 1, 2
```

This is more powerful than just declaring multiple variables in a single statement, because it pulls the values out of composite objects. I don't know if that will be a good fit for Wren but I'm hesitant to rule it out.

There's also some confusing cases like:

```
var a = 1, b = a // Is this valid?
var a = b, b = 1 // What happens here?
var a, b = 1 // Is a initialized or not?
```

Of course, these are solvable problems. It just makes me wonder if the feature is really worth the complexity. I'm trying *really* hard to keep Wren as minimal as possible.

Maybe it's just me, but the JS style of:

```js
var a = 1,
    b = 2,
    c = 3;
```

Has been particularly appealing to me since it doesn't actually save any space and actually makes it a bit harder to refactor that code. I tend to just do a `var` on each line.

What do you think?

gsmaverick:
Lots of good points which I didn't have in mind when I was tinkering with this feature. It's not clear that the solution I've proposed in this PR is the right solution even if you decide to add some sort of compound variable declaration to wren.

munificent:
> Lots of good points which I didn't have in mind when I was tinkering with this feature.

I'll just reiterate that even though this got closed, your tinkering is totally awesome. It's still a really good exercise for learning your way around the compiler code. :)

PR #71, opened 2015-01-05
Use versioned Python binary

edsrzf:
I'm running Arch Linux, where /usr/bin/python is Python 3.
This solution is suggested by PEP-0394.


munificent:
Closing this out since I took @zeckalpha's related patch instead. :)

PR #72, opened 2015-01-06
Make Python scripts Python 2 and 3 compatible. An alternative to #71.

zeckalpha:
See also: #71.

@edsrzf, does this work for you?

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -3,3 +3,4 @@ If you submit a patch to Wren, please add your name and email address to the
 end of this list.
 
 Robert Nystrom <robert@stuffwithstuff.com>
+Kyle Marek-Spartz <kyle.marek.spartz@gmail.com>
--- a/script/generate_builtins.py
+++ b/script/generate_builtins.py
@@ -1,4 +1,5 @@
 #!/usr/bin/env python
+
 import glob
 import os.path
 import re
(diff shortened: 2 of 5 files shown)


edsrzf:
Yep, this works too. (I actually didn't try `generate_docs.py` since it requires a Markdown package I don't have installed and don't really want to install.)

edsrzf:
I do think this is a bit more fragile, though, since I suspect most people are using Python 2 and they'll forget to test their changes with Python 3. Likewise I'd probably forget to test with Python 2.

Travis could run scripts with both Python 2 and 3, but that's getting a little silly.

zeckalpha:
I thought about writing a helper script to set up a virtualenv and install markdown. Note that docs also has a non-Python dependency: `sass`.

Really though, these should be written in Wren!

munificent:
Yeah, I like the idea of having the scripts work in Python 3 instead of forcing Python 2. Surely we of all people want to live on the bleeding edge of the languages we use, right? :smile:

I don't mind supporting both 2 and 3. If we accidentally break one or the other, hopefully someone will let us know, and it will generally be an easy fix.

At some point, it would be good to get the benchmark runner (`benchmark/run_bench`) supporting Python 3 too.

Sorry about the implicit SASS and Markdown dependencies. The site isn't something I've put much effort into making it easy for others to build and hack on yet (though it is, of course, important!)

PR #73, opened 2015-01-06
Use a flexible array for strings

edsrzf:


munificent:
This is an interesting change. I should have left more details in that TODO, but I actually tried doing this a while back and found it slightly regressed performance for mysterious reasons.

Can you run the benchmarks and see how they look? You'll want to:

```sh
$ git stash # Get back to the previous state.
$ make
$ cd benchmark
$ ./run_bench --generate-baseline # Store timings of the previous code.
$ git stash pop
$ cd ..
$ make
$ cd benchmark
$ ./run_bench -l wren # Benchmark the new code compared to the baseline
```

That will print out what the timings look like relative to the baseline.

edsrzf:
I get:

    binary_trees - wren           ..........  2978  0.34s   96.67% relative to baseline
    delta_blue - wren             ..........  5799  0.17s   97.76% relative to baseline
    fib - wren                    ..........  1863  0.54s  100.09% relative to baseline
    for - wren                    ..........  5347  0.19s   99.84% relative to baseline
    method_call - wren            ..........  3533  0.28s  102.55% relative to baseline

So probably no real change. Then again, these benchmarks aren't really exercising strings.

edsrzf:
Also, is compiling with Visual Studio a goal? I don't think it supports this C99 feature (as well as a number of other C99 features). :(

munificent:
>  Then again, these benchmarks aren't really exercising strings.

True, but it's good to make sure perf doesn't regress. Thanks to processor caching and other stuff, changes deep in the interpreter can have unexpected perf consequences.

> Also, is compiling with Visual Studio a goal? I don't think it supports this C99 feature (as well as a number of other C99 features). :(

Yes, it is, but it will probably take a while and some rethinking on my part to get there.

PR #75, opened 2015-01-06
Remove semi-colons from delta_blue.wren

edsrzf:


munificent:
Ha, oop! How did those get there? :)

PR #78, opened 2015-01-06
Have Travis CI run the benchmark and metrics.

zeckalpha:
This will likely fail since Travis may not have all the interpreters, but I'm making a PR to see how Travis responds.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,4 +2,4 @@ language: c
 compiler:
   - gcc
   - clang
-script: make && make test
+script: make && make test && make benchmark && make metrics
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,4 +2,4 @@ language: c
 compiler:
   - gcc
   - clang
-script: make && make test && make benchmark && make metrics
+script: make && make test && make benchmark
--- a/Makefile
+++ b/Makefile
@@ -75,3 +75,11 @@ watchdocs:
 # Build the docs and copy them to a local "gh-pages" directory.
 gh-pages: docs
 	cp -r build/docs/. gh-pages
+
+# Run the benchmark suite.
+benchmark: wren
+	@pushd benchmark && ./run_bench || popd
+
+# Run the metrics script.
+metrics:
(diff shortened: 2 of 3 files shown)


munificent:
I don't think running the metrics does anything useful. It's basically just reporting information for humans and there's no "wrong" output.

Running the benchmarks to test for perf regression would be *fantastic*. For that, we only need run the Wren ones and we can skip the other languages. What we do need, though, is a baseline file. I don't know much about Travis. Do you know if we can output a file that will be visible to later builds?

zeckalpha:
I'm new to it as well. I'll see if I can figure something out.

zeckalpha:
http://docs.travis-ci.com/user/build-configuration/#Travis-CI-Preserves-No-State-Between-Builds

On the other hand, we could include the baseline in the repository, and update it when merging to master, but that could get annoying fast. Unsure about that. I'm about to add a commit to remove metrics.

zeckalpha:
I could hand-generate baseline.txt from the results that Travis prints out, but then that file would get overwritten all the time. Perhaps there could be a separate travis-baseline.txt or even separate travis-clang/travis-gcc baselines and then run_bench could do some host/compiler detection to determine which baseline to use. I'm not sure I like that solution.

munificent:
Yeah, I think for now this just isn't generally worth worrying about. We can probably find more useful things to put time into. :)

If it's OK with you, I'll just close this out. Let me know if you're scrambling to find something to do and I'll see what I can come up with.

zeckalpha:
This was just an idea to try, and to play around with Travis a bit. I couldn't try it out without opening the PR, and it wasn't ready for a true PR.

I've got some ideas forthcoming; no worries.

Also, I'm going to cherry-pick [this](https://github.com/zeckalpha/wren/commit/19d3208f4d9d2c41a22a196ccbb7c45c6e8366f7) out for a separate PR. I think it is useful for people who don't have all the interpreters installed to be able to run some of the benchmarks.

PR #83, opened 2015-01-08
Benchmark: Alert that an interpreter was not found rather than halting.

zeckalpha:


zeckalpha:
Cherry-picked from #78

munificent:
Thanks!

PR #92, opened 2015-01-09
Fonts weren't working in Firefox due to Mixed Content errors.

zeckalpha:
https://developer.mozilla.org/en-US/docs/Security/MixedContent


zeckalpha:
Good point. I added a commit for that.

munificent:
Swell, thanks!

PR #95, opened 2015-01-09
Segfault when calling super at top-level with no arguments.

zeckalpha:
If enclosingClass is null, we can't proceed with compilation since we don't have its name.


munificent:
Great job finding this bug!

zeckalpha:
Is that what you were thinking?

munificent:
Perfect, thank you!

Issue #86, opened 2015-01-09
Optional stdlib / cookbook


zeckalpha:
I know the point is to keep Wren core small, but perhaps an opt-in standard library with broader features would be useful. This could be implemented as more of a cookbook, including only the needed portions when embedding.

munificent:
Yeah, it's sort of set up that way now. An embedder can optionally not include IO, for example. I agree a larger "standard" library that's written in pure Wren code would be good. Whether it's literally bundled with Wren or is just "third party" and managed by the module system is an open question. :)

PR #96, opened 2015-01-09
Set example updated syntax

zeckalpha:
This is a rebase of #30, and could be part of #86.

As mentioned near the end of #30, I should be adding tests and documentation. I'm not expecting this to be merged until then; I just wanted to get the syntax caught up.


builtin/core.wren
```diff
@@ -41,6 +41,15 @@ class List is Sequence {
     }
     return result
   }
+
+  contains(element) {
+    for (item in this) {
+      if (element == item) {
```
munificent: Style nit: how about making this a one-liner:

```dart
if (element == item) return true
```

I like omitting the `{}` for `if` (in both C and Wren) as long as the entire thing fits on one line and there's no `else` clause.

example/set.wren
```diff
@@ -0,0 +1,110 @@
+class Set {
+  new {
+    _list = []
+    _clean = true
+  }
+
+  new (list) {
```
munificent: Style nit, but I think style nits matter for examples. Can you remove the space before `(`?

example/set.wren
```diff
@@ -0,0 +1,110 @@
...
+  map(f) {
+    return new Set(_list.map(f))
+  }
+
+  where(f) {
+    return new Set(_list.where(f))
+  }
+
+  |(that) {
+    // Union
```
munificent: How about moving this comment (and the ones below) to above the method declaration so they're more like doc comments?

example/set.wren
```diff
@@ -0,0 +1,110 @@
...
+      _list.where { |element|
+        return that.contains(element)
+      } + that.where { |element|
+        return _list.contains(element)
+      })
+  }
+
+  -(that) {
+    // Set minus
+    return new Set(
```
munificent: You could fit this as a one liner:

```dart
  -(that) { new Set(_list.where {|element| !that.contains(element) }) }
```

It is a bit dense though.

example/set.wren
```diff
@@ -0,0 +1,110 @@
...
+  }
+
+  &(that) {
+    // Intersection
+    return new Set(
+      _list.where { |element|
+        return that.contains(element)
+      } + that.where { |element|
+        return _list.contains(element)
+      })
```
munificent: How do you feel about styling this like:

```dart
    return new Set(_list.where {|element|
      return that.contains(element)
    } + that.where {|element|
      return _list.contains(element)
    })
```

munificent:
Oh, yeah! I forgot all about this. I tend to go through periods where I cycle between side projects and I put Wren down for a while.

I like this example. Added a few comments. As you noted, some documentation and tests for `contains` will be good, but otherwise this looks rad.

Once we get maps in the language, I think that will imply more comprehensive support for hash codes. At that point, it will probably make sense to revise this. But until then, this is groovy.

example/set.wren
zeckalpha: Interesting. I've seen doc comments both internal and external to the method. I've got mostly a Python background, where doc comments are internal. This allows the doc comments to be accessed as an attribute of the object. Something to think about. I'll move them for now.

example/set.wren
zeckalpha: Actually, I think I can just say `where {|element| !that.contains(element) }` since I defined where for Set!

example/set.wren
zeckalpha: I've got this right now:

~~~
    return where { |element|
      return that.contains(element)
    } + that.where { |element|
      return contains(element)
    }
~~~

example/set.wren
munificent: Yeah, I did [something similar](http://magpie-lang.org/program-structure.html#doc-comments) for my last language.

I think I had convinced myself that other core lib methods were documented on the outside, but now I see that they aren't documented at all (well, except for on the site).

If we get to the point where we start hacking on an automated doc generator, we can worry more seriously about it then.

example/set.wren
munificent: :grinning:

example/set.wren
munificent: :heart:

zeckalpha:
I see you merged this, but I was still working on it! I'll have a follow-up with some tweaks soon.

munificent:
Oops! Feel free to send me your tweaks. I was maybe a little overeager to get to zero open pull requests. :)

PR #97, opened 2015-01-09
Add waffle.io badge / link to kanban board.

zeckalpha:
I'm not sure what your thoughts are on Kanban, but it is another GitHub integration.


munificent:
It doesn't seem that interesting to me, but maybe I'm missing something. I'll close this out speculatively, but let me know if you think it's worth doing and why. :)

PR #99, opened 2015-01-11
Adds some common methods to strings.

gsmaverick:
Adds the following new functions on strings: 
  * `capitalize` - returns the capitalized string
  * `endsWith(substring)` - checks if the string ends with *substring*
  * `indexOf(substring)` - returns the index of the first occurrence of *substring*
  * `startsWith(substring)` - checks if the string starts with *substring*
  * `toLowerCase` - returns the entire string in lower case
  * `toUpperCase` - returns the entire string in upper case

I wanted to get this PR opened for discussion and once you're happy with the names and functionality I'll add documentation around each of these functions.

### DIFF
--- a/src/wren_core.c
+++ b/src/wren_core.c
@@ -930,6 +930,47 @@ DEF_NATIVE(string_count)
   RETURN_NUM(count);
 }
 
+DEF_NATIVE(string_endsWith)
+{
+  if (!validateString(vm, args, 1, "Argument")) return PRIM_ERROR;
+
+  ObjString* string = AS_STRING(args[0]);
+  ObjString* search = AS_STRING(args[1]);
+
+  // Corner case, if the search string is longer than return false right away.
+  if (search->length > string->length) RETURN_FALSE;
+
+  int result = memcmp(string->value + string->length - search->length,
+                       search->value, search->length);
+
+  RETURN_BOOL(result == 0);
+}
+
+DEF_NATIVE(string_indexOf)
+{
+  if (!validateString(vm, args, 1, "Argument")) return PRIM_ERROR;
+
+  ObjString* string = AS_STRING(args[0]);
+  ObjString* search = AS_STRING(args[1]);
+
+  char* firstOccurrence = strstr(string->value, search->value);
+
+  RETURN_NUM(firstOccurrence ? (firstOccurrence - string->value) : -1);
+}
+
+DEF_NATIVE(string_startsWith)
+{
+  if (!validateString(vm, args, 1, "Argument")) return PRIM_ERROR;
+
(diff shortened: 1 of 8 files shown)


edsrzf:
I feel a little uncomfortable with the methods that operate on character case. They seem like they should work with Unicode characters, but then we have to pull in a bunch of Unicode data that would bloat the implementation. I think I would just leave them out, personally.

src/wren_core.c
```diff
@@ -930,11 +941,73 @@ DEF_NATIVE(string_count)
...
+{
+  if (!validateString(vm, args, 1, "Argument")) return PRIM_ERROR;
+
+  ObjString* string = AS_STRING(args[0]);
+  ObjString* search = AS_STRING(args[1]);
+
+  // Corner case, if the search string is longer than return false right away.
+  if (search->length > string->length) RETURN_FALSE;
+
+  int result = strncmp(string->value + string->length - search->length,
```
edsrzf: You could use `memcmp` here since the string lengths are known.

src/wren_core.c
```diff
@@ -930,11 +941,73 @@ DEF_NATIVE(string_count)
...
+}
+
+DEF_NATIVE(string_startsWith)
+{
+  if (!validateString(vm, args, 1, "Argument")) return PRIM_ERROR;
+
+  ObjString* string = AS_STRING(args[0]);
+  ObjString* search = AS_STRING(args[1]);
+
+  RETURN_BOOL(strncmp(string->value, search->value, search->length) == 0);
```
edsrzf: You could use `memcmp` here too. You also probably want to avoid the function call when the length of the search string is greater than the length of the string being searched, like you did in `endsWith`.

munificent:
I agree with @edsrzf. Case-conversion methods are handy, but bring in the complexity of Unicode. To keep Wren small, my intention is to support *storing* strings in UTF-8 internally so that any Unicode string can be handled by Wren, but to not implement any Unicode-specific *operations* so that Wren users don't have to bring in a giant Unicode library whenever they use the language.

The other three look great, and the tests are swell. `lastIndexOf` might be nice too at some point, but it definitely doesn't have to be in this PR. Small patches are always nice. :)

gsmaverick:
Good points about the casing operations with regards to how they interact with Unicode so I will take them out of this PR. That being said I feel like these are important operations to have on strings even if they only operate in the ASCII region (a la [Ruby's approach](http://www.ruby-doc.org/core-2.2.0/String.html#method-i-capitalize)). I'll probably open an separate issue about this so we can have a focused discussion there rather than in the comments of this PR.

gsmaverick:
@munificent pr updated.

munificent:
Awesome. This rocks!

PR #103, opened 2015-01-12
Allow better debugging

pwoolcoc:
This will ensure that we have better output when running `wren` under
{g,ll}db.


munificent:
:+1:

PR #105, opened 2015-01-12
Missing semicolons.

zeckalpha:
I'm unsure about the one inside the macro, but the other ones were making my syntax highlighting break.

### DIFF
--- a/src/wren_utils.h
+++ b/src/wren_utils.h
@@ -18,7 +18,7 @@
     } name##Buffer; \
     void wren##name##BufferInit(WrenVM* vm, name##Buffer* buffer); \
     void wren##name##BufferClear(WrenVM* vm, name##Buffer* buffer); \
-    void wren##name##BufferWrite(WrenVM* vm, name##Buffer* buffer, type data);
+    void wren##name##BufferWrite(WrenVM* vm, name##Buffer* buffer, type data)
 
 // This should be used once for each type instantiation, somewhere in a .c file.
 #define DEFINE_BUFFER(name, type) \
--- a/src/wren_utils.h
+++ b/src/wren_utils.h
@@ -18,7 +18,7 @@
     } name##Buffer; \
     void wren##name##BufferInit(WrenVM* vm, name##Buffer* buffer); \
     void wren##name##BufferClear(WrenVM* vm, name##Buffer* buffer); \
-    void wren##name##BufferWrite(WrenVM* vm, name##Buffer* buffer, type data)
+    void wren##name##BufferWrite(WrenVM* vm, name##Buffer* buffer, type data);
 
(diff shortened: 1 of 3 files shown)


src/wren_utils.h
```diff
@@ -18,7 +18,7 @@
     } name##Buffer; \
     void wren##name##BufferInit(WrenVM* vm, name##Buffer* buffer); \
     void wren##name##BufferClear(WrenVM* vm, name##Buffer* buffer); \
-    void wren##name##BufferWrite(WrenVM* vm, name##Buffer* buffer, type data)
+    void wren##name##BufferWrite(WrenVM* vm, name##Buffer* buffer, type data);
```
zeckalpha: I don't think I want this one. Removing it shortly.

zeckalpha:
Looking around a bit, I removed the one from the inside of the macro.

munificent:
I'm actually a bit surprised it compiles with those extra semicolons, but it seems to, so I guess they're OK.

zeckalpha:
Are they extra? It could have been a bug in my editor's C highlighting.

These examples suggest including them:

https://en.wikipedia.org/wiki/C_preprocessor#Token_concatenation
https://gcc.gnu.org/onlinedocs/gcc-2.95.3/cpp_1.html#SEC23

munificent:
Yeah, they are. When the preprocessor expands:

```c
DEFINE_BUFFER(String, char*);
```

You end up with:

```c

void wrenStringBufferInit(WrenVM* vm, StringBuffer* buffer)
{
  buffer->data = NULL;
  buffer->capacity = 0;
  buffer->count = 0;
}

void wrenStringBufferClear(WrenVM* vm, StringBuffer* buffer)
{
  wrenReallocate(vm, buffer->data, 0, 0);
  wrenStringBufferInit(vm, buffer);
}

void wrenStringBufferWrite(WrenVM* vm, StringBuffer* buffer, type data)
{
  if (buffer->capacity < buffer->count + 1)
  {
    int capacity = buffer->capacity == 0 ? 8 : buffer->capacity * 2;
    buffer->data = wrenReallocate(vm, buffer->data,
        buffer->capacity * sizeof(type), capacity * sizeof(type));
    buffer->capacity = capacity;
  }
  buffer->data[buffer->count] = data;
  buffer->count++;
};
```

Note the `;` at the end.

zeckalpha:
That makes sense. It looks like this might be related to: atom/language-c/issues/23. Feel free to revert this and I'll take this up there.

zeckalpha:
Ha, GitHub's highlighting is funky, too: https://github.com/atom/language-c/issues/23#issuecomment-69855901

I suppose that makes sense; language-c is probably based on GitHub's C support.

Issue #87, opened 2015-01-09
File IO


zeckalpha:
I'd like to be able to read and write files.

munificent:
Per our awesome large discussion, I'm going to close this out.

zeckalpha:
I closed out a few others and migrated them.

munificent:
Sounds good!

PR #108, opened 2015-01-14
File IO: IO.readFile(filename)

zeckalpha:
The start of #87.

### DIFF
--- a/src/wren_io.c
+++ b/src/wren_io.c
@@ -7,7 +7,8 @@
 #include <time.h>
 
 // TODO: This is an arbitrary limit Do something smarter.
-#define MAX_READ_LEN 1024
+#define MAX_LINE_LEN 1024
+#define MAX_FILE_LEN 1024 * 1024
 
 // This string literal is generated automatically from io.wren. Do not edit.
 static const char* libSource =
@@ -117,8 +118,8 @@ static void ioWriteString(WrenVM* vm)
 
 static void ioRead(WrenVM* vm)
 {
-  char buffer[MAX_READ_LEN];
-  char* result = fgets(buffer, MAX_READ_LEN, stdin);
+  char buffer[MAX_LINE_LEN];
+  char* result = fgets(buffer, sizeof buffer, stdin);
 
   if (result == NULL) {
     // TODO: handle error.
@@ -132,12 +133,31 @@ static void ioClock(WrenVM* vm)
   wrenReturnDouble(vm, (double)clock() / CLOCKS_PER_SEC);
 }
 
+static void ioReadFile(WrenVM* vm)
+{
+  const char* filename = wrenGetArgumentString(vm, 1);
+
+  char file_buffer[MAX_FILE_LEN];
+  char line_buffer[MAX_LINE_LEN];
+
+  FILE *fp = fopen(filename, "r");
+
+  while (fgets(line_buffer, sizeof line_buffer, fp) != NULL) {
+    strcat(file_buffer, line_buffer);
+  }
+
+  fclose(fp);
+
+  wrenReturnString(vm, file_buffer, (int)strlen(file_buffer));
+}
+
 void wrenLoadIOLibrary(WrenVM* vm)
 {
   wrenInterpret(vm, "", libSource);
   wrenDefineStaticMethod(vm, "IO", "writeString_", 1, ioWriteString);
   wrenDefineStaticMethod(vm, "IO", "clock", 0, ioClock);
   wrenDefineStaticMethod(vm, "IO", "read", 0, ioRead);
+  wrenDefineStaticMethod(vm, "IO", "readFile", 1, ioReadFile);
 }
 
 #endif
--- a/test/io/read_file.wren
+++ b/test/io/read_file.wren
@@ -1,0 +1,1 @@
+IO.write(IO.readFile("example/hello.wren")) // expect: IO.print("Hello, world!")


gsmaverick:
What about having a File class instead of sticking it on IO as a static method? I was playing around with some basic implementations of #87 as well and it seemed a nicer API to have a separate class for dealing with files.

munificent:
Giant braindump!

This is a small step down a very long road. Before we start incrementally adding stuff like this, we should think about the long term plan.

I look at Wren mostly as an embedded scripting language. That means it has a pretty minimal standard library and, in particular, the standard library is mostly concerned with interacting with Wren objects and doesn't reach out to the outside world.

At the same time, Wren-the-project also includes a little standalone command line shell for the VM. It's like Lua's little command-line runner where it basically lets you play with the language but doesn't do that much else. It's also certainly handy for the tests and benchmarks!

Then there's things like node.js and [luvit](https://luvit.io/) which take an embeddable language (JS and Lua respectively) and wrap them in a command-line shell that provides a full panoply of IO operations.

All three of these levels (embeddable, thin shell, thick shell) are, I think, useful in their own right and worth doing. What it's important to me is making sure that one doesn't interfere with the others.

Embedders very may not want to suck in file IO, so they need to be able to opt out of that. For the thin embedder, we'll want to keep the file support (if any) minimal and simple. But a more full-featured command line wrapper may want more detailed file support. Is there a way we can rationalize those two together? It would be confusing if there were two different file APIs.

One option, of course, is to ditch the thin embedder. Just have a pure embeddable language with a very minimal standard library, and a kickass command-line app that uses it and offers a full slew of IO libraries. That's certainly appealing, but it's a *ton* of work, even with all of you also cranking on it.

I also think that until we have a solid module story, it will be really difficult to grow its libraries. We don't want to dump everything in the global namespace.

Another option is to create "node.wren" as a separate repo. It would use Wren as embedded scripting language and have its own shell around it and provide whatever IO libraries it wants. This would both exercise the embedding API (which is a super high priority for me) and make sure I'm not the bottleneck for getting changes landed.

At some point, I would be happy to have it all part of one monolithic "Wren project", but maybe treating it separately would let us all move quickly on the things we're excited about.

What do you think?

zeckalpha:
Yeah, most of this was to get us thinking about that. I'm going to digest @munificent's comments and reply later.

Regarding File as its own class, I agree, with one slight caveat. I like IO systems that maximize code reuse across streams/handles/files/sockets/etc. This makes going from a stdin/stdout implementation to a server just swapping out the handles. Here's a presentation I did demonstrating that in Haskell: https://github.com/HaskellMN/haskell-mn-presentations/tree/master/2014/09/echo-server

To do something like this in wren, IO could have methods with a default handle argument to stdin/stdout, but optionally passing in handles to other things, such as files.

Here's what I'm envisioning (note the block passed into File, with an implicit file close).

~~~ dart
File.open(filename) { |file|
  var contents = IO.read(file)
  IO.write(contents)  // to stdout
  IO.write(contents, file) // to file
}
~~~

zeckalpha:
@munificent, thinking a bit more about dividing it up, but along different lines...

- **wren**: this repo; the embeddable portion, minimal built-ins
- **wren-cookbook**: a new repo with stuff like File IO (maybe all IO?) available with compile flags to enable and disable support for individual features
- **wren-repl**: a thin wrapper around wren-cookbook written in wren to do about what you describe as node.wren

What do you think?

gsmaverick:
@zeckalpha 
1) What do you think about building the IO related functionality on top of the idea of streams (inspired in part by the Node implementation of streams)? I think streams are a pretty powerful concept that could see use in IO but they can also be used for other use cases.

2) I like where that idea is headed although I would envision there to be just two parts: wren and then wren-repl which would be similar to a distribution of Python or Ruby. I feel as though the REPL is a standard thing that should come with a language (unless it's embedded of course).

zeckalpha:
I meant C streams like stdin, stdout. I don't know much about what node
does, but I'd imagine wren's fibers and blocks might be more along those
lines. Unless they are streams as in lazy lists?

munificent:
> note the block passed into File, with an implicit file close

I absolutely love this. "Scoped" code like this for managing resources is one of my main motivations for adding block syntax to the language. Love it.

But, given that, I wouldn't lean on static methods on `IO`. We have objects, we should use them. :) I'd do something like:

```dart
File.open(filename) {|file|
  // This is ambiguous with IO.read(prompt), but perhaps that should be a different method.
  var contents = file.read
  IO.write(contents) // to stdout
  file.write(contents)
}
```

Re **wren-cookbook** and **wren-repl**: I'm not sure I understand how the separation between will work. Are you thinking of the cookbook as basically a collection of libraries written in C + Wren but with no application shell to load it all together?

Do we get a lot of value from separating them?

> 1) What do you think about building the IO related functionality on top of the idea of streams (inspired in part by the Node implementation of streams)?

For what it's worth, I actively dislike streams. Wren has fibers so that we don't have to use asynchrony. We can writing blocking code and let the scheduler handle suspending fibers as needed. In other words, Wren follows Go's model, not node's.

When I was working on Magpie (which also has fibers, but a bit more similar to Goroutines than Wren's fibers are), its core libraries sat on top of libuv and it was a total match made in heaven. Getting all of the speed, concurrency, and portability of libuv + the ease of use of blocking-style code. It's *amazing*.

If we're going to slap some real IO libraries in Wren, I really really hope that's the direction we go in. You can see some of the relevant code [here](https://github.com/munificent/magpie/blob/master/src/VM/Scheduler.cpp). It was surprisingly straightforward to get it hooked up, and I'd be happy to hack on this for Wren.

zeckalpha:
> Are you thinking of the cookbook as basically a collection of libraries written in C + Wren but with no application shell to load it all together?

Yes. The libraries would be selectively compilable and embeddable where ever Wren is embedded. If an embedder would like File IO, they enable that flag. Would all embedders who want, File IO, more math functionality, etc., want a repl, too?

munificent:
> The libraries would be selectively compilable and embeddable where ever Wren is embedded.

Ah, I think I understand that better.

I could be wrong, but my intuition is that libraries will fall into two buckets:

1. Isolated ones that can mostly be implemented in pure Wren. These are the kind of thing you'd see on CPAN, RubyGems, etc. Just, you know, *packages*, where they may as well be third-party.

2. Ones that have a decent amount of native code, close ties to the scheduler/event loop, do IO, and are pretty intertwined. I think these are basically what node *is* and are part of the built-in standard library in other languages. For these, I think it probably makes the most sense to have them as one monolithic "platform" layer. That's what I have in mind when I say "node.wren".

> Would all embedders who want, File IO, more math functionality, etc., want a repl, too?

Everyone wants a REPL! :)

zeckalpha:
I'm not sold on the name, but otherwise that sounds great.

munificent:
Oh, I'm definitely not saying we should actually call it "node.wren", just that that's the idea: Wren-the-language + libuv-io-libraries. For a real name, maybe something like "wren.io"?

zeckalpha:
Sure. Do you want to make a GitHub organization for wren, and wren-io can be another repo in there? I would feel weird making my own wren-io repo, but I do have a few ideas for how to get started.

munificent:
I feel like an organization is getting a bit ahead of ourselves. I formally grant you the freedom to make a wren-io repo. :)

And, just to be clear here, I'm not asking you to do this because I think it's less important or want to push it off to the side or don't want you bugging me or anything like that. I think something like wren-io is *super awesome* and I'm *pumped* that you're keen on it. I just don't want you having to send pull requests to me to slow you down. I look forward to sending *you* pull requests on it. :)

Do it! It's going to be awesome! I just have a few suggestions since I went down a similar path with Magpie:

1. Do use libuv. It's *fantastic*.
2. Try not to use gyp (the build/project system libuv uses). It's an undocumented pain in the ass.
3. Do use fibers in Wren to make the IO *APIs* synchronous while the underlying behavior is async. I can give you help here if you want, since I did this for magpie.

In the meantime, this will let me focus on making Wren itself a great language for building this platform on. I'll try to get a module system figured out, flesh out the embedding API, etc.

zeckalpha:
Done. zeckalpha/wren-io

Nothing there yet. I'll add some stuff later today, hopefully.

munificent:
I'll be watching it!

PR #109, opened 2015-01-14
Adds `forall` method on Sequences.

gsmaverick:


munificent:
Very nice!

gsmaverick:
I added the two tests you requested.

On the truthiness-negation I changed that line to `if (!f.call(element)) return false` in which case if you return a non-bool object from the `predicate` you get a runtime error. I'll open a separate PR to tackle moving the negation operator up to the object instead, sound good?

munificent:
These look great! I had one more comment that I *thought* I typed up but now don't see. Here it is:

When I first saw this PR, I intuited that `forall` was like `forEach` in JS, et. al. The "for" implies that to me. How do you feel about calling this `all` instead?

gsmaverick:
Good point, what do you think about using `every`?

gsmaverick:
Actually I take that back, looks like a number of languages already use `all`. Will update.

gsmaverick:
@munificent Renamed should be good to go now.

munificent:
Awesome, thanks!

doc/site/core-library.markdown
```diff
@@ -313,6 +313,10 @@ Removes all items from the list.
 
 The number of items in the list.
 
+### **forall(predicate)**
```
lluchs: You forgot to rename here.

PR #112, opened 2015-01-15
Add reduce function on Sequence

lluchs:


builtin/core.wren
```diff
@@ -21,6 +21,16 @@ class Sequence {
...
   }
+
+  reduce(acc, f) {
+    for (element in this) {
+      acc = f.call(acc, element)
+    }
+    return acc
+  }
+
+  reduce(f) { this[1..-1].reduce(this[0], f) }
```
zeckalpha: Do all sequences have indexing? Sure, right now `List` is the only implementing class, but...

zeckalpha:
:+1:

builtin/core.wren
lluchs: I considered putting it down into the List class, but that felt a bit wrong as well. I guess it could be implemented without the indexing, but I don't think it's worth the complexity.

munificent: > Do all sequences have indexing?

No, I don't think `Sequence` should assume that.

> Sure, right now List is the only implementing class, but...

Range does too. Maybe String?

gsmaverick: If we are going to keep the single arity `reduce` then we should swap the order of arguments in the 2-arity `reduce`.

test/list/reduce_single_item.wren
```diff
@@ -0,0 +1 @@
+[1].reduce {|a, b| a } // expect runtime error: Range start out of bounds.
```
gsmaverick: My expectation would be that this returns 1 instead of throwing a runtime error. I think it should only throw an error if the list is empty and the single arity `reduce` is called.

builtin/core.wren
munificent: >  we should swap the order of arguments in the 2-arity reduce.

This is an interesting style point. In other dynamic languages that don't support overloading, I think there's a strong convention to have extra arguments only be added to the end. That makes it easier to describe and name the earlier parameters since they have a consistent interpretation in the different simulated overloads.

My experience from statically typed languages that actually support overloading is that it's less important there. Of course, you don't want to just shuffle parameters arbitrarily, but I think you can get away with things like:

```
random(max) { ... }
random(min, max) { ... }
```

I think Wren is closer to a statically-typed language here since we do have real overloads by arity. My feeling is that the accumulator seed naturally comes before the function in `reduce`. Putting the function last *certainly* makes it play better with block args. Actually, now that I think about it, that alone is enough justification to me to have the fn at the end in both arities.

test/list/reduce_single_item.wren
munificent: :+1:

test/list/reduce.wren
```diff
@@ -0,0 +1,9 @@
+var a = [1, 4, 2, 1, 5]
+var max = new Fn {|a, b| a > b ? a : b }
+var sum = new Fn {|a, b| a + b }
```
munificent: A few more things to test:

1. Test that it works on a range too.
2. Test an empty sequence causes a runtime error. (Put this in a separate test, please.)
3. It would be good to have a test that demonstrates the order that the fold happens. Maybe concatenate strings in the function?

I know it's a lot of tests (in addition to the great ones you already did), but stability is our lifeblood. :)

munificent:
Lots of discussion, but I really like this and look forward to pulling it in. :)

builtin/core.wren
gsmaverick: Good point, didn't think about how much nicer it would be with the block syntax. 

I've just recently started working with a static language professionally (Scala) so I'm still trying to undo some of the damage that JS did to me! After your comment I poked around at the APIs in Scala and saw that they also put the seed/acc value in the first place, if used.

test/list/reduce_single_item.wren
lluchs: You're right, not sure what I was thinking :)

builtin/core.wren
lluchs: I initially had the accumulator behind the function, like in Haskell: `foldl :: (a -> b -> a) -> a -> [b] -> a` (i.e. first the function, then the accumulator, then the list), but not being able to use a block argument would be bad.

@munificent: Your version of reduce isn't as bad as I feared, I wasn't aware of the underlying iterator methods :)

test/list/reduce.wren
lluchs: >It would be good to have a test that demonstrates the order that the fold happens.

I wasn't aware right-to-left folds are a thing. In Haskell, both fold variants fold left-to-right (as this is the only sensible thing to do with lists):
```haskell
Prelude> foldl (++) "" ["W", "o", "r", "l", "d"]
"World"
Prelude> foldr (++) "" ["W", "o", "r", "l", "d"]
"World"
```
in contrast to JS:
```javascript
["W", "o", "r", "l", "d"].reduce(function(a, b) { return a + b })
"World"
["W", "o", "r", "l", "d"].reduceRight(function(a, b) { return a + b })
"dlroW"
```
I guess right-associative folds aren't useful outside of lazy languages.

munificent:
Awesome!

test/list/reduce.wren
zeckalpha: No, foldr is useful, it just does something a little different. The trick is to view lists as trees rather than as lists. One example of an function that would get different results would be exponentiation. Here's the diagrams that I always have to look at:

https://en.wikipedia.org/wiki/Fold_%28higher-order_function%29#Folds_as_structural_transformations

PR #113, opened 2015-01-16
Fix all docs

gsmaverick:


munificent:
About time you added yourself! :)

zeckalpha:
The title of this PR confused me at first... then I got it.

Actually, given some of the additions to sequence and lists, perhaps the docs need yet another reorg. Should everything in `core-library.markdown` be there, or should they be broken out into their own files like Lists and Maps?

munificent:
They should be broken out into their own files. I put some time into that a while ago, but didn't get it to a point that I liked. I'll probably take another stab at it soon. Until then, at least having docs—even if on one page—is still good.

PR #114, opened 2015-01-16
POC for `methods` call on all objects.

gsmaverick:
This is my first stab at a `methods` function on all objects which returns a list of methods that you can call on that object. It has one big blocker in that it returns operators and behind the scenes methods:

```dart
  class Foo {}
  var foo = new Foo
  IO.print(foo.methods) 
  // current output: [== , != , methods, new, toString, type,  instantiate]
  // correct output: [methods, toString, type]
```

Now it appears those symbols are always mapped to the same spot in the symbol table so we could just filter out symbols by name or their index in the symbol table. I was hoping there would be some more clever/correct(?) ideas out there.

### DIFF
--- a/src/wren_core.c
+++ b/src/wren_core.c
@@ -818,6 +818,50 @@ DEF_NATIVE(object_bangeq)
   RETURN_BOOL(!wrenValuesEqual(args[0], args[1]));
 }
 
+DEF_NATIVE(object_methods)
+{
+
+  ObjClass* classObj;
+
+  // For booleans, null and numbers we may be dealing with the primitive values
+  // instead of an object instance so we have to get the class that implements
+  // them directly.
+  if (IS_BOOL(args[0])) classObj = vm->boolClass;
+  else if (IS_NULL(args[0])) classObj = vm->nullClass;
+  else if (IS_NUM(args[0])) classObj = vm->numClass;
+  else
+  {
+    // For every other class simply pull the class object off of the instance.
+    ObjInstance* objInstance = AS_INSTANCE(args[0]);
+    classObj = objInstance->obj.classObj;
+  }
+
+  ObjList* methodList = wrenNewList(vm, 0);
+
+  for (int i = 0; i < classObj->methods.count; i++)
+  {
+    Method* method = &classObj->methods.data[i];
+    switch (method->type)
+    {
+      case METHOD_BLOCK:
+      case METHOD_PRIMITIVE:
+      case METHOD_FOREIGN:
+      {
+        // Retrieve the method name from the global symbol table.
+        char* symbolName = vm->methodNames.data[i];
+
+        Value methodName = wrenNewString(vm, symbolName, strlen(symbolName));
+        wrenListAdd(vm, methodList, methodName);
+        break;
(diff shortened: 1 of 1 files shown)


munificent:
Pretty neat! I think most of those methods should show up in the output, though `new` and ` instantiate` are a bit weird. But the operators at least *are* valid (inherited) methods on `Foo`.

Have you heard of Gilad Bracha's idea of [mirrors](http://bracha.org/mirrors.pdf) (pdf)? The TL;DR is that reflective meta-operations should be split out from normal operations. For example:

```dart
class Bird {
  tweet { ... }
  poopOnStatue { ... }
}

var bird = new Bird
bird.tweet
bird.poopOnStatue
```

The above makes sense. But `bird.methods` seems strange: birds don't know anything about "methods", right? Gilad's idea is to move that to a separate API.

I don't necessarily know if I want something *exactly* like his idea of mirrors, but I do think reflective APIs should be hung off of a different set of classes. So, instead of `someObject.methods`, more like:

```dart
Meta.reflect(someObject).methods
```

...or something in that vein.

Metaprogramming in general is a bit tricky. I think it's useful, of course, but it's *really* easy to paint yourself into a corner implementation-wise when you start exposing reflective APIs.

For example, Wren's approach to local variables is very simple, takes little memory, and is very fast. They just reduce to stack offsets at compile time and we discard everything else. But, if we tried to support Pythons `locals()` function, we would have to keep track of all of that data, which adds a lot of complexity.

I'm not opposed to reflection in general, but I'm hoping to procrastinate on it until real compelling needs appear so that we have as much freedom in the implementation as possible for now.

Give the last bit, I'll close this PR, but I do think it's really cool that you did this. :)

gsmaverick:
I quickly scanned that paper but I'll certainly dig into it more because I've never come across that idea before.

To give background on this PR, I was playing around with a (unit) testing framework and trying out some different paradigms. I really like the Python [unittest](https://docs.python.org/3/library/unittest.html) way of writing tests; where you define a sequence of function prepended with `test_` and those get run automatically by the test runner; as opposed to the Jasmine/mocha way of using nested `describe` and `it` blocks to write your tests.

Now I don't think that's quite the compelling need you're looking for but it does make writing elegant testing tools more difficult. I think this meta programming/reflection stuff is interesting but I'm happy to help out on other bugs/features that you see as higher priority.

munificent:
I've tried both styles and I personally like the Jasmine style, though I see arguments for both. For what it's worth, I redesigned the canonical unittest test library for Dart and I migrated it specifically from a jUnit to a Jasmine style, in large part because Dart at the time also didn't have much in the way of reflection.

I ended up liking the result: it made it possible to define tests imperatively, in a loop, etc.

> I think this meta programming/reflection stuff is interesting but I'm happy to help out on other bugs/features that you see as higher priority.

Any kind of tinkering is useful. Just trying stuff and seeing what breaks definitely helps.

I have a lot of stuff that exists purely in my head (or in TODOs in the code), and I'll try to dump some of that into the bug tracker so it's easier to see what needs to be done.

The stuff you've been doing on the core libraries is great. :)

PR #118, opened 2015-01-17
Feature wren return bool

hazbo:
First off, Wren is awesome! So thanks.

Also wasn't sure if this was something you wanted to implement or not. It's useful for a project I'm using Wren with so figured I'd send it back.


munificent:
Awesome, thank you! I definitely want this in there, I just haven't put as much time into the API as I want to yet.

PR #123, opened 2015-01-19
Adds range subscripting for strings.

gsmaverick:


munificent:
Very nice! I flatted out the struct into a couple of output params, but otherwise I really like how you reused the code between lists and strings.

This does dig us a bit deeper into the "not handling UTF-8" correctly hole, but that's OK for now.

PR #124, opened 2015-01-19
Feature wren get argument bool

hazbo:
Getting bool arguments for methods defined with `wrenDefineMethod`.

### DIFF
--- a/include/wren.h
+++ b/include/wren.h
@@ -129,6 +129,9 @@ double wrenGetArgumentDouble(WrenVM* vm, int index);
 // function returns, since the garbage collector may reclaim it.
 const char* wrenGetArgumentString(WrenVM* vm, int index);
 
+// Reads a boolean argument for a foreign call. This must only be called within
+// a function provided to [wrenDefineMethod]. Retrieves the argument at [index]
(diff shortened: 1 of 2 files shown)


hazbo:
I've also noticed, in regards to this pull request, that there may be a problem with either the current API or the comments.

`Retrieves the argument at [index]
// which ranges from 0 to the number of parameters the method expects - 1.`

As far as I can see when I've been using this, the range begins at 1 as appose to 0. Ideally it would start at 0, but then that's changing the API. So, I'm not sure what's best to do here.
Just thought I'd point that out.

munificent:
Awesome, thanks!

> As far as I can see when I've been using this, the range begins at 1 as appose to 0.

It does begin at zero. The zeroth argument is actually the receiver of the method call. So in something like:

```dart
"string".someMethod("arg")
```

If `someMethod` is a foreign method, when it calls `wrenGetArgumentString(vm, 0)` it will return `"string"` and `wrenGetArgumentString(vm, 1)` will return `"arg"`.

hazbo:
Ahhhh! I see, sorry I didn't take that into account. Well that's good news then. :)

PR #126, opened 2015-01-20
Adding MSVC compatibility

MarcoLizza:
I made some minor changes to the source files to add a MSVC (2013) build files and to resolve the platform errors and warnings.

Hope it helps!

### DIFF
--- a/src/wren_compiler.c
+++ b/src/wren_compiler.c
@@ -2550,7 +2550,10 @@ void statement(Compiler* compiler)
     return;
   }
 
-  if (match(compiler, TOKEN_FOR)) return forStatement(compiler);
+  if (match(compiler, TOKEN_FOR)) {
+    forStatement(compiler);
+    return;
+  }
 
   if (match(compiler, TOKEN_IF))
   {
@@ -2603,7 +2606,10 @@ void statement(Compiler* compiler)
     return;
   }
 
-  if (match(compiler, TOKEN_WHILE)) return whileStatement(compiler);
+  if (match(compiler, TOKEN_WHILE)) {
+    whileStatement(compiler);
+    return;
+  }
 
   // Expression statement.
   expression(compiler);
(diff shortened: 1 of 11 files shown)


src/wren_compiler.c
```diff
@@ -894,7 +898,7 @@ static void ignoreNewlines(Compiler* compiler)
 // discards any duplicate newlines following it.
 static bool consumeLine(Compiler* compiler, const char* errorMessage)
 {
-  bool result = consume(compiler, TOKEN_LINE, errorMessage);
+  bool result = consume(compiler, TOKEN_LINE, errorMessage) != NULL;
```
zeckalpha: Would this make more sense inside consume?

src/wren_vm.c
```diff
@@ -959,7 +963,7 @@ static bool runInterpreter(WrenVM* vm)
       // Capture upvalues.
       for (int i = 0; i < prototype->numUpvalues; i++)
       {
-        bool isLocal = READ_BYTE();
+        bool isLocal = READ_BYTE() != 0;
```
zeckalpha: Same here?

src/wren_compiler.c
MarcoLizza: No, it wouldn't. The "consume(...)" function returns a pointer and this is only a special case check. (same applies for the other similiar change).

munificent:
Thank you for doing this! I like that you split it out into multiple commits (woo!) but as a PR it's a bit monolithic to pull in just yet. I :+1: the commits that I think are pretty much ready as is and pushed a few changes that should address some things you had to work around.

zeckalpha:
Sorry, I didn't bother to look at the methods; just noticed the types
didn't change. Looks great!

MarcoLizza:
I'm closing this pull-request. The new one supersedes it.

Issue #128, opened 2015-01-20
WREN_PIN breaks under some optimizations


MarcoLizza:
Just reviewing the code. In the "wren_value.c" (lines 98 to 100) I guess there a not needed additional WREN_UNPIN "call". I'm a right?

MarcoLizza:
Sorry, didn't noticed the first "pin". However, I'm getting some strange access violations when compiling in release mode in MSVC. Investigating.

MarcoLizza:
It seems due to MSVC release-mode optimizations. For now I'm disabling it, but need further investigation.

MarcoLizza:
Investigated a bit more.

Selectively disable optimization for "wrenPinObj" and "wrenUnpinObj" functions resolve the issue. This is quite odd (and scary).

I guess the MSVC2013U5 optimizer is getting a bit confused when inlining these functions.

MarcoLizza:
I've been inspecting some more. To be honest, I'm not sure the `WREN_PIN` macro to be correct.

In its "body" an automatic variable is declared and its address is stored and used in the internal pin-list. We cannot expect it to be valid outside the `do{ } while(true)` scope. The `WrenPinnedObj` should be allocated dynamically, and freed upon unpinning.

I've change the two functions (and the macro and the signatures as well) as follows

```
void wrenPinObj(WrenVM* vm, Obj* obj)
{
  WrenPinnedObj *pinned = vm->reallocate(NULL, 0, sizeof(WrenPinnedObj));
  pinned->obj = obj;
  pinned->previous = vm->pinned;
  vm->pinned = pinned;
}

void wrenUnpinObj(WrenVM* vm)
{
  WrenPinnedObj *previous = vm->pinned->previous;
  vm->reallocate(vm->pinned, 0, 0);
  vm->pinned = previous;
}
```

*( these modifications render the `WREN_PIN` and `WREN_UNPIN` macros redundant, but I suggest to keep them for clarity )*

Is it correct or I'm a missing something?

munificent:
Yeah, that was totally broken. I don't know what I was thinking there. :open_mouth:

I want to avoid heap allocation here so I changed it to just have a little fixed-size array of temporary roots. I'd initially planned to expose the pinning API to embedders which means it needed to be more flexible, but since then I've tried to make the embedding API not require that. So, for now at least, a fixed-size array is fine.

MarcoLizza:
[...]

PR #130, opened 2015-01-21
Microsoft Visual Studio 2013 project files and compatibility

MarcoLizza:
Here's the new pull-request with the revised changes following the messages we shared during the last hours.

I took the liberty to revise the header inclusion order, too (see commit message).

The optimizer bug has been addressed by selectively disable the inliner for the two incriminated functions. However, I updated the issue #128, please check and let me know.


src/main.c
```diff
@@ -1,3 +1,5 @@
+#include "wren_common.h"
```
munificent: We definitely don't want this. The only public header Wren exposes is wren.h. Including this leaks implementation details of the VM.

MarcoLizza: To be honest, the `main.c` change wasn't needed. That is, since it resides in the wren-library client side, it should be the client duty to ensure not to get warnings.

I'll revert the change for `main.c`.

munificent:
Can you pull the math stuff out into a separate pull request? Also note that many of those methods are already implemented in Wren as instance methods on numbers. Since Wren is deeply OOP, I try to hang methods off natural objects whenever possible. So, where in Java you'd do `math.abs(someNumber)`, in Wren it's just `someNumber.abs`.

MarcoLizza:
Yep, sorry.

That library was meant to be just a "proof of concept" for trying and use the embedding API.

I second the use of methods such as `anotherNumber.int` (can we go ahead and implement this?).

Moreover, the RNG should be `Num.rand` or something like that.

munificent:
Awesome, thank you!

munificent:
> I second the use of methods such as anotherNumber.int (can we go ahead and implement this?).

Yeah, but let's call it `truncate`.

MarcoLizza:
Fair enough, `truncate` is fine.

PR #132, opened 2015-01-21
Fix precedence of % operator

edsrzf:
It previously had the same precedence as + and -.

I assume this was a bug, but feel free to close this pull request if it's not.


munificent:
Wow, amateur hour on my part. Nice catch!

PR #135, opened 2015-01-24
Abstract List's `toString` method to a more general `join` method on Sequence.

gsmaverick:


lluchs:
I think join should be implemented for all Sequences. The if in the loop can also be eliminated by using iterators which may be better for performance.

munificent:
Great idea!

zeckalpha:
Does this solve #3?

gsmaverick:
Thanks for all the comments, I've updated the PR to address them.

@zeckalpha This does not solve #3.

munificent:
Awesome! I love all the tests. :grinning:

PR #138, opened 2015-01-26
Define equality on ranges.

gsmaverick:
I was working on this before you got maps into master but it's now trivial with the new `wrenValuesEqual` method!

### DIFF
--- a/src/wren_core.c
+++ b/src/wren_core.c
@@ -1032,12 +1032,12 @@ DEF_NATIVE(object_not)
 
 DEF_NATIVE(object_eqeq)
 {
-  RETURN_BOOL(wrenValuesSame(args[0], args[1]));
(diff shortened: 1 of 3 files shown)


munificent:
Actually, now that we have `wrenValuesEqual`, it probably makes sense to have *Object*'s `==` operator call that. Then String and Range don't need to override it at all. Feel like doing that in this pull request?

Either way, the test you added for ranges is swell. :)

Also, this reminds me that `wrenValuesEqual` can use `memcmp` instead of `strncmp`.

gsmaverick:
@munificent Good idea, updated PR to include that and the `memcmp` fix as well.

munificent:
Awesome, thanks! I love seeing test coverage grow while total lines of code shrinks. :)

PR #139, opened 2015-01-26
Signed unsigned warnings

MarcoLizza:
I began by snooking some minor signed/unsigned warnings in the max indexing variable use..

.. and finished in changing the maps/lists/strings size-tracking variable type.

It should be `size_t`, as standard "suggests".

### DIFF
--- a/src/wren_core.c
+++ b/src/wren_core.c
@@ -1209,8 +1209,8 @@ DEF_NATIVE(string_iterate)
 
   if (!validateInt(vm, args, 1, "Iterator")) return PRIM_ERROR;
 
-  int index = (int)AS_NUM(args[1]);
-  if (index < 0) RETURN_FALSE;
+  size_t index = (uint32_t)AS_NUM(args[1]);
+  if (index >= string->length) RETURN_FALSE;
 
   // Advance to the beginning of the next UTF-8 sequence.
   do
--- a/src/wren_core.c
+++ b/src/wren_core.c
@@ -659,10 +659,10 @@ DEF_NATIVE(list_iterate)
 
   if (!validateInt(vm, args, 1, "Iterator")) return PRIM_ERROR;
 
-  int index = (int)AS_NUM(args[1]);
+  size_t index = (size_t)AS_NUM(args[1]);
 
   // Stop if we're out of bounds.
-  if (index < 0 || index >= list->count - 1) RETURN_FALSE;
+  if (index >= list->count - 1) RETURN_FALSE;
 
   // Otherwise, move to the next index.
   RETURN_NUM(index + 1);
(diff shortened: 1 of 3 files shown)


munificent:
Awesome. I've been really lazy about number types because I honestly don't know what the best strategy is, so thanks for hacking on this. I think moving away from `int` is probably good, but I think I'd prefer `uint32_t` over `size_t` in most of these places. Maybe it's just me, but it feels weird that object sizes would change (when not needed) when compiled as 64-bit. What do you think?

MarcoLizza:
You're correct, in my opinion, in thinking that (some) types should not change when moving from 32- to 64-bit architecture. Moreover, developing for 64-bit is not only merely a matter of using the correct variable types.

I think, however, that using `size_t` will give the code consistency. Memory allocation and string manipulation (among the others) use `size_t` in the standard C library. So, it would be advisable to use it as much as possible, not only for strings.

Either way ,I suggest to go and use `uint32_t` everywhere and dismiss the `size_t` usage for good.

*( by the way, `ftell` does not return a `size_t` type value )*

Please note that there will be places where using `int` would still perfectly make sense. The usage of negative values (for indexes and lengths) to indicate default cases and/or reverse indexing is very effective, although not always type-safe and obscure (relying of some kind of "side-effect" of the type itself).

munificent:
If it's OK with you, I'm going to close this out because it's a bit too big for me to easily handle. If this is still an area you'd like to hack on, that's totally fine. But I think it would help me to break it into small PRs.

Some random thoughts to consider:

* I don't think `size_t` makes much sense for lists. Since the index comes from user code where it's stored as a `double`, you're never going to get a full 64 bits, even on a 64 bit machine. I think `uint32_t` is plenty. (And honestly, who needs a list with more than 2^32 entries in a scripting language?!)

* We'll have to make sure that some of the various list functions do take signed integers if they accept "negative to index from the end" values.

* For numbers that ultimately end up in the bytecode (local variable slots, method symbols, etc.) they definitely do already have a required size, so it probably makes sense to use that.

* Especially for code in the interpreter, we should make sure changing number types doesn't regress any of the benchmarks.

Sound good?

MarcoLizza:
Ok, I'll go and devise smaller and more easy-to-merge pull requests.

As for your thoughts
* as I wrote in another comment, I also think that `uint32_t` is enough for everybody :grin:;
* yep, I notices the negative index usage, I'll keep that in mind;
* I don't expect noticeable changes in the benchmarks, but let's keep them in as a priority.

MarcoLizza:
I've pondering a bit more, while sweeping throughout the Wren code.

I came to the conclusion that may be better to go the other way around, and use `int` everywhere possible.

Since we decided not to adhere to the "standard" `size_t` alias, there's not benefit in the 32th unsigned bit (we are not expecting to use that amount of chars/items for the reasons you exposed). Moreover, the side-effects in using negative values (e.g. to indicate the whole string length to a function) are valuable.

So, I'll go and switch to `int`.

PR #140, opened 2015-01-26
Using array of pointers to store the predefined types.

MarcoLizza:
Noticed the "TODO" in the source and switched to use an array-of-pointers to store the predefined types variables.

I'm not completely sure it was worth the effort. Surely a "elegant" approach, but may affect the performances due to the needed array referencing math (although the optimizer should dampen this). Moreover, it's not so easy on the eyes when reading.

What's your opinion?


munificent:
Yeah, I think my TODO is wrong. The original code is clearer. I guess now we know! I'll close this out.

MarcoLizza:
Agreed. Let's dismiss the changes it and keep the original cleaner code.

munificent:
:+1:

PR #143, opened 2015-01-27
Bitfields usage

MarcoLizza:
Just noticed that the 'Obj' structure is using bitfields to track the object type and GC flag.

In my opinion this should be avoided. Bitfields are useful when handling packed structures, but that's not the case.

Having not specified a strict struct-field alignment, we would end allocating `sizeof(int)` bytes for each field.

Also, the required math to access the bitfields (mask and shift) can affect performances.


munificent:
:+1:

Thanks!

MarcoLizza:
You're welcome! :tada:

PR #144, opened 2015-01-27
Change Makefile to differentiate between static and shared library

verpeteren:
Previously, just the static libraries (libwren*.a) were created. In the Makefile documentation these were mentioned in the comments as shared.
Now both variants will be created.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,8 @@ wrend
 wren-cpp
 libwren.a
 libwrend.a
+libwren.so
+libwrend.so
 libwren-cpp.a
 
 # XCode user-specific stuff.
--- a/Makefile
+++ b/Makefile
@@ -24,7 +24,7 @@ SOURCES := $(wildcard src/*.c)
 HEADERS := $(wildcard src/*.h)
 OBJECTS := $(SOURCES:.c=.o)
 
-# Don't include main.c in the shared library.
+# Don't include main.c in the libraries.
 DEBUG_OBJECTS := $(addprefix build/debug/, $(notdir $(OBJECTS)))
 RELEASE_OBJECTS := $(addprefix build/release/, $(notdir $(OBJECTS)))
 RELEASE_CPP_OBJECTS := $(addprefix build/release-cpp/, $(notdir $(OBJECTS)))
@@ -38,17 +38,21 @@ RELEASE_CPP_LIB_OBJECTS := $(subst build/release-cpp/main.o,,$(RELEASE_CPP_OBJEC
 all: release
 
 clean:
-	@rm -rf build wren wrend libwren.a libwrend.a
+	@rm -rf build wren wrend libwren.a libwrend.a libwren.so libwrend.so
 
 prep:
 	@mkdir -p build/debug build/release build/release-cpp
 
 # Debug build.
-debug: prep wrend libwrend.a
+debug: prep wrend libwrend.a libwrend.so
 
-# Debug shared library.
+# Debug static library.
 libwrend.a: $(DEBUG_LIB_OBJECTS)
 	$(AR) $@ $^
+
+# Debug shared library.
+libwrend.so: $(DEBUG_LIB_OBJECTS)
+	$(CC) $(DEBUG_CFLAGS) -shared -Wl,-soname,$@ -o $@ $^
 
 # Debug command-line interpreter.
 wrend: $(DEBUG_OBJECTS)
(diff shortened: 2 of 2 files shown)


munificent:
This looks really nice! It will need some tweaking to work on OS X too, but I can take care of that.

First a (possibly obvious) high level question: do we actually care about building a shared lib? Is static what users are going to want anyway? If so, maybe we should just fix the docs in the Makefile and keep things simpler. :)

MarcoLizza:
> do we actually care about building a shared lib?

To be honest, I don't think we do.

When embedding Wren I'll prefer and add the single source/header files rather than having a dependency from an externally built static library.

munificent:
Yeah, I think most people will just use it from source, but I can understand also want to statically link in the library. What I'm mostly asking about is if it's worth also having a *shared* library too.

I don't mind doing the latter if users think it's useful, but I'm always on the lookout for things we can simplify by omitting. :)

MarcoLizza:
Personally, I would go and create/use a dynamically-loaded library, more than a statically-linked one.

I avoided on purpose to include in the MSVC project both the static and the (added) dynamic library builds.

They will be certainly useful in the future once Wren spreads and conquer the embedded-scripting world. :+1: 

I also think it would be more urgent to the define language specification releases, or at lest language versions. I don't want it to sound too pretentious, but it would certainly help during the (future) iterations.

*( also, will help when the precompiled-binaries of the aforementioned libraries will be released :grin: )*

lluchs:
Maybe having an amalgamation (like [sqlite](https://sqlite.org/amalgamation.html)) would be useful too? It makes embedding super easy, as you don't have to build the library separately.

verpeteren:
Hai

I feel that I should explain my pull request a little more:


I am experimenting to build wren into my server project. 
(http://www.urbin.info)

It is basically a daemon with a async-nonblocking networkeventloop with 
sofar:
   * fast webserver
   * webclient
   * database connections to mysql/postgresql
   * many ambitious other plans :-)
I know that such a thing is useless unless it is scriptable in someway.

Sofar I have already embedded mozilla's javascript interpreter, but 
that has a few drawbacks:
  * for spidermonkey is very tedious to write any (cpp-)source that let 
the (javascript) scripts execute any (c)code of the server (and 
visa-versa).
    Spidermonkey's GC-rooting strategy sounds simple (hope to hear more 
on this at FOSSDEM upcomming weekend) but the cpp you need to write, in 
order to achive it feels like javaprogramming.
    I have the feeling that wren might strike a good balance between 
runtime-speed and development speed.
    Besides that: I love uncharted territories; So when I read that 
there is not much (documentation) yet, I need to stick my nose in it 
:-).
  * for Spidermonkey as as static library, the linking takes ages, To 
speed up my code-run-segfault-gdb-cycle, I crafted my Makefiles to also 
handle shared libs.
    I guessed that I could use the same routine with wren as for 
spidermonkey.

So yesterday I was happy fooling around, got something working and 
remembered http://finishonethingtoday.com/ and 
http://www.brepettis.com/blog/2009/3/3/the-cult-of-done-manifesto.html.

Peter Reijnders

munificent:
OK, having support for a shared library sounds good to me. Alas, this patch alone break OS X. Clang has its own syntax for building a shared library. I've got a patch that I *think* gets this working without breaking Linux. It's here: https://github.com/munificent/wren/pull/147

@verpeteren, what do you think?

> I also think it would be more urgent to the define language specification releases, or at lest language versions. I don't want it to sound too pretentious, but it would certainly help during the (future) iterations.

Versions and a real process will definitely be important at *some* point, but I think I want to wait until the implementation is a little more mature before I do that. I don't think the language is fully usable right now, so it's not like "version 0.1.0" of Wren based on today's code would be that useful.

What I think matters most now is just being able to make quick progress towards a feature-complete, stable language. I think having to cut releases and version in stuff will just add friction to that. Once we're more on top of things, I'd be delighted to start versioning Wren like a real project. :)

MarcoLizza:
The language is currently in the early stages of development, and agile and quick iterations are a priority. Versioning, right now, would only slow the things down due to unnecessary bookkeeping.

I'm looking forward to read and collaborate to the first language specification when the time's come. :+1:

munificent:
> I'm looking forward to read and collaborate to the first language specification when the time's come.

Me too! :)

verpeteren:
@verpeteren, what do you think?

looks good. It still compiles on linux :-) (that is, without the -Werror flag)

cc --version
cc (Debian 4.7.2-5) 4.7.2

make
cc -c -std=c99 -Wall -fPIC -Os -Iinclude -o build/release/wren_core.o src/wren_core.c
src/wren_core.c: In function ‘validateKey’:
src/wren_core.c:244:13: warning: cast to pointer from integer of different size [-Wint-to-pointer-cast]

munificent:
> looks good. It still compiles on linux :-)

Awesome, thanks! I'll merge this in then.

>     src/wren_core.c:244:13: warning: cast to pointer from integer of different size [-Wint-to-pointer-cast]

That's fixed by: https://github.com/munificent/wren/commit/8e25aa646e1e30f099192d5598c1bd66dc53f1d3

PR #145, opened 2015-01-28
String lengths as 'int'

MarcoLizza:
Here's the first change.

I switched to use 'int' as the reference type for string lengths, in place of 'size_t'.

I scanned throughout the code to find and change every string length usage (also on function signatures).

### DIFF
--- a/src/wren_compiler.c
+++ b/src/wren_compiler.c
@@ -508,8 +508,8 @@ static void skipBlockComment(Parser* parser)
 // Returns true if the current token's text matches [keyword].
 static bool isKeyword(Parser* parser, const char* keyword)
 {
-  int length = parser->currentChar - parser->tokenStart;
-  int keywordLength = (int)strlen(keyword);
+  uint32_t length = parser->currentChar - parser->tokenStart;
+  uint32_t keywordLength = (uint32_t)strlen(keyword);
   return length == keywordLength &&
       strncmp(parser->tokenStart, keyword, length) == 0;
 }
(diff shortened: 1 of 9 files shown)


munificent:
> I can't see major differences.

Yup, that's well within the noise, so it's fine. Do you still want me to pull this in given my comments on the other PR, or should I sit on this for a bit?

MarcoLizza:
Wait I bit, I'll fix this today (see other comments).

munificent:
Wow, that ended up requiring so many casts. :( Are we sure this is really making the code overall better? I like storing the string length as `uint32_t` in ObjString, but I wonder if we should keep most of the rest of the code the same?

MarcoLizza:
I was asking myself, as well.

I don't like explicit coercion. We always miss something, when we do it. Somewhere in my brain, every time I see a cast, a tiny little bell start ringing telling my the something could be designed better.

*( well, of course there a notable exceptions to this )*

Storing the string length as `uint32_t` but leaving `int` everywhere else is nonsense (assuming that having a 2^32 long string makes sense). Using unsigned types for lengths plays nice with our karma, but is it worth?

Here's my proposal:

1. We model the string length as `uint32_t` and we adapt to the context for the precise type to be used. String-related functions should have a `uint32_t` length in the signature (if not negative value special meaning is required);

2. `AS_NUM` should always be casted to `int`, since coming from a `double` internal representation;

3. Code that manipulates the string itself should use `uint32_t` for indexing and such.

How does this sound?

munificent:
> We model the string length as uint32_t and we adapt to the context for the precise type to be used.

So, by this you mean ObjString and String use `uint32_t` for the length, but if a function is calling, say `strlen()`, we allow the function to use `size_t` for local variables if it leads to less casting?

If so, that sounds good.

> String-related functions should have a uint32_t length in the signature (if not negative value special meaning is required);

:+1:

> AS_NUM should always be casted to int, since coming from a double internal representation;

:+1:

> Code that manipulates the string itself should use uint32_t for indexing and such.

:+1:

MarcoLizza:
> So, by this you mean ObjString and String use uint32_t for the length, but if a function is calling, say strlen(), we allow the function to use size_t for local variables if it leads to less casting?

Yes, that's exactly what I was thinking of! :+1:

munificent:

MarcoLizza:
Those "basic guidelines" will be useful in the future. I hope we'll be using it in the next iterations.

Sometime in the next days I'll fix this again. ;)

munificent:
Yeah, I agree the discussions were really useful.

PR #146, opened 2015-01-28
Changed maps size/capacity type

MarcoLizza:
That one is an easy one. :baby: Add coherence to the other length/size uses.

### DIFF
--- a/src/wren_core.c
+++ b/src/wren_core.c
@@ -776,13 +776,13 @@ DEF_NATIVE(map_iterate)
   if (map->count == 0) RETURN_FALSE;
 
   // If we're starting the iteration, return the first entry.
-  int index = -1;
+  uint32_t index = UINT32_MAX;
   if (!IS_NULL(args[1]))
   {
     if (!validateInt(vm, args, 1, "Iterator")) return PRIM_ERROR;
-    index = (int)AS_NUM(args[1]);
+    index = (uint32_t)AS_NUM(args[1]);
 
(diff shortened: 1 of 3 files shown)


MarcoLizza:
> Yep. Unfortunately, I know this. :sob:

As for the cache and data-alignment that could affect and boost/reduce the performance when applying modifications, I did some test by changing the `gcc` optimization flag to `-O3` (favouring speed over size).

It *seems* that noise fluctuations are reduces. Can you confirm this? (if it's worth the matter)

munificent:
This pull request is a bit intertwined with [the list one](https://github.com/munificent/wren/pull/150/) and the commit history is getting a bit hairy. I know I asked you to separate them out, but would mind squashing the commits and combining these two into a single pull request? Here's what I'm thinking it would have:

* ObjList and ObjMap store their capacity as `uint32_t`.
* The list and map functions in wren_value.h/.c take `uint32_t`.
* `validateIndex()` in wren_core is changed to return `uint32_t` (it still takes `int` as a parameter).
* Any method in wren_core that takes a result from `validateIndex()` and passes it to a function in wren_value.c should be good. The remaining ones are `iterate()` in list and map. Those should check for a negative argument and return and only after that can we cast to a `uint32_t`.

How does that sound?

munificent:

MarcoLizza:
I'll certainly merge the two branches into a single one, but for now I'll concentrate and on other stuff. ;)

PR #149, opened 2015-01-29
Symbols length in the symbol table

MarcoLizza:
Just a minor optimization, by getting rid of `strlen()` when accessing the symbol-table.


munificent:
Cool. I assume this doesn't really affect perf, right?

MarcoLizza:
I got the same "cache-presumed" performance boost we discussed in the other comment. Performances have not decreased.

munificent:
:+1:

PR #150, opened 2015-01-30
Lists type to 'uint32_t'

MarcoLizza:
Just to come to full circle, here's the modification for the list type, too.

I performed a benchmark regression test on a "unification" branch with all my recent pull-request, and here's the results
```
binary_trees - wren            ..........  2545  0.39s  107.32% relative to baseline
delta_blue - wren              ..........  4927  0.20s  104.56% relative to baseline
fib - wren                     ..........  1566  0.64s   96.72% relative to baseline
for - wren                     ..........  3671  0.27s  104.67% relative to baseline
method_call - wren             ..........  3340  0.30s  114.32% relative to baseline

```

### DIFF
--- a/src/wren_core.c
+++ b/src/wren_core.c
@@ -659,10 +659,10 @@ DEF_NATIVE(list_iterate)
 
   if (!validateInt(vm, args, 1, "Iterator")) return PRIM_ERROR;
 
-  int index = (int)AS_NUM(args[1]);
+  uint32_t index = (uint32_t)AS_NUM(args[1]);
 
   // Stop if we're out of bounds.
-  if (index < 0 || index >= list->count - 1) RETURN_FALSE;
+  if (index >= list->count - 1) RETURN_FALSE;
 
   // Otherwise, move to the next index.
   RETURN_NUM(index + 1);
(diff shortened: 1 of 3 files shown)


src/wren_core.c
```diff
@@ -659,10 +659,10 @@ DEF_NATIVE(list_iterate)
 
   if (!validateInt(vm, args, 1, "Iterator")) return PRIM_ERROR;
 
-  int index = (int)AS_NUM(args[1]);
+  uint32_t index = (uint32_t)AS_NUM(args[1]);
 
   // Stop if we're out of bounds.
-  if (index < 0 || index >= list->count - 1) RETURN_FALSE;
+  if (index >= list->count - 1) RETURN_FALSE;
```
munificent: I think these two lines should stay the same. A negative value may be passed to this, and we should handle it correctly. Casting a negative value to a `uint32_t` might do something wrong here.

MarcoLizza: For the sake of clearness, I agree with you.

However, casting a negative value to unsigned should give a value greater than the list length, so we can get rid of a comparison (with a minor performance improvement).

However, I'm in favour of maintaining clearness over (minor) optimizations.

Should I revert this?

munificent:

MarcoLizza:
Fine. Perhaps in the next days I'll just propose a fix for some minor warning.

I'll keep the branches for future reference.

PR #157, opened 2015-02-02
8 bit clean

MarcoLizza:
Here's a preview of the modifications.

*(still needs more test cases, maybe)*

Also, as I wrote in a `TODO` comment, a better string matching algorithm should be implemented. Will work on the subject this evening.

### DIFF
--- a/src/wren_compiler.c
+++ b/src/wren_compiler.c
@@ -1265,7 +1265,8 @@ static int copyName(Compiler* compiler, char* name)
     length = MAX_METHOD_NAME;
   }
 
-  strncpy(name, token->start, length);
+  memcpy(name, token->start, length);
+  name[length] = '\0';
   return length;
 }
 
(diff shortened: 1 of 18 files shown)


MarcoLizza:
I did some benchmarks to compare the naive brute-force (BF) algorithm with some "smarter" search algorithms, that is Quick-Search (QS), Knuth-Morris-Pratt (KMP), and Boyer-Moore-Horspool (BMH).

There can't be a definitive all-answer since the "best" algorithm heavily depends to the data itself.

On small strings BF is faster than the other (not requiring the initialization step). On longer strings, BMH behaves better on the average. BMH is also the cheaper one, with respect to the additional memory.

If we aim to have fast string operations I think we should go and use the naive algorithm for small strings and switch to BMH for longer ones. Defining the threshold could be difficult, although.

Thoughts?

munificent:
This looks really awesome, but it's also a big pull request. I only get small pieces of free time, so it's hard for me to find enough time to handle this all at once. Do you think you could break it into separate smaller pull requests? Most of these commits could be standalone PRs, I think.

Thanks!

MarcoLizza:
> Do you think you could break it into separate smaller pull requests? Most of these commits could be standalone PRs, I think.

Yep, will do.

Not every commit can be isolated in a single pull-request, but I'll do my best.

MarcoLizza:
Done. We can close this pull-request and go for the other ones!

PR #159, opened 2015-02-04
Adding '\0' escape sequence for null characters.

MarcoLizza:
With this (trivial) pull-request we'll be recognising the '\0' escape sequence, as well.


munificent:
:metal:

PR #161, opened 2015-02-04
Replacing 'strstr()' with 'wrenStringFind()'

MarcoLizza:
In a single commit/pull-request we are getting rid of "strstr()" and using the new function, implementing BMH.

### DIFF
--- a/src/wren_core.c
+++ b/src/wren_core.c
@@ -1159,7 +1159,7 @@ DEF_NATIVE(string_contains)
   // Corner case, the empty string contains the empty string.
   if (string->length == 0 && search->length == 0) RETURN_TRUE;
 
-  RETURN_BOOL(strstr(string->value, search->value) != NULL);
+  RETURN_BOOL(wrenStringFind(vm, string, search) != string->length);
 }
 
 DEF_NATIVE(string_count)
@@ -1191,9 +1191,9 @@ DEF_NATIVE(string_indexOf)
   ObjString* string = AS_STRING(args[0]);
   ObjString* search = AS_STRING(args[1]);
 
-  char* firstOccurrence = strstr(string->value, search->value);
+  uint32_t index = wrenStringFind(vm, string, search);
 
-  RETURN_NUM(firstOccurrence ? firstOccurrence - string->value : -1);
+  RETURN_NUM(index == string->length ? -1 : (int)index);
 }
 
 DEF_NATIVE(string_iterate)
--- a/src/wren_core.c
+++ b/src/wren_core.c
@@ -1156,8 +1156,8 @@ DEF_NATIVE(string_contains)
   ObjString* string = AS_STRING(args[0]);
   ObjString* search = AS_STRING(args[1]);
 
-  // Corner case, the empty string contains the empty string.
-  if (string->length == 0 && search->length == 0) RETURN_TRUE;
+  // Corner case, the empty string is always contained.
+  if (search->length == 0) RETURN_TRUE;
 
   RETURN_BOOL(wrenStringFind(vm, string, search) != string->length);
 }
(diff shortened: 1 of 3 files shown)


munificent:
I did some clean-up on it, but it looks great! I love the choice of Boyer-Moore-Horspool.

MarcoLizza:
Love you're cleaning-up and commenting, too! :v: 

It wasn't easy to decide the return value for the string-not-found case. `UINT_MAX` is perhaps a better choice, saving us a struct-field access on result check.

I notice that you preferred not to use the union but a straight cast. Guess I was I bit too cautious in using it (made the code not so easy on the eyes, although not affecting performances).

Glad you appreciated the use of Boyer-Moore-Horspool! I like it quite a bit! :wink:

*( although Quick-Search behaves very well to... I'll go and try it just out of curiosity to see the performance impact )*

munificent:
> UINT_MAX is perhaps a better choice, saving us a struct-field access on result check.

Yeah, more importantly, it makes it possible to correctly handle an empty string for both the haystack and needle. :)

> I notice that you preferred not to use the union but a straight cast. Guess I was I bit too cautious in using it (made the code not so easy on the eyes, although not affecting performances).

Yup! I wasn't sure how C specified this so I spent a bit of time researching. Both C and C++ specify that casting between signed and unsigned numbers (or pointers to them) is always allow and just reinterprets the bits without doing any conversion. So just adding a couple of explicit casts here is safe and well-defined.

> ( although Quick-Search behaves very well to... I'll go and try it just out of curiosity to see the performance impact )

If you want, it would be cool to have a benchmark or two for testing performance of string functions.

MarcoLizza:
> Yeah, more importantly, it makes it possible to correctly handle an empty string for both the haystack and needle. :)

Yep, you're right. In my version I left the empty-needle-string as an additional (corner case) check in the caller. This way is better defined.

> Both C and C++ specify that casting between signed and unsigned numbers (or pointers to them) is always allow and just reinterprets the bits without doing any conversion.

As long as you are casting from `signed` and `unsigned` of the same type we are safe. Not sure when converting, for example, from `signed char` to `unsigned int` (but that's not the case).

> If you want, it would be cool to have a benchmark or two for testing performance of string functions.

I was thinking the same thing! A specific string functions benchmark could be very useful. I'll devise and write them down! :+1: 

Speaking of strings... I was thinking about string formatting. At the moment, Wren does not have an explicit formatting feature, am I right? It's not a big issue right now, but could be useful in some occasions.

Perhaps, something like a `toString` arity variant method that accept a format specifier (something like .NET, so to speak) would be nice. I would avoid a `String.format(...)`.

What's your opinion?

munificent:
> At the moment, Wren does not have an explicit formatting feature, am I right?

Right.

> What's your opinion?

I haven't thought about it much. They are useful, but I worry that implementing it will be a lot of code (either Wren or C) and I'm always trying to keep the codebase as small as possible.

MarcoLizza:
> They are useful, but I worry that implementing it will be a lot of code (either Wren or C) and I'm always trying to keep the codebase as small as possible.

That's my opinion, too (like with regex). Adding them to Wren would mean going in the wrong direction. Wren is, first of all, *small* and *light* (or it won't be named "wren").

In the future, when Wren will feature dynamic-library-based modules (and it surely will, won't it? :wink:) one could implement and extend it by himself/herself.

munificent:
> In the future, when Wren will feature dynamic-library-based modules (and it surely will, won't it? :wink:)

Maybe, though it may not be part of the core Wren language and instead live in [wren-io](https://github.com/zeckalpha/wren-io). Formatting strings, though could probably be written in pure Wren, so it might not even need dynamically loaded C.

zeckalpha:
Help with wren-io is appreciated! I've had a busy week or two, but I hope to get back to it next week.

zeckalpha:
Oh, and I did start a pure-wren regex library which will be in wren-io once it is ready.

PR #162, opened 2015-02-04
Concatenate null-containing strings

MarcoLizza:
Another easy one, we are changing 'wrenStringConcat()' to support 8-bit strings.


munificent:
\o/

MarcoLizza:
Yay! Sorry for missing the 80-column limit, although!

PR #168, opened 2015-02-06
```make clean``` now removes the libraries

### PR DESCRIPTION
**Pull request by Phyllostachys**
```make clean``` now removes ```libwren.a``` and ```libwren.so``` where it only removed ```build``` and ```wren``` before.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -48,7 +48,7 @@ RELEASE_CPP_LIB_OBJECTS := $(subst build/release-cpp/main.o,,$(RELEASE_CPP_OBJEC
 all: release
 
 clean:
-	@rm -rf build wren wrend libwren libwrend
+	@rm -rf build wren wrend libwren* libwrend*
 
 prep:
 	@mkdir -p build/debug build/release build/release-cpp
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: Makefile [hunk 1]
```diff
@@ -48,7 +48,7 @@ RELEASE_CPP_LIB_OBJECTS := $(subst build/release-cpp/main.o,,$(RELEASE_CPP_OBJEC
 all: release
 
 clean:
-	@rm -rf build wren wrend libwren libwrend
+	@rm -rf build wren wrend libwren* libwrend*
```
munificent: Seeing any kind of `*` after `rm -rf` gives me the heebie-jeebies, but I agree that the current Makefile is wrong. How about just listing all four names explicitly?

Phyllostachys: I agree on the heebie-jeebie aspect of it; seems like a code smell. I'll make it more explicit.

Phyllostachys:
Oh, I forgot to revert the change and add it as just one commit. :-/

munificent:
Thanks! Squashing is definitely appreciated, but it's not a big deal.

Issue #141, opened 2015-01-26
Consider alternate hash table representation


munificent:
Once other stuff is more mature and we have realistic benchmarks for maps, it may be worth looking at [this representation](http://morepypy.blogspot.dk/2015/01/faster-more-memory-efficient-and-more.html) PyPy switched to.

In addition to saving memory and improving cache coherence, this would also give us a deterministic iteration order for maps, which would be nice.

PR #169, opened 2015-02-09
Benchmark script

MarcoLizza:
As we discussed some days ago, I moved (and renamed) the benchmark driver script to the `script` folder.

Also, I added some simple map benchmark scripts (very basic, just to have some kind of comparison with other languages). Oddlily, on my setup `luajit` gives a wrong result... :question: 

For the sake of curiosity, here's the "splitted" benchmark results I get
```
INSERT
wren   0.117443
python 0.013316
ruby   0.036993835
lua    0.00715400

LOOKUP
wren   0.021521
python 0.017457
ruby   0.0208651
lua    0.00565200

REMOVE
wren   0.189044
python 0.00957
ruby   0.017944434
lua    0.00477000
```
The benchmark scripts, however, calculate just the cumulative INSERT/LOOKUP/REMOVE time.

*(string benchmarks will follow in the next days)*

### DIFF
--- a/script/run_bench.py
+++ b/script/run_bench.py
@@ -44,6 +44,8 @@ BENCHMARK("for", r"""499999500000""")
 BENCHMARK("method_call", r"""true
 false""")
 
+BENCHMARK("maps", r"""5000050000""")
+
 LANGUAGES = [
   ("wren",           [os.path.join(HOME_DIR, 'wren')], ".wren"),
   ("lua",            ["lua"],                          ".lua"),
--- a/script/run_bench.py
+++ b/script/run_bench.py
@@ -11,8 +11,8 @@ import subprocess
 import sys
 
 # Runs the tests.
-WREN_DIR = os.path.dirname(os.path.dirname(os.path.realpath(__file__)))
-BENCHMARK_DIR = os.path.join(WREN_DIR, 'benchmark')
+HOME_DIR = os.path.dirname(os.path.dirname(os.path.realpath(__file__)))
+BENCHMARK_DIR = os.path.join(HOME_DIR, 'benchmark')
 
 # How many times to run a given benchmark.
 NUM_TRIALS = 10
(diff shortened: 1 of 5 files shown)


script/run_bench.py
```diff
@@ -11,8 +11,8 @@
 import sys
 
 # Runs the tests.
-WREN_DIR = os.path.dirname(os.path.dirname(os.path.realpath(__file__)))
-BENCHMARK_DIR = os.path.join(WREN_DIR, 'benchmark')
+HOME_DIR = os.path.dirname(os.path.dirname(os.path.realpath(__file__)))
```
zeckalpha: Hmm, HOME_DIR seems to imply something else. Maybe BASE_DIR?

MarcoLizza: It mostly depend on what you're calling "home". :)

In my opinion, "home" is the place where all the Wren source-and-binary stuff resides. I'm open to suggestions, however.

munificent: Given that "wren" is the directory you'll end up with when you clone the Git repo, I think `WREN_DIR` works pretty well here. I agree with @zeckalpha—I'd intuit `HOME_DIR` to mean `~`.

munificent:
Thanks for this! I highlighting something totally broken in the map implementation, which is super helpful. I'll work on fixing that soon.

script/run_bench.py
MarcoLizza: Sounds good for me.

MarcoLizza:
That's awesome! I think we can improve the performances more, although.

Lua uses quite a neat implementation, basing on the results... needs further studies...

munificent:
Yeah, Lua is pretty smart (though some of its optimizations don't apply because it unifies arrays and tables, which Wren doesn't do). I'm also interested in #141. But I don't want to put too much effort into optimizing this until we have more real-world code to benchmark against.

Until then, I'm going for "as simple as possible while not totally dumb". The previous code (which I pushed a fix for last night) was still in "totally dumb" territory. :)

MarcoLizza:
> Yeah, Lua is pretty smart (though some of its optimizations don't apply because it unifies arrays and tables, which Wren doesn't do).

Yep, Lua's tables/arrays/heaps are quite its core... it would be a pity if they weren't so optimized.

As you may know, I totally agree that we should start with decent and clean (but not naive) implementations in order to keep the implementation accessible.

As for #141, I did noticed it and I'm quite intrigued in trying it... but there will be time for it.

PR #174, opened 2015-02-16
IO.time

MarcoLizza:
In case someone wants to seed a PRNG with some kind of "noise", something like this is needed.


munificent:
:+1:

MarcoLizza:
:sunny:

PR #178, opened 2015-02-18
Initial implementation of raw/multiline strings

pwoolcoc:
This implementation of raw strings pretty much copies the Lua way exactly. The delimiters are the same, `[[` and `]]`, and can be nested arbitrarily. It also ignores the first character if it is a newline.

### DIFF
--- a/src/wren_compiler.c
+++ b/src/wren_compiler.c
@@ -732,6 +732,59 @@ static void readString(Parser* parser)
   makeToken(parser, TOKEN_STRING);
 }
 
+// Finishes lexing a string literal.
+static void readRawString(Parser* parser)
+{
+  wrenByteBufferClear(parser->vm, &parser->string);
+
+  int nesting = 0;
+
+  if (peekChar(parser) == '\n') {
+    // To behave the same as Lua multiline strings, ignore
+    // the first char if it is a newline.
+    nextChar(parser);
+  }
+
+  for (;;) {
+    char c = nextChar(parser);
+
+    if (c == '[') {
+      char c2 = peekChar(parser);
+      if (c2 == '[') {
+        nesting++;
+      }
+    }
+
+    if (c == ']') {
+      char c2 = peekChar(parser);
+      if (c2 == ']') {
+
+        if (nesting > 0) {
+          nesting--;
+        } else {
+          nextChar(parser);
+          break;
+        }
+      }
+    }
+
+    if (c == '\0')
+    {
+      lexError(parser, "Unterminated string.");
+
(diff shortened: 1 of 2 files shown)


pwoolcoc:
I should add, right now there are some map & list tests failing with this build, I wanted to get some feedback here before fixing them, since the tests are failing specifically because `[[` is now valid syntax. So, I will fix those tests if it is decided that `[[` and `]]` are the delimiters we want to go with.

munificent:
Yeah, that syntax would definitely be problematic in Wren. `[[123]]` is already valid—it's a one-element list containing a one-element list containing `123`.

For the sake of keeping things minimal and simple (always my number one goal with Wren), I wasn't planning to add raw strings or other string forms. Do you feel they're really that useful?

One option would be to allow newlines in *all* strings. That wouldn't cover raw strings, but would at least cover multiline strings without adding any real complexity.

What do you think?

pwoolcoc:
I thought I saw you mention somewhere that raw strings would be preferred to regex literals, so I took a stab at adding them. I don't generally use them too much in python (outside of regexes), but they are nice to have when you want to avoid escape-hell.

I would have no problem with allowing newlines in all strings.

MarcoLizza:
> For the sake of keeping things minimal and simple (always my number one goal with Wren), I wasn't planning to add raw strings or other string forms. Do you feel they're really that useful?

Quite frankly, just my two cents, I never found "verbatim/multi-line" strings to be useful at all. Maybe this is due to my C/C++ background, but the first time I stumbled upon them (in PHP or Perl several years ago, don't remember) I found them quite useless (although they made sense in their context).

In recent times, I found more useful C#'s `@-strings`. But I still use the old-fashioned C way only.

munificent:
> I thought I saw you mention somewhere that raw strings would be preferred to regex literals, so I took a stab at adding them.

Ah, foolish me, I led you astray! I do think I would prefer raw strings over regex literals, but I'd still probably prefer *neither* unless we find ourselves *really* wanting them.

> I would have no problem with allowing newlines in all strings.

Let's punt on this for a while too, just in case it does cause problems. We can always add it later.

pwoolcoc:
> Ah, foolish me, I led you astray! 

Nah, foolish me, for not checking in first :)

lluchs:
>C/C++ background

C++11 actually added raw string literals. They look like this:
```c++
R"(foo
bar)"
```

MarcoLizza:
@lluchs, C++ was quite far from having raw string literals when I first started using it... ;)

PR #183, opened 2015-02-23
Add a `fromString` static method on `Num` that converts strings to numbers.

gsmaverick:


gsmaverick:
@lluchs good call, thanks for catching that.

test/number/from_string.wren
```diff
@@ -0,0 +1,8 @@
+IO.print(Num.fromString("123") == 123) // expect: true
+IO.print(Num.fromString("-123") == -123) // expect: true
+IO.print(Num.fromString("-0") == -0) // expect: true
+IO.print(Num.fromString("12.34") == 12.34) // expect: true
+IO.print(Num.fromString("-0.0001") == -0.0001) // expect: true
+
+// Test a non-number literal and ensure it returns null.
+IO.print(Num.fromString("test1") == null) // expect: true
```
munificent: Can you add a couple more edge case tests? I'm thinking `""` (empty string), `"  12  "` (number with extra whitespace), and `"1.2text"` (prefix is a number).

munificent:
This is a great addition! I added one little comment but aside from that, this looks swell. At some point, we should handle range errors (and test them), but I'm OK with that being a TODO for now.

gsmaverick:
@munificent pr updated.

munificent:
Awesome, thanks! I tweaked it a bit to not allow trailing non-number characters since that felt weird to me. But otherwise, this is in now!

gsmaverick:
Thanks! I think this might be the last roadblock to writing a JSON encoding/decoding library.

munificent:
> I think this might be the last roadblock to writing a JSON encoding/decoding library.

Crazy! Almost like a real language! :)

PR #186, opened 2015-02-23
C++ warnings

MarcoLizza:
Some minor fixes:
* when compiling as C++ with GCC there were some blocking warnings due to (missing) `const` string modifier;
* I've added the current working directory detection.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -11,7 +11,8 @@
 // This is the source file for the standalone command line interpreter. It is
 // not needed if you are embedding Wren in an application.
 
-char* rootDirectory = "";
+char buffer[2048] = { '\0' };
+char* rootDirectory = buffer;
 
 static void failIf(bool condition, int exitCode, const char* format, ...)
 {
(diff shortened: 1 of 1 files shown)


munificent:
Thanks for cleaning this up! Instead of having to detect the platform, I did the simpler thing. I realized if you want a path relative to the cwd, the easiest way to get that is... just a relative path. :)

MarcoLizza:
> I realized if you want a path relative to the cwd, the easiest way to get that is... just a relative path. :)

True.

But will this work even when running Wren from within (for example) the home directory while the executable itself resides in (let's say) `/usr/bin`?

munificent:
> But will this work even when running Wren from within (for example) the home directory while the executable itself resides in (let's say) /usr/bin?

Yup! The cwd is based on what the cwd was when the executable was started, not the directory containing the executable itself.

MarcoLizza:
Goody-goody.

That's what I expected, too. Fine.

PR #187, opened 2015-02-23
Bitwise precedence

MarcoLizza:
With this one we should have fixed the bitwise operators precedence.

### DIFF
--- a/src/wren_compiler.c
+++ b/src/wren_compiler.c
@@ -1382,10 +1382,9 @@ typedef enum
   PREC_NONE,
   PREC_LOWEST,
   PREC_ASSIGNMENT,    // =
-  PREC_LOGIC,         // && ||
-  PREC_BITWISE_OR,    // |
-  PREC_BITWISE_XOR,   // ^
-  PREC_BITWISE_AND,   // &
+  PREC_TERNARY,       // ?:
+  PREC_LOGICAL_OR,    // ||
+  PREC_LOGICAL_AND,   // &&
   PREC_EQUALITY,      // == !=
   PREC_IS,            // is
   PREC_COMPARISON,    // < > <= >=
@@ -2166,7 +2165,7 @@ static void and_(Compiler* compiler, bool allowAssignment)
 
   // Skip the right argument if the left is false.
   int jump = emitJump(compiler, CODE_AND);
-  parsePrecedence(compiler, false, PREC_LOGIC);
+  parsePrecedence(compiler, false, PREC_LOGICAL_AND);
   patchJump(compiler, jump);
 }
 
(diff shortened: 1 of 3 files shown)


src/wren_compiler.c
```diff
@@ -1381,17 +1381,20 @@ typedef enum
...
-  PREC_IS,         // is
-  PREC_COMPARISON, // < > <= >=
-  PREC_RANGE,      // .. ...
-  PREC_BITWISE,    // | ^ & << >>
-  PREC_TERM,       // + -
-  PREC_FACTOR,     // * / %
-  PREC_UNARY,      // unary - ! ~
-  PREC_CALL,       // . () []
+  PREC_ASSIGNMENT,    // =
+  PREC_LOGIC,         // && ||
```
munificent: We should probably split these out too. C makes `&&` higher precedence than `||`.

src/wren_compiler.c
```diff
@@ -1381,17 +1381,20 @@ typedef enum
...
+  PREC_BITWISE_OR,    // |
+  PREC_BITWISE_XOR,   // ^
+  PREC_BITWISE_AND,   // &
```
munificent: We can and should give these each their own precedence levels, but let's move all of the bitwise operators to between `PREC_COMPARISON` and `PREC_RANGE`.

That's different from C, but I think C's precedence choice there was a mistake. It means you can't do natural things like:

```c
if (flags & SOME_BIT == SOME_BIT) ...
```

Making it higher than range would be useful because the result of a bitwise expression is usually a number, which is a meaningful argument to a range. But given that we don't put spaces around `..`, I think it would read confusingly if bitwise operators had higher precedence:

```c
a & b..c
```

I think most people would parse that as `a & (b..c)`, not `(a & b)..c`.

test/number/bitwise_precedence.wren
```diff
@@ -0,0 +1,13 @@
+IO.print(2 | 1 << 1) // expect: 2
```
munificent: Yay tests! This test really isn't about numbers as much as it is syntax (though it, of course, uses method calls on numbers to verify the parse), so let's move it up to just `test/` next to `precedence.wren`.

src/wren_compiler.c
MarcoLizza: Yep. Wil fix it at once! :+1:

test/number/bitwise_precedence.wren
MarcoLizza: Okie-dokie!

src/wren_compiler.c
MarcoLizza: I second this. I also think that C precedence aspect was in a grammar quirk.

As for the range-bitwise precedence, I'm disputed, but agree that the common sense would interpret `a & b..c` as `a & (b..c)`.

I'll fix these ones, too.

MarcoLizza:
All done, @munificent!

munificent:
:metal:

PR #189, opened 2015-02-24
test/number/literals.wren: test hex literals

hachibu:
I wrote a few lines to exercise printing different types of hex literals.


gsmaverick:
There's already some tests for this in https://github.com/munificent/wren/blob/master/test/number/hex_literals.wren. I just forgot to update this comment when I added them.

hachibu:
Ah ok, cool.

PR #192, opened 2015-02-25
Generating a runtime error on subclassing of builtins

verpeteren:
As discussed in https://github.com/munificent/wren/issues/70, it is not allowed to subclass built-in types.

Please review, especcially  test/closure/subclassing_closure.wren

### DIFF
--- a/test/closure/subclassing_closure.wren
+++ b/test/closure/subclassing_closure.wren
@@ -1,0 +1,16 @@
+var f = null
+
+{
+  var a = "a"
+  f = new Fn {
+    IO.print(a)
+  }
+}
+//@FIXME: repair the segfault create a propper test for the subclassing
+
+//var closureSegfault = f
+//class SubClosureSegfault is closureSegfault {} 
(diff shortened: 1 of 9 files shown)


test/closure/subclassing_closure.wren
```diff
@@ -0,0 +1,16 @@
...
+{
+  var a = "a"
+  f = new Fn {
+    IO.print(a)
+  }
+}
+//@FIXME: repair the segfault create a propper test for the subclassing
+
+//var closureSegfault = f
+//class SubClosureSegfault is closureSegfault {}
```
munificent: This is running into another bug in the VM. In this case, the superclass isn't the closure *class* it's the closure *instance*. There's another issue (with a TODO) where we need to ensure the superclass is a class at all. It's not checking that, hence the nasty segfault.

munificent:
Awesome, thank you! I reorganized a bit, but otherwise this is right on the money. :+1:

PR #196, opened 2015-02-26
Set rootDirectory to current working directory so imports work.

verpeteren:
I started with a 'simple' TODO that I found in main.c

The rootDirectory was not set if the scriptname did not have a '/'.
There is no test yet, but the following works now:

* scenario 1:

  ```
  cd example/import-module
  ../../wren lovecraft.wren 
  Ph'nglui mglw'nafh Cthulhu R'lyeh wgah'nagl fhtagn
  ```

* scenario 2:

  ```
  ./wrend example/import-module/lovecraft.wren 
  Ph'nglui mglw'nafh Cthulhu R'lyeh wgah'nagl fhtagn
  ```

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -3,6 +3,7 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
+#include <unistd.h>
 
 #include "wren.h"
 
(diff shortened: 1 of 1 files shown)


MarcoLizza:
See #186 .

PR #200, opened 2015-02-28
doc/site/classes.markdown: add static fields example

hachibu:
I added a static fields example to classes documentation. Any suggestions or changes are totally welcome.


munificent:
Awesome, thanks! I made a few changes, but I really appreciate this. Eventually, I'd like to change this to use an example that feels a little more tangible. I try to avoid "foo" and "bar" in docs because it can be a little vague for people. But this is perfect for a start.

(Granted, `Unicorn` isn't *that* much more meaningful, but at least it's something the reader can picture. :grinning:)

hachibu:
No problem. I like the changes you made. And I agree with you on the foo bar stuff.

PR #201, opened 2015-02-28
Added Sequence.any as complement to Sequence.all

bjorn:
Just a small addition that I think is useful & expected.

A small off-topic question but I thought I'd ask it here. Why does `List.addAll` return the argument rather than modified list? Returning the modified list would allow chaining while I'm not sure what use the argument would be.


gsmaverick:
Nice! You'll want to update the list docs to mention this new method as well.

bjorn:
@gsmaverick You're right of course. Added now. :-)

doc/site/core/sequence.markdown
```diff
@@ -18,6 +18,18 @@ If it returns `false`, stops iterating and returns `false`. Otherwise, returns
     [1, 2, 3].all {|n| n > 2} // False.
     [1, 2, 3].all {|n| n < 4} // True.
 
+### **any**(predicate)
+
+Tests whether any element in the sequence passes the `predicate`.
+
+Iterates over the sequence, passing each element to the function `predicate`.
+If it returns `true`, stops iterating and returns `true`. Otherwise, returns
```
gsmaverick: Wouldn't `[1, 2, 3].any { |n| 1 }` return true?

bjorn: Hmm, it does, but only because `1` is considered trueish. You'd have to write `[1, 2, 3].any { |n| n == 1 }` or `[1, 2, 3].contains(1)` to get what you want. This does not represent a problem with this patch, right?

munificent: I think @gsmaverick's point is that the predicate doesn't have to return literal boolean `true`, but can return any truthy value. `all` is the same way. Maybe we should clarify the docs by saying something like, "a true value"?

test/list/any.wren
```diff
@@ -0,0 +1,10 @@
+var a = [1, 2, 3]
+
+var b = a.any {|x| x > 3 }
+IO.print(b) // expect: false
+
+var d = a.any {|x| x > 1 }
+IO.print(d) // expect: true
+
+var e = [].any {|x| true }
+IO.print(e) // expect: false
```
munificent: Great tests!

Can you also add a test that it throws a runtime error if the argument isn't callable? You can just copy the `test/list/all_non_function_arg.wren` and change it to call `.any`.

munificent:
This is swell!

doc/site/core/sequence.markdown
gsmaverick: Yeah, I probably should have written more than one sentence there. I was intending to point out that the documentation wasn't accurately describing the behaviour of the implementation.

bjorn:
Alright, reworded the documentation a little and added more tests based the ones of `any`. A small cleanup patch sneaked in as well.

I hope the wording in the documentation is alright like this. Only `false` and `null` will "evaluate to false" but I thought it was anyway too verbose to mention both values.

munificent:
Wonderful, thanks!

PR #202, opened 2015-02-28
Made the fib benchmark a bit faster

bjorn:
This way `fib.wren` was consistently about 6% faster for me.

Feel free to close this pull request if the alternative notation deviates too much from the other languages. I found it mainly interesting that apparently free functions are slower to work with than classes, at least in this case. It's probably because of the recursive calls saving one lookup?

Note that the `static` did not matter much in the benchmark at all, I only wrote that to simplify things. The same speed advantages was present when leaving it out and doing `var fib = new Fib` instead.

### DIFF
--- a/benchmark/fib.wren
+++ b/benchmark/fib.wren
@@ -1,10 +1,12 @@
-var fib = new Fn {|n|
-  if (n < 2) return n
-  return fib.call(n - 1) + fib.call(n - 2)
+class Fib {
+  static get(n) {
+    if (n < 2) return n
+    return get(n - 1) + get(n - 2)
+  }
 }
 
 var start = IO.clock
 for (i in 1..5) {
-  IO.print(fib.call(28))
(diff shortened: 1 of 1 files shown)


soveran:
I think the goal is to compare functions across languages, thus even if `Fn` is slower, that's what should be benchmarked.

munificent:
> I found it mainly interesting that apparently free functions are slower to work with than classes, at least in this case. It's probably because of the recursive calls saving one lookup?

My guess is that it's because functions have to handle checking the arity at runtime where method calls rely on the static signature to handle that. I'd have to profile to see.

> I think the goal is to compare functions across languages, thus even if Fn is slower, that's what should be benchmarked.

Yeah, I agree with you in general. But I think this patch is OK. Methods are the natural way to define something callable in Wren. For example, a C# or Java benchmark would likely use method calls for this too even though they both also have more direct first-class function-like things.

PR #205, opened 2015-03-03
Generate an error if a number literal is too large to be represented with the `Num` class.

gsmaverick:
This closes out a couple of TODOs around really big number literals.


munificent:
:+1:

PR #208, opened 2015-03-04
More `Num` methods

MarcoLizza:
Here thery are some additions I had stashed some time ago. Guess now it's time for them to be commited.


src/wren_core.c
```diff
@@ -878,6 +878,24 @@ DEF_PRIMITIVE(num_cos)
   RETURN_NUM(cos(AS_NUM(args[0])));
 }
 
+DEF_PRIMITIVE(num_decimal)
```
munificent: This isn't the right name, since it doesn't deal with decimal representation. How about `fraction`?

src/wren_core.c
```diff
@@ -878,6 +878,24 @@ DEF_PRIMITIVE(num_cos)
...
+{
+  double value = AS_NUM(args[0]);
+  RETURN_NUM(value - (int32_t)value);
```
munificent: This will do the wrong thing on double values that don't fit in an `int32_t`. I'd use `modf()` from math.h.

test/number/decimal.wren
```diff
@@ -0,0 +1,8 @@
+IO.print(123.decimal)      // expect: 0
+IO.print((-123).decimal)   // expect: 0
+IO.print(0.decimal)        // expect: 0
+IO.print((-0).decimal)     // expect: -0
+IO.print(0.123.decimal)    // expect: 0.123
+IO.print(12.3.decimal)     // expect: 0.3
+IO.print((-0.123).decimal) // expect: -0.123
+IO.print((-12.3).decimal)  // expect: -0.3
```
munificent: Test a really big number that doesn't fit in an int.

src/wren_core.c
```diff
@@ -878,6 +878,24 @@ DEF_PRIMITIVE(num_cos)
...
+}
+
+DEF_PRIMITIVE(num_deg)
```
munificent: Maybe it's just me, but I don't think this is that useful. Leave it out?

src/wren_core.c
```diff
@@ -878,6 +878,24 @@ DEF_PRIMITIVE(num_cos)
...
+{
+  RETURN_NUM(floor(AS_NUM(args[0]) * 57.2957795130823208768));
+}
+
+DEF_PRIMITIVE(num_div)
+{
+  int32_t left = (int32_t)AS_NUM(args[0]);
+  int32_t right = (int32_t)AS_NUM(args[1]);
+  RETURN_NUM(left / right);
```
munificent: Like `decimal`, this will also do the wrong thing on > 32-bit values.

That may be OK, if this is what the user wants. I guess it's a question of is this for forced-32-bit integer divide, or just "how many times does the RHS completely divide the LHS?" My intuition is the latter in which case I think you want to just divide the args as doubles and then use `fmod()` to truncate the result.

src/wren_core.c
```diff
@@ -888,6 +906,26 @@ DEF_PRIMITIVE(num_isNan)
   RETURN_BOOL(isnan(AS_NUM(args[0])));
 }
 
+DEF_PRIMITIVE(num_rad)
```
munificent: Ditto for `deg`. I don't feel this is that useful.

src/wren_core.c
```diff
@@ -888,6 +906,26 @@ DEF_PRIMITIVE(num_isNan)
...
+{
+  RETURN_NUM(floor(AS_NUM(args[0]) / 57.2957795130823208768));
+}
+
+DEF_PRIMITIVE(num_sign)
+{
+  double value = AS_NUM(args[0]);
+  double sign;
+  if (value > 0) {
```
munificent: Style nits: Can you put the `{` on the next line?

src/wren_core.c
```diff
@@ -888,6 +906,26 @@ DEF_PRIMITIVE(num_isNan)
...
+    sign = 1;
+  } else
+  if (value < 0) {
```
munificent: ...and put this on the same line as the `else`?

src/wren_core.c
```diff
@@ -888,6 +906,26 @@ DEF_PRIMITIVE(num_isNan)
...
 
+DEF_PRIMITIVE(num_rad)
+{
+  RETURN_NUM(floor(AS_NUM(args[0]) / 57.2957795130823208768));
+}
+
+DEF_PRIMITIVE(num_sign)
+{
+  double value = AS_NUM(args[0]);
+  double sign;
```
munificent: Instead of using a temp variable, I'd probably just return directly from within each branch.

src/wren_core.c
```diff
@@ -928,6 +966,11 @@ DEF_PRIMITIVE(num_toString)
   RETURN_VAL(wrenNewString(vm, buffer, length));
 }
 
+DEF_PRIMITIVE(num_truncate)
+{
+  RETURN_NUM((int32_t)(AS_NUM(args[0])));
```
munificent: `fmod()` is your friend here too. :)

test/number/div.wren
```diff
@@ -0,0 +1,5 @@
+IO.print((8).div(2))   // expect: 4
+IO.print((8.5).div(2)) // expect: 4
+IO.print((9).div(2))   // expect: 4
+IO.print((9.9).div(2)) // expect: 4
+IO.print((10).div(2))  // expect: 5
```
munificent: More stuff to test:

* Large number
* Zero LHS.
* Zero RHS.
* Floating point RHS.
* Negative arguments.

test/number/sign.wren
```diff
@@ -0,0 +1,6 @@
+IO.print(123.sign)      // expect: 1
+IO.print((-123).sign)   // expect: -1
+IO.print(0.sign)        // expect: 0
+IO.print((-0).sign)     // expect: 0
```
munificent: Good corner case test. Is this the result as user would expect, though?

test/number/truncate.wren
```diff
@@ -0,0 +1,8 @@
+IO.print(123.truncate)      // expect: 123
+IO.print((-123).truncate)   // expect: -123
+IO.print(0.truncate)        // expect: 0
+IO.print((-0).truncate)     // expect: 0
+IO.print(0.123.truncate)    // expect: 0
+IO.print(12.3.truncate)     // expect: 12
+IO.print((-0.123).truncate) // expect: 0
+IO.print((-12.3).truncate)  // expect: -12
```
munificent: Test a large number here too.

munificent:
I added lots of comments, but thank you for doing these!

src/wren_core.c
MarcoLizza: Yep. I agree, in fact was my first option...

src/wren_core.c
MarcoLizza: Yup. Agreed.

src/wren_core.c
MarcoLizza: In fact I wasn't really sure both `deg` and `rad` are so useful for everybody (due to the fact that they could be easily computed in Wren itself), but due to the presence of the trigonometric methods, I thought it would be funny to write something like
```
270.rad.sin
```
:)

src/wren_core.c
MarcoLizza: Fine.

src/wren_core.c
MarcoLizza: Yep, sorry.

I tend to adhere to the Wren coding-styles but from time to time other projects (and/or personal) style crawls into. :)

src/wren_core.c
MarcoLizza: I tend not to use multiple returns but will fix this. :+1:

test/number/sign.wren
MarcoLizza: According to the definition of sign-function, yes that's what I expect to return. In fact, `0` and `-0` are both, ehm, zero... hence their sign is zero.

What's your opinion?

MarcoLizza:
Got a lot of homework to do. Better roll up my sleeves and do them. :+1:

src/wren_core.c
MarcoLizza: In my plans this was meant to be a integer division, that is the same as
```
lsh.truncate / rhs.truncate
```
and not
```
(lhs / rhs).truncate
```
However, it now occurs to me that since both the division types can be easily obtained in Wren itself (now that we have the `truncate` method) perhaps the `div` method is useless.

Let's keep it simple, as usual, and leave it out. We would be forcing an (arbitrary) integer-division method when one may prefer the other and compute it by himself in Wren.

What do you think?

test/number/sign.wren
munificent: Yeah, I think I'd expect it to follow `signbit()`. Negative zero does feel pretty squishy to me, but IEEE doubles support it and Wren follows that so my normal guideline is to just assume the C standard lib folks have put more thought into this than I have so just follow what they do. :)

MarcoLizza: Agreed, I'll fix this evening...

munificent:
:D

MarcoLizza:
Still missing the sign change to adehere with signbit. Will throw a pull request to change this. :+1:

munificent:
Sounds good. I'd say we should just *use* `signbit()`, but I think it's C99-only so I'm hesitant to depend on that. :confused:

MarcoLizza:
Yep, in fact `signbit` is C99/C++11 only.

Perhaps we could find our way by mean of a union/cast/masking? Sounds too hacky?

munificent:
Yeah, I'm happy with the solution you already did. :)

MarcoLizza:
All rights, let's keep this then! =D

PR #213, opened 2015-03-08
Fix unreachable links for IE11 (edge mode)

RickGlimmer:
This corrects the flow of content. As main did not have a float style it
overlapped the nav tag making the links unreachable in IE11.


munificent:
Thanks!

PR #214, opened 2015-03-08
[RFC] Wren debugger prototype

gsmaverick:
This is a prototype of a basic Wren debugger; as you try to write bigger programs it's nice to be able to play around live at a specific point. Thoughts?

### DIFF
--- a/src/wren_debug.c
+++ b/src/wren_debug.c
@@ -1,6 +1,10 @@
 #include <stdio.h>
+#include <string.h>
 
 #include "wren_debug.h"
+
+// TODO: This is an arbitrary limi; do something smarter.
+#define MAX_READ_LEN 1024
 
 void wrenDebugPrintStackTrace(WrenVM* vm, ObjFiber* fiber)
 {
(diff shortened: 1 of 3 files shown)


MarcoLizza:
Having a debugger is definitely a "must" for bigger Wren use-cases.

I put some thoughts on it, from time to time (and certainly @munificent did to a greater degree) and I think that the debugger should be implemented by means of a debugging API (just like Lua and the others does).

munificent:
Neat!

I definitely agree having some kind of live debugging experience is helpful. My current thought is that we'd probably want it as a separate core library (i.e. `wren_debug.c` or `wren_meta.c`, etc.) That way, users don't have to include it in their shipped apps if they don't want it.

There's a lot of other introspection-type stuff that may make sense to include in there. Things like `eval()` and other primitives we might need to write good REPL mostly in Wren.

The tricky part is balancing that with keeping the core interpreter simple and memory-light. For example, it would be really nice to have access the values of local variables in a debugger, but the compiler throws away the information we'd need to reconstitute that. There are solutions for this, of course (stack maps). We just have to implement them in such away that users who don't want to have the debugger enabled don't pay the price for them.

Being able to execute code in a given context (inside some method) is particularly hard for Wren. It's something other dynamic languages can do because they treat scopes more as dynamic maps of values so things like variable *names* are available at runtime. But that's also why those languages are slower than Wren. :)

I think debugging is *really* important, but I'm not keen on sacrificing speed or simplicity to get it. So we have to see how we can layer debugging on top in an optional way.

gsmaverick:
Some really great feedback and ideas. In that vein what do you think is a good MVP that we/I should try and tackle first (or at least experiment with)?

MarcoLizza:
>  [...] but the compiler throws away the information we'd need to reconstitute that.

We should have the Wren corresponding to a "debug build". :dancer: 

*( just kidding... or not? )*

Anyway, I second not to sacrificate Wren simplicity.

The single most useful debugging feature in my opinion is statement execution callback (and to some degree inspection). Should we start with this or something else?

munificent:
> In that vein what do you think is a good MVP that we/I should try and tackle first (or at least experiment with)?

Good question. If I were putting time into this, I think I'd probably start with making the REPL itself not totally lame (which it pretty much is right now). One missing piece for that is the ability to eval() a chunk of code and get the result back. That has to be a primitive method (not a foreign one) because the FFI can't expose Wren objects.

My plan for this was to make a third built-in library called "meta" that contains this stuff. If that's enabled, you get a `Meta` class that gives you access to meta-programmy stuff like eval. So maybe start with that? Create `wren_meta.c` and add a primitive static `eval` method on on a `Meta` class?

I'm not sure if you'll run into trouble implementing that or not, but give it a try and see how it goes. How does that sound?

munificent:
> We should have the Wren corresponding to a "debug build".

More like a "debug*ger*" build, but, yeah that's pretty much what I'm thinking. Users should be able to opt in or out of all of the extra infrastructure for introspection, debugging, etc. Disabling that is good for security, performance, and memory usage. But enabling it is certainly nice to figure out what your program is doing. :)

gsmaverick:
Meta library PR is here https://github.com/munificent/wren/pull/233

PR #216, opened 2015-03-09
If module file can't be loaded try to load it as a module folder instead

gsmaverick:
This is a quick win that is self-contained in `main.c` to make module loading a little more powerful.

### DIFF
--- a/src/cli/io.c
+++ b/src/cli/io.c
@@ -49,12 +49,12 @@ void setRootDirectory(const char* path)
   rootDirectory = path;
 }
 
-char* readModule(WrenVM* vm, const char* module)
+char* wrenFilePath(const char* name)
 {
   // The module path is relative to the root directory and with ".wren".
   size_t rootLength = rootDirectory == NULL ? 0 : strlen(rootDirectory);
-  size_t moduleLength = strlen(module);
-  size_t pathLength = rootLength + moduleLength + 5;
+  size_t nameLength = strlen(name);
+  size_t pathLength = rootLength + nameLength + 5;
   char* path = (char*)malloc(pathLength + 1);
 
   if (rootDirectory != NULL)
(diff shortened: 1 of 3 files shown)


soveran:
It looks like the same code is duplicated ([here](https://github.com/gsmaverick/wren/blob/f2f42b4cc75e20ba442edb6a208609a0e54d43a9/src/main.c#L56-L67) and [here](https://github.com/gsmaverick/wren/blob/f2f42b4cc75e20ba442edb6a208609a0e54d43a9/src/main.c#L73-L81)). An alternative way to solve this problem may be, for example, to call `readModule` with `"foo"` as the module name, and if it fails you could try `readModule` again but with `"foo/index"` as the module name, and that way you may avoid code duplication. Aside from that, at first I thought the idea of this patch was to address the problem of vendored libraries, but now I think it covers a different use case. Would you mind expanding on the use case for this change?

munificent:
Sorry for the long delay. Real life got busy. :)

Right now module loading is super primitive in the command-line VM and it definitely needs some love so thanks for looking into this. The idea here is that it lets you have a hierarchy of modules where a parent can be both a module and a container of submodules, right? Like:

```
foo/
  index.wren
  bar.wren
```

Which lets you do both:

```dart
import "foo"
import "foo.bar"
```

I'm guessing you want `foo/index.wren` instead of just `foo.wren` so that everything is contained inside the `foo` directory? I'm not *too* crazy about using "index" for this, for reasons I can't fully justify. :)

What do you think of using the directory name? So `import "foo"` would look for `foo.wren` and then `foo/foo.wren`? This would prevent import paths like "foo.foo", but I don't know if that's useful anyway.

Thoughts?

gsmaverick:
Hmm, I guess I would see those options as equivalent with the caveat that it seems that other programming languages normally have a "reserved file" for this case. Python with its `__init.py__` or Scala's `package.scala`. That being said I don't have a strong case for using `index.wren` or any other reserved name in favour of the `foo.wren` pattern.

I would agree that a pattern like `import "foo/foo" doesn't seem very useful.

munificent:
Oh, crap, sorry. I broke this PR under you when I moved some stuff around. :(

What do you think of `module.wren`? I like the idea of a canonical name, but `index` seems wrong to me, since it doesn't describe or list the other modules in the directory.

bjorn:
> Oh, crap, sorry. I broke this PR under you when I moved some stuff around. :(

Well, you broke all pull requests with that. :-)

Fortunately git is pretty smart about applying diffs at the right place when merging/rebasing, though it's a bit annoying that it won't work for newly added files of course. No problem, anyway.

gsmaverick:
@munificent Pushed up a new version that doesn't have conflicts and loads from `module.wren` inside of a package.

munificent:
Woo!

I tweaked it just a little bit. Mainly I avoided using the word "package" because to me that likely implies something larger like a versioned ball of stuff. If Wren ever gets a package manager, I'd like to reserve the word for that. :)

PR #218, opened 2015-03-11
Range privitive count

### LINKED ISSUE #211: Make Range implement count
Right now, it inherits `count` from Sequence, which is really slow—`O(n)`. It should be able to calculate and implement it directly in C.

MarcoLizza:
This should close #211...

### DIFF
--- a/test/range/count.wren
+++ b/test/range/count.wren
@@ -1,0 +1,25 @@
+// Ordered range.
+IO.print((2..5).count) // expect: 4
+IO.print((3..3).count) // expect: 1
+IO.print((0..3).count) // expect: 4
+IO.print((-5..3).count) // expect: 9
+IO.print((-5..-2).count) // expect: 4
+
+// Backwards range.
+IO.print((5..2).count) // expect: 4
+IO.print((3..0).count) // expect: 4
+IO.print((3..-5).count) // expect: 9
+IO.print((-2..-5).count) // expect: 4
+
+// Exclusive ordered range.
+IO.print((2...5).count) // expect: 3
+IO.print((3...3).count) // expect: 0
+IO.print((0...3).count) // expect: 3
+IO.print((-5...3).count) // expect: 8
+IO.print((-5...-2).count) // expect: 3
+
+// Exclusive backwards range.
+IO.print((5...2).count) // expect: 3
+IO.print((3...0).count) // expect: 3
+IO.print((3...-5).count) // expect: 8
+IO.print((-2...-5).count) // expect: 3
--- a/src/wren_core.c
+++ b/src/wren_core.c
@@ -1215,6 +1215,19 @@ DEF_PRIMITIVE(object_instantiate)
   RETURN_ERROR("Must provide a class to 'new' to construct.");
 }
 
+DEF_PRIMITIVE(range_count)
+{
+  ObjRange* range = AS_RANGE(args[0]);
+  if (range->isInclusive)
+  {
+    RETURN_NUM(fabs(range->to - range->from) + 1.0);
+  }
+  else
+  {
+    RETURN_NUM(fabs(range->to - range->from));
+  }
+}
+
 DEF_PRIMITIVE(range_from)
 {
   ObjRange* range = AS_RANGE(args[0]);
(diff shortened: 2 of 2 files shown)


test/range/count.wren
```diff
@@ -0,0 +1,25 @@
...
+IO.print((3...3).count) // expect: 0
+IO.print((0...3).count) // expect: 3
+IO.print((-5...3).count) // expect: 8
+IO.print((-5...-2).count) // expect: 3
+
+// Exclusive backwards range.
+IO.print((5...2).count) // expect: 3
+IO.print((3...0).count) // expect: 3
+IO.print((3...-5).count) // expect: 8
+IO.print((-2...-5).count) // expect: 3
```
munificent: Can you add some tests for floating point ranges too? Iterating over them isn't that common, but they are well-defined so should have the right behavior here too.

Thanks!

MarcoLizza: I will surely do...

... but first I have to grasp how they are defined for non integer values! :wink:

bjorn:
Don't forget the documentation. :-)

test/range/count.wren
MarcoLizza: @munificent, please what results would you expect from following ones?
```
IO.print(((0.0)..(1.9)).count)
IO.print(((0.0)...(1.9)).count)
```

bjorn: I know you're not asking me, but shouldn't the result just be the same as what it currently is without the `Range.count` implementation? `Sequence.count` should already be giving you the answer (in this case, it's 2 in both cases because 2 is never part of the range, whether the 1.9 is "inclusive" or not, so in both cases the range iterates just 0 ad 1).

MarcoLizza: @bjorn, I was already using the current implementation as a reference.

I was asking because (my fault?) I cannot really decide what should be the correct behaviour/result of the above statements.

More precisely, how can we explain those result (with the current implementation)?
```
IO.print(((3.1)..(6.1)).count) // expect: 4
IO.print(((3.1)...(6.1)).count) // expect: 3
IO.print(((0.0)..(1.9)).count) // expect: 2
IO.print(((0.0)...(1.9)).count) // expect: 2
```
Any help is appreciated.

bjorn: I think to explain you just need to look at the values. `Range` will iterate with a step size of 1, starting from the `from` value and ending at the `to` value (either inclusive or exclusive).

So a range of `(3.1)..(6.1)` goes through the values `3.1, 4.1, 5.1, 6.1`. The last value here is the `to` value, but since the range is "inclusive" this value is still considered part of the range. In contrast `(3.1)...(6.1)` is "exclusive" so once it sees 6.1 it will consider its range exhausted and stop iterating, yielding only the values `3.1, 4.1, 5.1`.

The way to calculate the number of elements in the range could be:
```c++
double delta = std::abs(to - from);
double count = std::floor(delta);
if (exclusive && delta == count)
    count -= 1.0;
```
Since whenever `delta` is not a whole number, it does not matter whether the range is exclusive or inclusive because the last value is never hit anyway. But when the delta is a whole number, it will hit exactly the last value and then the exclusive range is one element shorter.

bjorn: I had to edit my code snippet because my initial version was off by 1.

MarcoLizza: I agree with you analysis (and my code is quite similar to yours), but I'm still getting some weird results.

For example, the statement `((3.1)..(6.1)).count` with your code would return `3`, *but* the current implementation returns `4`.

I suspect this is due to floating point misrepresentation issues...

bjorn: Arg, you're right, double can't do this because it is not precise when fractions are involved.
```lua
print(6.1 - 3.1, math.floor(6.1 - 3.1), 3.0, math.floor(3.0))
-- output: 3.0	2	3.0	3
```
Now I wonder if implementing this correctly with fractions is possible at all. :-/

MarcoLizza: > Now I wonder if implementing this correctly with fractions is possible at all. :-/

That's my point. We could implement count as a (primitive) iteration, but it's complexity would be O(n). Or we could drop/change the the current behaviour and (let's say) truncate/round the bounds to integer prior calculation (of course this would mean the the current behaviour won't be preserved).

Thoughts?

MarcoLizza: @munificent (and everyone else interested), do you have any thoughts about this? I would like to approach and close it but I fear in choosing a personal/bad approach. :wink:

munificent: Sorry for the long delay! Too many things to try to stay on top of.

I *think* you can get around most of the precision problems here with a bit of judicious `+ 0.5` usage.

munificent:
I tried to resurrect this using various Git shenanigans, but wasn't able to get it to a mergable point that I could build off of. I'm going to (sadly) close this out.

PR #220, opened 2015-03-14
Added Sequence.count(predicate)

bjorn:
Returns the number of elements in the sequence that pass the `predicate`.

It could also have been implemented as:

    count(f) { reduce(0) {|a, b| f.call(b) ? a + 1 : a } }

But I considered the simple version more readable.

Also documented `Sequence.count`.


munificent:
Cool idea. Do you think it's particularly useful? I can't recall wanting this very much when I've worked with collections.

bjorn:
> Do you think it's particularly useful? I can't recall wanting this very much when I've worked with collections.

I've needed the function in the past week (while coding in Lua, but still) so I definitely think it is useful. Just consider the alternatives if you get into the situation where you could use this function, and you'll be glad it exists when you do.

Quite short would be `list.where { ... }.count`, but this is wrong in terms of efficiency. :-)

Also it's pretty common:

Ruby: http://ruby-doc.org/core-2.2.1/Array.html#method-i-count
C#/Linq: https://msdn.microsoft.com/en-us/library/vstudio/bb535181%28v=vs.100%29.aspx

bjorn:
Rebased to latest master to resolve merge conflict.

munificent:
> Quite short would be list.where { ... }.count, but this is wrong in terms of efficiency. :-)

If we're moving to lazy collections, do you think this is still worth adding?

bjorn:
Well, not in terms of memory usage, but it's still a nice shortcut and I think one that is expected to be available (see references to Ruby and C#).

munificent:
Fair enough. :)

bjorn:
Yay! :-)

PR #221, opened 2015-03-14
Additional Num methods

bjorn:


munificent:
I dig it. How about some tests?

bjorn:
I'll look at adding some tests.

One function that is still missing which is quite useful is `atan2` (especially in games, @MarcoLizza, since it's used for calculating the angle of a line with arbitrary start and end points ;)). I wonder a bit how to add that one, since it takes x and y parameters. Would the following be alright?

```dart
var x = 3
var y = 4
var atanYdividedX = y.atan(x)
```

The C function goes like this:

```c++
double atan2(double y, double x);
```

bjorn:
@munificent Alright, I've added `Num.pi` since this made some well-known test cases easier to write, and then added some tests. I've also added support for `atan2` as described above.

test/core/number/cos.wren
```diff
@@ -0,0 +1,6 @@
+IO.print(0.cos)             // expect: 1
+IO.print(Num.pi.cos)        // expect: -1
+IO.print((2*Num.pi).cos)    // expect: 1
```
munificent: Nit: Can you add spaces around binary operators here and below?

bjorn: Ah, sure. I don't know what was wrong with me when I wrote that. :-)

munificent:
`atan2()` is great. I've always found it infinitely more useful than `atan()`.

bjorn:
Updated to fix code style.

test/core/number/cos.wren
bjorn: Are you really sure we can't rely on precise results? I would expect the standard functions on double to be pretty strictly defined, though if not we should of course do a fuzzy compare.

But maybe it can wait until there is actually a test framework? I find this `IO.print` line harder to reach in case of manual testing, and it makes it much less obvious what the problem might be if this test fails.

MarcoLizza: > Are you really sure we can't rely on precise results?

The floating-point representation itself is non-precise by definition (`pi` itself won't be represented exactly, not being an internally detected constant that changes/fixes the trigonometric functions result).

Moreover, some algorithms are more (or less) stable than others and the result diverges more (or less). Using an *epsilon* threshold is a common practice in numerical analysis.

bjorn: Of course I realize doubles have limited precision. With "precise results" I meant that we should be able to expect exactly the same outcome each time we run the application. Maybe I should have called it "consistent results" instead. If the results are consistently the same imprecise numbers, then there is no need to resort to fuzzy comparison in a test.

So what I was basically asking, is whether we can expect these functions to return different results on different platforms, or whether the standard is strict enough to ensure that it's the same imprecise result everywhere.

munificent:
Awesome, thanks!

PR #222, opened 2015-03-15
Moved List.contains to Sequence.contains, added tests and docs

bjorn:
This makes it work on a `Range` as well, though it should get a more optimal override there.


munificent:
Beautiful, thank you!

PR #223, opened 2015-03-15
Fixed the source locations used by script/generate_builtins.py

bjorn:


munificent:
Oof, good catch. Thanks!

PR #224, opened 2015-03-15
Reverse the argument order of List.insert

bjorn:
The previous order, insert(element, index), was counter-intuitive. I'm not aware of any list API that uses this order. I've checked:

* Ruby `Array.insert(index, obj...)`
* JavaScript `array.splice(start, deleteCount[, item1[, item2[, ...]]])`
* C++ / `QList::insert(int i, const T & value)`
* C++ / `std::vector::insert`
* Lua `table.insert (list, [pos,] value)`
* C# `List<T>.Insert(int index, T item)`
* Java `Interface List<E>.add(int index, E element)`
* Python `list.insert(i, x)`

So it seemed to me more like an oversight in Wren.

### DIFF
--- a/doc/site/core/list.markdown
+++ b/doc/site/core/list.markdown
@@ -17,7 +17,7 @@ Removes all items from the list.
 
 The number of items in the list.
 
-### **insert**(item, index)
+### **insert**(index, item)
 
 **TODO**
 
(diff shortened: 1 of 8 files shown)


munificent:
Ah, good idea. I think I inherited this order from a previous language with named arguments, so you'd do `list insert("element", at: index)`, but I agree matching other languages makes more sense.

PR #226, opened 2015-03-16
Fix Visual Studio project so it builds

bncastle:
-Fixed source file locations for Visual Studio Solution
-Modified .gitignore to ignore Visual Studio cache files
-Added new include path to the project


MarcoLizza:
Nice! I was going to fix it today. Will check this evening if it builds correctly, too.

munificent:
Thanks!

PR #227, opened 2015-03-16
Visual Studio Project Modifications

bncastle:
- Visual Studio: Fixed project to work with changes made in aedf9a85715aae9cf188108fd3339a248dbdb7f3 and 91a23da5b0d6e38578c895cafccf9518fb698459
- Visual Studio: Added another project to the solution that builds Wren as a static library, and modified the original VS project to use it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -31,3 +31,7 @@ ipch/
 *.user
 *.userosscache
 *.sln.docstates
+
+# Build results
+[Dd]ebug/
+[Rr]elease/
--- a/project/msvc2013/wren.sln
+++ b/project/msvc2013/wren.sln
@@ -1,9 +1,13 @@
-
 Microsoft Visual Studio Solution File, Format Version 12.00
 # Visual Studio 2013
 VisualStudioVersion = 12.0.31101.0
 MinimumVisualStudioVersion = 10.0.40219.1
-Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "wren", "wren/wren.vcxproj", "{EBF43135-4A7A-400A-8F23-DF49907025AA}"
+Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "wren", "wren\wren.vcxproj", "{EBF43135-4A7A-400A-8F23-DF49907025AA}"
+	ProjectSection(ProjectDependencies) = postProject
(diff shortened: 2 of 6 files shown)


MarcoLizza:
As a sidenote, I left the shared/static library builds off the MSVC solution because it was simply meant to build the REPL.

In my opinion (and that's my case) when including Wren in a bigger project as a scripting-language, the whole source set should be included/built and not the static library. However, this is to some degree a matter of personal preference (altough I find it easier) and one could find the static library useful.

bncastle:
Hmm yes I do see your point.

What I like about this Visual Studio solution config:
- Separates out the core Wren code from the REPL code. To me it feels cleaner and easier to work with especially as the REPL grows (recently became 3 .c files) and is refactored possibly into more files.
- Provides a static lib for those who want one.
- The Wren source files can still be pulled into another project and built if desired.

I have and still do it as you described with some of my projects (not Wren-related) and I'm not sure which one I prefer myself.

MarcoLizza:
> Honestly, I'm not sure which method I prefer.

Visual Studio solutions are quite powerful, also in terms of dependencies and build order.

For (very) big projects, when structured properly, having sub-project based upon static libraries can speed the build time quite a lot (especially during the development/debug phase).

As a personal matter of taste, I prefer to adopt a static-library oriented approach when the library need to be reused across the product. If the library is used just in the main engine, I will include the source files themselves.

*(potentially the engine could become a library itself but not the scripting-language, for example)*

However, back to the current context. having a static library is also a clear hint on which files to use when including Wren as a scripting language.

.gitignore
```diff
@@ -30,4 +30,8 @@ ipch/
 *.suo
 *.user
 *.userosscache
-*.sln.docstates
\ No newline at end of file
+*.sln.docstates
+
+# Build results
+[Dd]ebug/
+[Rr]elease/
```
munificent: Is it possible to change the solution to build into `build/vs`? I have the XCode project on Mac building into `build/xcode`. This ensures build outputs don't clutter up the top level for every platform.

bncastle:
Added an additional commit that changes Visual Studio's build output location to build/vs.

munificent:
Groovy, thanks! Feel free to throw me a pull request that adds yourself to the AUTHORS file too, if you'd like. :)

Issue #155, opened 2015-02-01
Make Fibers inherit Sequence and be iterable


gsmaverick:
Currently iterating over Fibers (especially when using them to represent streams, generators, etc) is quite verbose and the resulting code doesn't read naturally.

### Ideal Use Case

```dart
var counter = new Fiber {
  var count = 0
  while (true) {
    Fiber.yield(count)
    count = count + 1 
  }
}

for (i in counter) {
  IO.print(i)
}
```

The one outstanding question is whether this iterator invokes the `call` or `try` method on the Fiber being iterated. My money would be on `call` since it's the more straightforward way of implementing this. Thoughts?

munificent:
Yeah, I agree `call` is what you want. This has been something I've wanted to do for a long time.

One thing you'll find that's tricky—which is why I haven't gotten around to this yet—is figuring out how to handle the bit of code the fiber executes after the last `yield`. Consider:

```dart
var fiber = new Fiber {
  for (i in 1..3) Fiber.yield(i)
}

while (!fiber.isDone) {
  IO.print(fiber.call)
}
```

You'd think this prints `1`, `2`, `3`, but it actually prints `1`, `2`, `3`, `null`. The last null is because after yielding `3`, the fiber isn't done yet. It still has a tiny bit of code in can execute. There could actually be lots of code after the last yield:

```dart
var fiber = new Fiber {
  for (i in 1..3) Fiber.yield(i)
  IO.print("Whoa more stuff here")
}
```

That means if you treat fibers as sequences that you iterate over until they're done, you'll always get this lame extra value. I haven't spent the time to figure out a good solution for that yet. If you'd like to throw some brainpower at it, I would definitely like to hear what you can come up with.

In my last language, Magpie, I did make fibers extend the base iterable class because I think this is super powerful. It lets you call `map`, and `where` and stuff directly on fibers. Fun! We just have to figure out how to handle that bit of dangling code at the end.

gsmaverick:
How do you feel about having a primitive value to represent that a Fiber has no more values to yield? I have personally found the dangling `null` value you get from calling yield a little bit strange and I feel like it would make for nicer code to have some "value" represent the state of a Fiber that has yielded all of its values.

This is roughly what Python does in its generator implementation; when you call `next()` on a generator that has run out of values it throws a [StopIteration exception](https://docs.python.org/3/library/exceptions.html#StopIteration) to signal this.

munificent:
[...]

PR #228, opened 2015-03-16
Used deferred execution in Sequence methods

bjorn:
Here goes a proposal along with the suggested changes.

Methods are `Sequence.where` and `Sequence.map` are changed to use deferred execution. They return an instance of a new Sequence-derived class that performs the operation while iterating. I've also added `Sequence.skip` and `Sequence.take` in similar fashion.

In order to still provide a way to easily get a new list with the result, `Sequence.toList` was added.

The advantage of this approach is that it avoids the allocation of intermediate lists. The following example allocates 4 small Sequence-derived classes, but only a single list:
```dart
IO.print((1..100000000).
         where { |i| i % 2 == 0 }.
         map { |i| i*i }.
         skip(1).take(3).toList())

// expect: [16, 36, 64]
```
I read @zeckalpha and @munificent say it would be nice if `map` and `where` were lazy while they discussed making fibers iterable (#155), but this of course changes it for all sequences.

C#/LINQ takes this deferred execution approach.


bjorn:
Note that I'm not personally convinced that this is a good idea, mainly since the usage of these methods does get a little more complicated due to the required `toList` to turn it into a list. Secondly this adds possibly confusing methods like `skip` and `take` to `Range` and `List` classes.

I could imagine this alternative approach:

```dart
IO.print(new Query(1..100000000).
         where { |i| i % 2 == 0 }.
         map { |i| i*i }.
         skip(1).take(3).toList())

// expect: [16, 36, 64]
```

Here, a new `Query` class is added, which would derive from `Sequence` and overrides/adds methods in this style. I think `Sequence.toList` would still be useful to have.

munificent:
I definitely believe that in a general-purpose language, higher-order functions on sequences are more useful and efficient if they're lazy. (For example, my day job is working on Dart, which does this.) I shied away from doing that in Wren because I thought they would add too much weight to the core library.

But, looking at the PR, it's actually pretty compact. The way the iterator protocol works in Wren actually turns out to be pretty helpful here. (In other languages, you have to define both an `__Iterable` and `__Iterator` class for each higher-order method.) Not to pat myself on the back *too* much, but I think it's pretty cool that the MapSequence is only twelve lines of code. :smile: 

I'm sold.

Landing this will affect a lot of things:

* We'll need to rearrange the tests—which has kind of been needed for a while—so that most of the tests for the methods on Sequence live under `test/core/sequence`. We should test the higher-order functions without implicitly calling `toList()` on them to ensure the immediately returned object (the lazy sequences) are directly tested.

* The docs will need to be updated.

* Then there's some new methods like `take()`, `skip()`, and `toList()` here that all need tests and docs.

Would you mind breaking this into a few separate PRs? My free time is pretty limited and it's easier for me to land a couple of small patches a day then one giant one. I'd probably sequence it out something like:

1. Add `toList` as a method on Sequence. (Let's not require `()` since it doesn't mutate the receiver). Maybe call it just `list` like the verb "to list something"? Docs, tests, etc.

2. Make `where()` lazy. Move and update its tests and docs.

3. Same for `map()`.

4. Add `take()`.

5. Add `skip()`.

What do you think? I'm super excited about this. I'd always felt making these methods eager was a hacky compromise that I had to do for simplicity, but I much prefer the lazy behavior.

bjorn:
> The way the iterator protocol works in Wren actually turns out to be pretty helpful here. (In other languages, you have to define both an __Iterable and __Iterator class for each higher-order method.) Not to pat myself on the back too much, but I think it's pretty cool that the MapSequence is only twelve lines of code. :smile:

I think Lua has a very nice approach to iterators as well, which requires only a single function call for iterating and getting the value. But, it relies on multiple return values and can't deal with `null` values. I agree anyway that Wren's approach is quite comfortable too. :)

> Add toList as a method on Sequence. (Let's not require () since it doesn't mutate the receiver). Maybe call it just list like the verb "to list something"? Docs, tests, etc.

I agree simply `list` is more comfortable than `toList()`, so I'll gladly rename it. I was not aware of the mutation being the relevant bit in deciding whether to put a `()` or not, but it sounds reasonable.

> Would you mind breaking this into a few separate PRs? My free time is pretty limited and it's easier for me to land a couple of small patches a day then one giant one.

I can certainly split it up into multiple PRs. Personally I thought it made more sense to do it in one since it would be strange to leave Wren into an inconsistent state (where `map` is lazy and `where` isn't, for example) and also there will generally always be merge conflicts between these smaller pull requests.

My spare time is also severely limited though, so it will take a bit.

> What do you think? I'm super excited about this. I'd always felt making these methods eager was a hacky compromise that I had to do for simplicity, but I much prefer the lazy behavior.

Well, I'm glad you are also excited about the lazy behavior! Like I said, my only gripe is the need for the `.list` at the end when you need it, but that may not be so often and it may be easy to get used to. I've not personally used this lazy behavior a lot yet actually. :-)

munificent:
> My spare time is also severely limited though, so it will take a bit.

Totally understandable. If one big PR is easier for you, that's fine by me.

bjorn:
> Totally understandable. If one big PR is easier for you, that's fine by me.

I'll split it up in multiple RPs as follows:

* Adding `Sequence.list`
* Changing `Sequence.where` and `Sequence.map` to be lazy
* Adding `Sequence.skip` and `Sequence.take`

This is mostly because those are chunks that make sense to me. The note about my limited spare time is not because I didn't want to split it up, but just a general warning. :-)

munificent:
That sounds perfect to me.

> The note about my limited spare time is not because I didn't want to split it up, but just a general warning. :-)

Haha, that's OK. I know exactly what having limited time is like.

munificent:
Closing this out since you'll be breaking it into separate PRs. :)

PR #229, opened 2015-03-18
Added a check for allready freed in wrenFreeVm.

verpeteren:
If wrenFreeVm was called twice, this would lead to an error, because the memory was allready freed.
The used check is rather naive, suggesting that a valid vm does have at least one methodName set.


munificent:
Thanks!

PR #230, opened 2015-03-18
Changed int-fields in ObjList to match the uint32_t-types in ObjMap

verpeteren:
Both ObjList and ObjMap have the same fields (capacity and count). Therefor it makes sense to make them the same type.


munificent:
I had to make some more changes in core to make sure all of the signed/unsigned conversions were handled correctly, but thanks for getting this going!

PR #231, opened 2015-03-18
Added -Wextra to compile flags

verpeteren:
-Wextra does several extra checks during the compiling. Unfortunately one of the extra check is for unused parameters.
Several functions have unused parameters (mostly the WrenVM* vm).


munificent:
:+1:

PR #232, opened 2015-03-22
[WIP - need some help] Automatically print the result of expressions

patriciomacadden:
Hi @munificent!

First of all let me congratulate you for your work, I really like wren.

This is my attempt to automatically print the result of expressions. It basically consists of passing a struct  containing the status and the string value of the expression (which will be printed in the REPL). I renamed `wrenInterpretResult` to `wrenInterpretResultStatus` and made `wrenInterpretResult` such struct, containing both `wrenInterpretResultStatus` and a string (the result of the expression).

So far so good, all tests passes, but I can't make it return a c string from the result of the expressions. I tried this: `interpretResult.value = AS_CSTRING(result);` but didn't work. Obviously I am missing something and I can't figure out what it is. Maybe you can give me a hint or something?

### DIFF
--- a/src/vm/wren_vm.c
+++ b/src/vm/wren_vm.c
@@ -522,8 +522,9 @@ static bool isInstanceOf(WrenVM* vm, Value value, ObjClass *baseClass)
 // The main bytecode interpreter loop. This is where the magic happens. It is
 // also, as you can imagine, highly performance critical. Returns `true` if the
 // fiber completed without error.
-static bool runInterpreter(WrenVM* vm)
+static WrenInterpretResult runInterpreter(WrenVM* vm)
 {
+  WrenInterpretResult interpretResult;
   // Hoist these into local variables. They are accessed frequently in the loop
   // but assigned less frequently. Keeping them in locals and updating them when
   // a call frame has been pushed or popped gives a large speed boost.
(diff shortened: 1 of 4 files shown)


src/vm/wren_vm.c
```diff
@@ -1070,7 +1083,11 @@ static bool runInterpreter(WrenVM* vm)
       if (fiber->numFrames == 0)
       {
         // If this is the main fiber, we're done.
-        if (fiber->caller == NULL) return true;
+        if (fiber->caller == NULL)
+        {
+          interpretResult.status = WREN_RESULT_SUCCESS;
```
patriciomacadden: Here I tried to do this:

```c
interpretResult.value = AS_CSTRING(result); // result's type is Value
```

but it didn't work as I expected.

munificent: Yeah, that's not going to work for two reasons:

1. The result may not actually be a string. It could be any type. `AS_CSTRING` doesn't do any conversion, it just interprets the object's bits as that type.

2. Even in cases where the result *is* a string, you can't expose the raw memory for it outside of the VM. The GC may collect it at any time.

munificent:
Alas, I have set a trap for you by leaving that TODO in the code. Printing results seems like it should be easy, but it touches on the interaction between the VM and the host application, which makes it quite a bit more difficult thank you'd expect.

(For what it's worth, that's why the TODO is still there. *I'm* still not sure how to fix it either.)

patriciomacadden:
Hmmm, then it's harder than I thought. I'll see if I can do something about it.

src/vm/wren_vm.c
patriciomacadden: When you say:

> It could be any type.

you mean that it could be any `Value`? If so, wouldn't `AS_CSTRING` work?

munificent: > you mean that it could be any Value?

Right.

>  If so, wouldn't AS_CSTRING work?

No, unfortunately. That's only valid to use if you already know the value is an `ObjString*`.

munificent:
I'm closing this PR out because I don't think this direction will work, but I am still interested in having you hack on this. Thanks!

patriciomacadden:
No problem!

PR #233, opened 2015-03-22
Adds a Meta library with an eval function for interpreting code inline.

gsmaverick:
This PR adds a simple `Meta.eval(_)` function that accepts a string to evaluate inline. The biggest limitation right now is that you can't define new names inside the evalled source and then use those names afterwards because the compiler assumes those names aren't defined in the source and can't generate the correct bytecode for them.

### DIFF
--- a/src/vm/wren_meta.h
+++ b/src/vm/wren_meta.h
@@ -1,12 +1,8 @@
 #ifndef wren_meta_h
 #define wren_meta_h
 
-#include <stdio.h>
-
 #include "wren.h"
 #include "wren_common.h"
-#include "wren_value.h"
-#include "wren_vm.h"
 
 // This module defines the Meta class and its associated methods.
 #if WREN_USE_LIB_META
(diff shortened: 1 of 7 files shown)


zeckalpha:
:+1:

src/vm/wren_meta.h
```diff
@@ -0,0 +1,18 @@
+#ifndef wren_meta_h
+#define wren_meta_h
+
+#include <stdio.h>
+
+#include "wren.h"
+#include "wren_common.h"
+#include "wren_value.h"
+#include "wren_vm.h"
```
munificent: I don't think `wren_value` or `wren_vm` is used here.

src/vm/wren_meta.h
```diff
@@ -0,0 +1,18 @@
+#ifndef wren_meta_h
+#define wren_meta_h
+
+#include <stdio.h>
```
munificent: Can this be removed?

munificent:
> you can't define new names inside the evalled source and then use those names afterwards

That's the behavior I expect, I think. I believe the REPL works the same way if the variable isn't capitalized.

gsmaverick:
@munificent updated to remove the unnecessary header files.

munificent:
:metal:

Issue #156, opened 2015-02-02
Embedding API: How to store private data in a class instance?


verpeteren:
# Introduction

I read that the embedding is not fleshed out yet. But I am anxious to try this.
While browsing through wren's source, I could not find a way to do this but maybe I am missing something.

# Naive solution

What I want to do is something like this:

## wren 'user' source code
    // untested wren (pseudo) code
    webserver = new Webserver( "localhost", 8888, 60 )
    webserver.addDocumentRoot( "/static/(?<path>.*)", "../var/www/static" )

Looking at a wren.h, the following comes to mind:

## c embedding source code
    // untested c (pseudo ) code
    static void Wn_Webserver_Constructor( WrenVm * runtime ) {
        ObjClass * thisObj  = wrenGetArgumentObject( runtime, 0 );
        const char * hostName = wrenGetArgumentString( runtime, 1 );
        uint16_t portNr = (uint16_t)  wrenGetArgumentDouble( runtime, 2 );
        int timeout = (int)  wrenGetArgumentDouble( runtime, 3 );
        struct webserver_t ws = Webserver_New( hostName, portNr, timeout );
        if ( ws != NULL ) {   
            wrenSetPrivate( thisObj, (void*) ws );
            wrenReturnBool( runtime, true );
            return;
        }
        wrenReturnBool( runtime, false );
    }

    static void Wn_Webserver_AddDocumentRoot( WrenVm * runtime ) {
        ObjClass * thisObj  = wrenGetArgumentObject( runtime, 0 );
        struct webserver_t * ws = ( struct webserver_t *) wrenGetPrivate( thisObj );
        if ( ws != NULL ) {
            const char * pattern = wrenGetArgumentString( runtime, 1 );
            const char * folder = wrenGetArgumentString( runtime, 2 );
            Webserver_AddDocumentRoot( ws, pattern, folder );
            wrenReturnBool( runtime, true );
            return;
        }
        wrenReturnBool( runtime, false );
    }
   
    static void Wn_Webserver_Destructor( WrenVm * runtime ) {
        ObjClass * thisObj  = wrenGetArgumentObject( runtime, 0 );
        struct webserver_t * ws = ( struct webserver_t *) wrenGetPrivate( thisObj );
        if ( ws != NULL ) {
            Webserver_Delete( ws );
            wrenReturnBool( runtime, true );
            return;
        }
        wrenReturnBool( runtime, false );
    }

    void Wren_Run_MyStuff( WrenVM * runtime ) {
[...]

PR #234, opened 2015-03-24
Bind C methods to Wren using foreign declarations.

munificent:
I've been hacking on this for a while and finally got it to a place I like. But, before I merge it in, I want to see what others think. This replaces the old `wrenDefineMethod` and `wrenDefineStaticMethod` functions in the embedding API. Instead, in your Wren code, you declare a method "foreign" (and don't give it a body), like:

```dart
class SomeClass {
  foreign someMethodDefinedInC
}
```

When Wren executes that class body, it invokes a callback you give it (`bindForeignMethodFn`) in your `WrenConfiguration`. Wren passes you the name of the module and class containing the method, as well as the method's full signature. Your job is to then return a pointer to a C function that implements that method.

In other words, foreign methods are *pulled* in by the VM at an appropriate time. I think this is a little cleaner that the old API that basically jammed methods into classes from outside, implicitly creating the class if needed.

You can see how this is used by looking at the IO built-in library.

Passing you the module name should help make this composable when people want to reuse multiple different modules that all define their own foreign methods. (However, the Wren CLI doesn't do anything useful with this capability yet.)

Thoughts?

### DIFF
--- a/src/vm/wren_vm.c
+++ b/src/vm/wren_vm.c
@@ -37,6 +37,7 @@ WrenVM* wrenNewVM(WrenConfiguration* configuration)
   memset(vm, 0, sizeof(WrenVM));
 
   vm->reallocate = reallocate;
+  vm->bindForeign = configuration->bindForeignMethodFn;
   vm->loadModule = configuration->loadModuleFn;
 
   wrenSymbolTableInit(&vm->methodNames);
@@ -62,14 +63,18 @@ WrenVM* wrenNewVM(WrenConfiguration* configuration)
     vm->heapScalePercent = 100 + configuration->heapGrowthPercent;
   }
 
+  ObjString* name = AS_STRING(CONST_STRING(vm, "main"));
+  wrenPushRoot(vm, (Obj*)name);
+
   // Implicitly create a "main" module for the REPL or entry script.
-  ObjModule* mainModule = wrenNewModule(vm);
+  ObjModule* mainModule = wrenNewModule(vm, name);
   wrenPushRoot(vm, (Obj*)mainModule);
 
   vm->modules = wrenNewMap(vm);
   wrenMapSet(vm, vm->modules, NULL_VAL, OBJ_VAL(mainModule));
 
-  wrenPopRoot(vm);
+  wrenPopRoot(vm); // mainModule.
+  wrenPopRoot(vm); // name.
 
   wrenInitializeCore(vm);
   #if WREN_USE_LIB_IO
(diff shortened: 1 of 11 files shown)


MarcoLizza:
I *really* love such a change in perspective! Almost every other scripting language use the other approach and this feels a bit like taking the road less travelled. :wink:

This permits a very high level of run-time method binding/definition, without introducing any notable performance drop. I like that methods are "registered" by pulling them on the declaration to be ready and available to be called when needed.

However, I have some concerns on the clutter in the code it will introduce.

This is two-folded, since

* We would end in having a (very) long chain of `strcmp` calls which I'm not too sure to be happy to have nor too maintain (with a minor drop in performance due to the repeated calls... if the module expose `N` methods, we would end in having `(N*(N+1))/2` string comparisons);

* We would be needed in declaring the `foreign` functions in some kind of "header module" (which is redundant but good for the documentation sake, 'though).

I will think more about it and see if something else pops in my mind. :)

src/vm/wren_vm.c
```diff
@@ -270,26 +275,63 @@ static void closeUpvalue(ObjFiber* fiber)
...
+    fn = vm->bindForeign(vm, moduleName, className, isStatic, signature);
+    if (fn != NULL) return fn;
+  }
 
-  Method method;
-  method.type = METHOD_BLOCK;
-  method.fn.obj = AS_OBJ(methodValue);
+  // Otherwise, try the built-in libraries.
+  #if WREN_USE_LIB_IO
+  fn = wrenBindIOForeignMethod(vm, className, signature);
```
krypt-n: This results in a failed assert if the bindForeign function returns NULL and the class is not "IO".

krypt-n:
>We would end in having a (very) long chain of strcmp calls which I'm not too sure to be happy to have nor too maintain (with a minor drop in performance due to the repeated calls... if the module expose N methods, we would end in having (N*(N+1))/2 string comparisons);

The chained strcmp is not required, any way of mapping a string to a functionpointer works in this approach e.g. a hashmap

MarcoLizza:
> The chained strcmp is not required, any way of mapping a string to a functionpointer works in this approach e.g. a hashmap

Yep, of course. But having a hashmap to be allocated and populated (even if STL/Boost/whatever based) isn't much different in cluttering the code. I would prefer the `strcmp` chain sequence.

bjorn:
I'm mainly wondering why it is called `foreign` as opposed to `native`, but I guess it depends on your point of view. In Java they chose the latter and in general "native" is used to refer to C/C++ or anything that compiles to "native" machine code. On the other hand, FFI is a pretty common terminology as well.

As for the list of `strcmp`, this can of course be optimized also without allocating/populating hashmaps. The code for this could be auto-generated, but it's not very nice for readability.

What I really like about the pull model it is that you end up with a complete class definition in one place.

src/vm/wren_vm.c
munificent: Eek, you're right! I'll fix that.

munificent:
> Almost every other scripting language use the other approach and this feels a bit like taking the road less travelled.

For what it's worth, I modeled this after Java's `native` keyword, so it's not entirely untraveled. :)

> if the module expose N methods, we would end in having (N*(N+1))/2 string comparisons);

Yes, that is a drag. As @krypt-n notes, smart host applications could do something smarter here like a table or hash lookup, though it might be nice to see if there's something we can do to make that easier.

It's worth noting this perf hit is only while the class definition is being executed. Method *calls* are as fast as ever.

> We would be needed in declaring the foreign functions in some kind of "header module" (which is redundant but good for the documentation sake, 'though).

Yeah, I think this is generally a good thing, though. It gives module writers a place to include stuff like doc comments as well as mix in pure Wren implementations of some parts of the class.

>  But having a hashmap to be allocated and populated (even if STL/Boost/whatever based) isn't much different in cluttering the code. I would prefer the strcmp chain sequence.

I'm agnostic. This code lives in the host app, so they can do it however they want. :)

> I'm mainly wondering why it is called foreign as opposed to native, but I guess it depends on your point of view. In Java they chose the latter and in general "native" is used to refer to C/C++ or anything that compiles to "native" machine code.

Yeah, the terminology for both is a bit funny. What you consider "native" or a "foreign" depends entirely on where you're standing. Like you note with "FFI", "foreign" felt a little less ambiguous to me about which perspective it presumes. I also couldn't recall ever wanting to use "foreign" as an identifier in code of mine, so it seemed like a fairly innocuous reserved word.

I considered `extern` too but it felt a bit outdated to me. Also, I think `foreign class` reads well and might be a useful syntax/context for #156 where `extern class` doesn't sound right.

munificent:
> What I really like about the pull model it is that you end up with a complete class definition in one place.

Yes! This was definitely a goal.

MarcoLizza:
> For what it's worth, I modeled this after Java's native keyword, so it's not entirely untraveled. :)

Also, .NET P/Invoke and (from the history back) VB's `Declare` are, to some degree, in the same league... well, it's the road *less* traveled... almost. :)

> It's worth noting this perf hit is only while the class definition is being executed. Method calls are as fast as ever.

Yep, that was clear enough. :+1: 

> Yeah, I think this is generally a good thing, though. It gives module writers a place to include stuff like doc comments as well as mix in pure Wren implementations of some parts of the class.

In my opinion, this point alone is enough to *love* the "foreign pull" approach. I always suffered the lack of a clear documenting place in the embedding layer itself.

> Yeah, the terminology for both is a bit funny.

I do like `foreign`, to be honest. It clearly states that you are referring to something that do no resides into the Wren side of the world.

PR #236, opened 2015-03-26
Add script creating an amalgamation

lluchs:
I.e. it combines everything in a single wren.c file for super easy embedding. See [SQLite](https://www.sqlite.org/amalgamation.html).

The Python script is as simple as possible and makes a lot of assumptions:

 - All static variables and functions are unique across all files.
 - Macros need to be `#undef`ed before they are redefined in a different file (there was only a single case here which could also be renamed).
 - All header files (except wren.h) are in the same directory.

Building the CLI is then possible like this:

```sh
make wren.c
gcc -Os -std=c99 -Isrc/include wren.c src/cli/*.c -o bin/wren -lm 
```

The benchmark results are curious - apparently, some very different optimizations happen with the single file (comparing plain `make` and the build resulting from the two lines above):
```
binary_trees - wren            .......... 0.41s 0.0017 111.40% relative to baseline
delta_blue - wren              .......... 0.20s 0.0007  98.85% relative to baseline
fib - wren                     .......... 0.65s 0.0061  87.50% relative to baseline
for - wren                     .......... 0.23s 0.0019 106.27% relative to baseline
method_call - wren             .......... 0.33s 0.0011 101.85% relative to baseline
map_numeric - wren             .......... 0.77s 0.0019  99.19% relative to baseline
map_string - wren              .......... 0.18s 0.0012 104.65% relative to baseline
string_equals - wren           .......... 0.51s 0.0047 101.89% relative to baseline
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -53,4 +53,9 @@ watchdocs:
 gh-pages: docs
 	@ cp -r build/docs/. build/gh-pages
 
+# Build amalgamation of all Wren library files.
+wren.c: src/include/wren.h src/vm/*.h src/vm/*.c
+	./script/generate_amalgamation.py $^ > $@
+
+.DELETE_ON_ERROR: wren.c
 .PHONY: debug release all clean test builtin docs watchdocs gh-pages
--- a/Makefile
+++ b/Makefile
@@ -55,7 +55,7 @@ gh-pages: docs
 
 # Build amalgamation of all Wren library files.
 wren.c: src/include/wren.h src/vm/*.h src/vm/*.c
-	./script/generate_amalgamation.py $^ > $@
+	./script/generate_amalgamation.py > $@
 
 .DELETE_ON_ERROR: wren.c
 .PHONY: debug release all clean test builtin docs watchdocs gh-pages
--- a/script/generate_amalgamation.py
+++ b/script/generate_amalgamation.py
@@ -1,0 +1,28 @@
+#!/usr/bin/env python
+
+import sys
+import os.path
+import re
+
+INCLUDE_PATTERN = re.compile(r'^\s*#include "([\w.]+)"')
+
+seen_files = set()
+
+def add_file(filename):
+  basename = os.path.basename(filename)
+  # Only include each file at most once.
+  if basename in seen_files:
+    return
+  seen_files.add(basename)
+  path = os.path.dirname(filename)
+
+  with open(filename, 'r') as f:
+    for line in f:
+      m = INCLUDE_PATTERN.match(line)
+      if m:
+        add_file(os.path.join(path, m.group(1)))
+      else:
+        sys.stdout.write(line)
+
+for f in sys.argv[1:]:
+  add_file(f)
--- a/script/generate_amalgamation.py
+++ b/script/generate_amalgamation.py
@@ -1,10 +1,12 @@
 #!/usr/bin/env python
 
 import sys
-from os.path import basename, dirname, join
+from os.path import basename, dirname, join, realpath
+from glob import iglob
 import re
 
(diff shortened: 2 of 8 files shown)


MarcoLizza:
> The benchmark results are curious - apparently, some very different optimizations happen with the single file

That's really not so unexpected, to be honest. Code alignment and peephole optimizations are quite different, from time to time.

munificent:
Oh, crap. Sorry it took so long for me to get to this. In the meantime, I merged in another patch that broke this. :(

From the benchmark results, it isn't a clear win in performance. (But it's really cool that you implemented this so we can *know* that!)

Given that, do you think it's worth landing? We'd have to maintain it henceforth and forever, so I want to make sure it will provide enough value for end users. If so, any chance you could update it to work with the new opcode header?

lluchs:
I think the benefit of this would primarily be the easy inclusion in other projects for people who want to use Wren as a library. I mainly included the benchmark results because the SQLite people claim that the amalgamation improved performance for them.

I guess this isn't very important right now, but may be interesting once Wren gets versioned releases. The recommended way of using Wren could then just be "download this file, add it to your build, done". Adding this later on is likely to be more painful as it's not hard to avoid duplicate names, but annoying to remove them in bulk.

I will update the pull request. When merging this, it would also be important to have the result tested by the CI so that issues are caught immediately.

FWIW the comparison also looks a lot better when using clang (for both builds - when comparing gcc and clang, the clang build was a lot faster anyways):
```
binary_trees - wren            .......... 0.35s 0.0011  95.12% relative to baseline
delta_blue - wren              .......... 0.14s 0.0006 105.72% relative to baseline
fib - wren                     .......... 0.50s 0.0057  96.66% relative to baseline
for - wren                     .......... 0.15s 0.0005 124.67% relative to baseline
method_call - wren             .......... 0.22s 0.0024 111.48% relative to baseline
map_numeric - wren             .......... 0.60s 0.0008 116.36% relative to baseline
map_string - wren              .......... 0.14s 0.0014 110.12% relative to baseline
string_equals - wren           .......... 0.45s 0.0007 104.19% relative to baseline
```

munificent:
> The recommended way of using Wren could then just be "download this file, add it to your build, done". 

Yeah, I think this sounds cool.

> Edit: green means better, right?

Yes!

> I'm actually not sure how to interpret the percentage score now when I think about it...


It shows you that *Wren's speed compared to that language*. This seems kind of backwards, but it makes sense if you assume Wren is the "most important" language to someone running these benchmarks.

So, if for on Lua reports `123%`, it means Wren is 23% faster than Lua on this benchmark.

If the language for the benchmark is Wren that it compares the current Wren speed to the baseline. So this:

```
for - wren                     .......... 0.15s 0.0005 124.67% relative to baseline
```

Means your current code is 24.67% faster than the baseline performance.

> When merging this, it would also be important to have the result tested by the CI so that issues are caught immediately.

Agree totally. The simplest way to do this may just be to make the amalgamation the canonical way to build the VM. We could make the makefile compile the CLI and libraries using the amalgamated source.

lluchs:
I rebased my commits and fixed some new issues. (The old branch is available [here](https://github.com/lluchs/wren/tree/amalgamation1).)

>The simplest way to do this may just be to make the amalgamation the canonical way to build the VM. We could make the makefile compile the CLI and libraries using the amalgamated source.

This would suck for development though, as compilation errors wouldn't refer to the right files. Additionally, it may make compilation slower (compiling the amalgamation with gcc takes about 0,7s here and 0,3s with clang).

munificent:
Super groovy! Sorry it took so long to merge, but it's in now. \o/

PR #239, opened 2015-03-27
Added Sequence.list

bjorn:
This helper method turns any Sequence into a List.

(first part split off from #228, now named simply `list` with added docs and a test)


builtin/core.wren
```diff
@@ -90,6 +90,14 @@ class Sequence {
 
     return result
   }
+
+  list {
+    var result = new List
```
munificent: Nit, but how about just `[]` instead of `new List`?

test/core/sequence/list.wren
```diff
@@ -0,0 +1,11 @@
...
+  iterate(iterator) {
+    if (iterator == null) return 1
+    if (iterator == 3) return false
+    return iterator + 1
+  }
+
+  iteratorValue(iterator) { iterator }
+}
+
+IO.print((new TestSequence).list) // expect: [1, 2, 3]
```
munificent: This is great!

builtin/core.wren
bjorn: That's what I had before, but the Wren compiler not like the `[]` list constructor being used before the `List` class has been mentioned. But at that point it's too late to define `Sequence.list`. I then noticed all the other Sequence methods used `new List` so I used it as well.

munificent: Ah, groovy.

PR #240, opened 2015-03-28
Add list of currently imported modules to meta library.

gsmaverick:
I had to copy the existing wrenReturn[String, Bool, Double] function to return a list from the modules call. Not sure if there's a better way we can share those tools between wren_core, wren_io and wren_meta.


munificent:
Ah, interesting! I have two bits of feedback, one high level and one low level:

1. Is the module list useful for end users? I'm not opposed to it in principle, but I'd like to make sure we only expose API surface area where we're pretty confident it will be useful to people. Everything we add is something we have to test, maintain, etc. so I want to make sure it pulls its weight.

2. The previous `eval` method was implemented using a regular foreign method. However, as you can see, the FFI isn't powerful enough for what Meta is going to need to be able to do. Given that, it probably makes more sense to make it define its methods using primitives, like wren_core.c does, instead of foreign functions. It's probably a good idea to move the `PRIMITIVE`, `DEF_PRIMITIVE`, etc. macros from wren_core.c into wren_common.h.

Thanks for hacking on this!

src/vm/wren_meta.h
```diff
@@ -3,6 +3,7 @@
 
 #include "wren.h"
 #include "wren_common.h"
+#include "wren_vm.h"
```
bjorn: Why add this include in `wren_meta.h` rather than `wren_meta.c`? It looks like it should suffice to have it in the source file.

bjorn:
> 2. The previous `eval` method was implemented using a regular foreign method. However, as you can see, the FFI isn't powerful enough for what Meta is going to need to be able to do. Given that, it probably makes more sense to make it define its methods using primitives, like wren_core.c does, instead of foreign functions. It's probably a good idea to move the `PRIMITIVE`, `DEF_PRIMITIVE`, etc. macros from wren_core.c into wren_common.h.

This raises two questions for me:

- Why are you saying the FFI isn't powerful enough? Am I missing something about future additions to this library? Just wondering since this patch still uses `foreign` for both `Meta.eval` and `Meta.modules`.

- What is the reason for keeping the `PRIMITIVE` macros? I thought the plan was to have the `foreign` replace their use, so what limitation or disadvantage requires this system to be kept around?

munificent:
> Why are you saying the FFI isn't powerful enough?

The FFI (by design) doesn't have access to raw Wren objects or internal VM state.

> Just wondering since this patch still uses foreign for both Meta.eval and Meta.modules.

It does, but it also includes `wren_vm.h` directly and pokes at stuff like assigning directly to `foreignCallSlot` which is fine for a magic library in Wren to do, but certainly not something a normal foreign method should be doing.

> What is the reason for keeping the PRIMITIVE macros?

More of the same: primitives have raw access to the parameter stack, the VM's internal state, the current fiber, etc. They cast objects directly to know primitive types. Primitives can also do things like switch the currently executing fiber and directly generate a runtime error (though we probably will need support for foreign methods doing the latter I suppose).

I definitely like the idea of being able to write *most* things just using the FFI, but there's always going to be some core stuff that's too deeply tied to the VM for that to make sense. I believe Meta will end up being in that bucket.

PR #241, opened 2015-03-28
Added Sequence.each

bjorn:
This is a bit of a style preference since of course you can always write the same thing with a for loop. However, I think sometimes the code looks better when using this method.

It also provides an alternative to `Sequence.map` for cases where you don't need the resulting list, and one that becomes especially necessary when `Sequence.map` is changed to return a new sequence. The example in the `README.md` file was using `Sequence.map` in a way that required this alternative in that case.


test/core/list/each.wren
```diff
@@ -0,0 +1,3 @@
+var words = ""
+["One", "Two", "Three"].each {|word| words = words + word }
+IO.print(words) // expect: OneTwoThree
```
munificent: This is swell. Can you also test:

1. Passing an empty list to it.
2. Passing something that doesn't have a `.call` method.

I know it seems pointless for something so simple, but good test coverage matters a lot.

munificent:
One request, but otherwise I like this. I am a bit iffy about adding a method that's technically redundant with `for in` loop. But my experience is that sequences are some of the most commonly used objects in a language, and a good method set really helps.

test/core/list/each.wren
bjorn: Done.

bjorn:
> I am a bit iffy about adding a method that's technically redundant with `for in` loop. But my experience is that sequences are some of the most commonly used objects in a language, and a good method set really helps.

I fully understand the argument about redundancy, but actually I think this approach fits very well in Wren. It is also nicely familiar for Ruby programmers. You yourself have chosen to use this style in Wren's primary example! :-)

Actually in my opinion the rather strict avoiding of redundancy is one of the reasons many dislike the STL and in contrast why the Qt containers are so pleasant to use in comparison. You find yourself wanting to write helper functions all the time when using STL in order to express yourself in shorter and more understandable ways. Though it's always a balance... in Qt I sometimes find myself wondering whether to use `count()`, `size()` or `length()`, but they're all the same.

munificent:
> Actually in my opinion the rather strict avoiding of redundancy is one of the reasons many dislike the STL and in contrast why the Qt containers are so pleasant to use in comparison.

Right. I like the "there should be one way to do it" philosophy—especially for a minimal language like Wren—but that still leaves it up to us to define how fine-grained we define "it" to be.

But I do like a functional style, especially with sequences which tend to have long method chains. I think it reads better as:

```dart
collection
    .map {|e| e * 2}
    .where {|e| e < 4}
    .each {|e| IO.print(e) }
```

than to have to treat the trailing `each` specially:

```dart
for (e in collection
    .map {|e| e * 2}
    .where {|e| e < 4}) {
  IO.print(e)
}
```

(Of course, there's an existing problem that the above style isn't valid in Wren today because of newline handling, but we'll fix that eventually.)

munificent:
Now that I think about it, there's another compelling reason to have `each`. It lets you work with objects that only support `call(_)` but not the iterator protocol.

bjorn:
> Now that I think about it, there's another compelling reason to have each. It lets you work with objects that only support `call(_)` but not the iterator protocol.

Right, since you pass in a function object, if that's what you have in the first place then the `.each` becomes extra comfortable. Unless I misunderstood you because I don't understand the part about the iterator protocol.

> (Of course, there's an existing problem that the above style isn't valid in Wren today because of newline handling, but we'll fix that eventually.)

Yeah, I noticed that you have to put the `.` at the end like in the example I wrote in #228. Actually having the dot at the end looks a little cleaner to me anyway, so I don't think this is a big issue.

PR #243, opened 2015-03-28
Use deferred execution for Sequence.map and Sequence.where

bjorn:
The methods `Sequence.map` and `Sequence.where` are now implemented using deferred execution. They return an instance of a new Sequence-derived class that performs the operation while iterating. This has three main advantages:

* It can be computationally cheaper when not the whole sequence is  iterated.
* It consumes less memory since it does not store the result in a newly allocated list.
* They can work on infinite sequences.

Some disadvantages are:

* Iterating the returned iterator will be slightly slower due to the added indirection.
* You should be aware that modifications made to the original sequence will affect the returned sequence.
* If you need the result in a list, you now need to call `Sequence.list` on the result.

For correct behavior of the basic Wren example, this change depends on #241 to avoid usage of the lazy `Sequence.map` there.

This is the second part split off from #228.


builtin/core.wren
```diff
@@ -100,6 +88,29 @@ class Sequence {
...
+}
+
+class WhereSequence is Sequence {
+  new(seq, f) {
+    _seq = seq
+    _f = f
+  }
+
+  iterate(n) {
+    while ((n = _seq.iterate(n)) && !_f.call(_seq.iteratorValue(n))) {}
```
munificent: I think this might be a bit clearer as:

```dart
while (n = _seq.iterate(n)) {
  if (_f.call(_seq.iteratorValue(n)) break
}
return n
```

builtin/core.wren
```diff
@@ -100,6 +88,29 @@ class Sequence {
...
+    return n
+  }
+  iteratorValue(iterator) { _seq.iteratorValue(iterator) }
```
munificent: Nit, but can you add a blank line above this?

test/core/sequence/map.wren
```diff
@@ -0,0 +1,45 @@
...
+    if (iterator == null) return new FibIterator
+    iterator.iterate
+    return iterator
+  }
+
+  iteratorValue(iterator) { iterator.value }
+}
+
+var squareFib = (new Fib).map {|fib| fib * fib }
+var iterator = null
```
munificent: Can you add a line here:

```dart
IO.print(squareFib is Sequence) // expect: true
```

And do the same thing for `where()` as well.

munificent:
Very nice!

> Iterating the returned iterator will be slightly slower due to the added indirection.

While I certainly care about perf, especially in heavily-used types, this doesn't worry me *too* much. What matters more is that iterating *lists* is fast. If iterating one of these sequences is too slow, you can always call `.list` and iterate over that.

> You should be aware that modifications made to the original sequence will affect the returned sequence.

Yup, but I think this is a feature as much as a bug. We just need to make sure people know.

builtin/core.wren
bjorn: Yep, that definitely looks cleaner!

builtin/core.wren
bjorn: The reason there isn't is because you had not put a blank line either in `MapKeySequence` and `MapValueSequence`, but I will add one. I guess the rule is that there should be a blank line unless the methods are tightly related AND are on one line.

bjorn:
Updated with changes from feedback.

munificent:
:metal:

PR #244, opened 2015-03-30
refactor defineSingleClass and defineClass

patriciomacadden:
`defineSingleClass` and `defineClass` (in `src/vm/wren_core.c`) are identical, except for their names and a function call. I extracted the duplication into a macro and then called the macro, of course.

I'm not sure if this is good in terms of clarity, but it doesn't duplicate code. Also, I don't know if there's a better way of refactoring this.

Lastly, I could move the `#define DEF_CLASS` macro up in the file if you want.

### DIFF
--- a/src/vm/wren_core.c
+++ b/src/vm/wren_core.c
@@ -1432,24 +1432,23 @@ DEF_PRIMITIVE(string_subscript)
   RETURN_ERROR("Subscript ranges for strings are not implemented yet.");
 }
 
-static ObjClass* defineSingleClass(WrenVM* vm, const char* name)
-{
-  ObjString* nameString = AS_STRING(wrenStringFormat(vm, "$", name));
-  wrenPushRoot(vm, (Obj*)nameString);
-
-  ObjClass* classObj = wrenNewSingleClass(vm, 0, nameString);
-  wrenDefineVariable(vm, NULL, name, nameString->length, OBJ_VAL(classObj));
-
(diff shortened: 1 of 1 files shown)


munificent:
Oh, yeah. There is some redundancy there. Nice catch!

I think we can do this a bit simpler instead of adding a macro. How about:

1. Turn `wrenNewSingleClass` into a static `newSingleClass` function in wren_value.c.
2. Make `wrenNewClass` call that and return the result if `vm->classClass` is `NULL`.
3. Get rid of `defineSingleClass` in wren_core.c.

So, basically, make `wrenNewClass` automatically delegate to the `newSingleClass` behavior if it sees that the core `Object` and `Class` classes haven't been wired up yet.

What do you think?

patriciomacadden:
That's much better! I don't know how I didn't see that before. I guess I was trying to make it look like `DEF_PRIMITIVE`.

I ended getting rid of `defineSingleClass` and refactoring `defineClass` to handle both cases as you suggested (if `Object` and `Class` classes haven't been defined yet, then create "raw" classes as you said and explained [here](https://github.com/munificent/wren/blob/master/src/vm/wren_value.h#L593-L595)).

What do you think now?

munificent:
Groovy! Thanks!

If you haven't already, feel free to send a pull request that adds yourself to the AUTHORS file. Thank you!

patriciomacadden:
Sure, I will. Thank you!

PR #245, opened 2015-03-30
Fixed char-subscripts warning

ComFreek:
Without this fix, `make` would always produce the following warning/error:
```
ComFreek@ComFreeks-PC /cygdrive/p/wren
$ make
make[1]: Entering directory '/cygdrive/p/wren'
        cc src/vm/wren_core.c             -Os -std=c99
src/vm/wren_core.c: In function ‘prim_num_fromString’:
src/vm/wren_core.c:945:3: error: array subscript has type ‘char’ [-Werror=char-subscripts]
   while (*end != '\0' && isspace(*end)) end++;
   ^
script/wren.mk:128: recipe for target 'build/release/vm/wren_core.o' failed
make[1]: *** [build/release/vm/wren_core.o] Error 1
make[1]: Leaving directory '/cygdrive/p/wren'
Makefile:9: recipe for target 'release' failed
make: *** [release] Error 2

ComFreek@ComFreeks-PC /cygdrive/p/wren
```

My system:
- Windows 8 Pro
- Cygwin with GCC 4.9.2
- Branch `master` of Wren, last commit is https://github.com/munificent/wren/commit/54ee4987d19c1b5ee16297ffd85a8e571bd7b954

**NOTE:** I am not really a C programmer, so the change could possibly introduce ugly and/or inefficient code :) Please counter-check!

### DIFF
--- a/src/vm/wren_core.c
+++ b/src/vm/wren_core.c
@@ -942,7 +942,7 @@ DEF_PRIMITIVE(num_fromString)
   double number = strtod(string->value, &end);
 
   // Skip past any trailing whitespace.
-  while (*end != '\0' && isspace(*end)) end++;
+  while (*((unsigned char*)end) != '\0' && isspace(*((unsigned char*)end))) end++;
 
   if (errno == ERANGE)
   {


src/vm/wren_core.c
```diff
@@ -942,7 +942,7 @@ DEF_PRIMITIVE(num_fromString)
   double number = strtod(string->value, &end);
 
   // Skip past any trailing whitespace.
-  while (*end != '\0' && isspace(*end)) end++;
+  while (*((unsigned char*)end) != '\0' && isspace(*((unsigned char*)end))) end++;
```
bjorn: The first check can't produce the warning you're fixing, only the call to `isspace` would if it is implemented as a macro that does an array index, which is presumably the case in your headers.

Also, the cast could be done after dereferencing the pointer which looks a little cleaner:

```
while (*end != '\0' && isspace((unsigned char) *end)) end++;
```

munificent: :+1: It seems `isspace` in cygwin is a bit pickier than other C libs. @bjorn's suggestion is how I'd do this.

munificent:
> I am not really a C programmer

You're one step closer to being one now!

ComFreek:
> It seems  isspace  in cygwin is a bit pickier than other C libs. @bjorn's suggestion is how I'd do this.

I hope I correctly interpreted the last sentence that I should open another PR --> #247.

BTW, a big :+1: for the fast compilation speed of the source.

PR #247, opened 2015-03-31
Fixed char-subscripts warning

ComFreek:
See #245

### DIFF
--- a/src/vm/wren_core.c
+++ b/src/vm/wren_core.c
@@ -942,7 +942,7 @@ DEF_PRIMITIVE(num_fromString)
   double number = strtod(string->value, &end);
 
   // Skip past any trailing whitespace.
-  while (*end != '\0' && isspace(*end)) end++;
+  while (*end != '\0' && isspace((unsigned char) *end)) end++;
 
   if (errno == ERANGE)
   {


bjorn:
You should not open a new PR to make a fix. Generally, you amend your patch or you commit additional patches based on feedback, and then either force-push (in case of amending) or normal-push (in case of extra commits) to the same branch you had requested to be merged (so to the `patch-1` branch in your fork).

At this point, it is too late to do that, so this is just some information for next time. :-)

ComFreek:
I didn't want to clutter the commit history and I learnt that [force-pushing could be dangerous](http://stackoverflow.com/questions/1270514/undoing-a-git-push#comment11064932_1270608) &mdash; at least in shared repos. According to your comment, I suppose the latter is usually not the case when submitting PRs.
Thanks for the info!

bjorn:
> According to your comment, I suppose the latter is usually not the case when submitting PRs.

Right, a PR is explicitly a work-in-progress, and tweaks to the commits contained within are expected and a supported workflow, which requires force-pushing. It's also perfectly valid to just push additional commits, but as you say this can clutter the history.

munificent:
Thanks!

PR #249, opened 2015-04-01
Fix warning in clang 3.5 (-Wabsolute-value)

lluchs:
Travis builds with clang 3.4, so the warning didn't appear there.


munificent:
Thanks!

PR #256, opened 2015-04-06
Some documentation updates

bjorn:
Some `Sequence.list` references remained after it was renamed to `Sequence.toList`.


munificent:
\o/

PR #262, opened 2015-04-25
Add syntax example file

lluchs:
The file contains most syntactic constructs in wren and is mainly interesting when creating syntax highlighting definitions.

[Example rendering](https://rawgit.com/lluchs/ff2c0e380af8c58e0154/raw/syntax.wren.html) with [vim-wren](https://github.com/lluchs/vim-wren)


munificent:
Great idea. Thanks!

PR #263, opened 2015-04-26
new typeof keyword

minirop:
I've added a new keyword `typeof` that returns the type of the expression as a string (`typeof(4+3) == Num`). Can be useful for error messages:

```cpp
class Plane {
  pilot=(value) {
    if(!(value is Pilot)) {
      IO.print("Error: 'Pilot' expected got '", typeof(value), "'.")
    }
  }
}

var p = new Plane
p.pilot = 3 // print "Error: 'Pilot' expected got 'Num'."
```

if you (Bob or any Wren-user) have any comment, remark, anything, feel free!

ps: I couldn't run the tests (yay, windows!), but I assume I didn't broke anything and wrote the new tests properly.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -12,3 +12,4 @@ Marco Lizza <marco.lizza@gmail.com>
 Raymond Sohn <raymondsohn@gmail.com>
 Thorbjørn Lindeijer <bjorn@lindeijer.nl>
 Patricio Mac Adden <patriciomacadden@gmail.com>
+Alexander Roper <minirop@gmail.com>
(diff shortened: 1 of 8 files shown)


bjorn:
Personally, I think I would expect `typeof` to return the type of its argument rather than the name of the type.

minirop:
I thought about the JS `typeof` while writing the thing (and I'm not fully fluent in Wren), but that's probably a better option.

bjorn:
Hmm, on second thought I'm afraid this just needlessly duplicates `Object.type`. Or do you see any advantage of having this as a keyword?

minirop:
none, I just didn't know that existed. (but nonetheless it was fun hacking wren internal).

munificent:
Yup, like @bjorn says, this is redundant with `.type`. I'm still glad you tried this out though, because I'm sure it's a good learning experience. :)

PR #264, opened 2015-05-03
Allow lists and maps with trailing comma

lluchs:
This is useful for multi-line lists and maps as it allows shuffling/adding/removing lines without having to care about commas. It's also supported in other languages with braces such as C(++) and JS (although only in their newer variants).

### DIFF
--- a/src/vm/wren_compiler.c
+++ b/src/vm/wren_compiler.c
@@ -1815,6 +1815,9 @@ static void map(Compiler* compiler, bool allowAssignment)
     {
       ignoreNewlines(compiler);
 
+      // Map with trailing comma.
+      if (peek(compiler) == TOKEN_RIGHT_BRACE) break;
+
       // Push a copy of the map since the subscript call will consume it.
       emit(compiler, CODE_DUP);
 
--- a/src/vm/wren_compiler.c
+++ b/src/vm/wren_compiler.c
@@ -1778,6 +1778,9 @@ static void list(Compiler* compiler, bool allowAssignment)
     {
       ignoreNewlines(compiler);
 
+      // List with trailing comma.
+      if (peek(compiler) == TOKEN_RIGHT_BRACKET) break;
+
       // Push a copy of the list since the add() call will consume it.
       emit(compiler, CODE_DUP);
 
--- a/test/language/list/trailing_comma.wren
+++ b/test/language/list/trailing_comma.wren
@@ -1,0 +1,7 @@
+var list = [
+  "a",
+  "b",
+]
+
+IO.print(list[0]) // expect: a
+IO.print(list[1]) // expect: b
--- a/test/language/list/trailing_comma.wren
+++ b/test/language/list/trailing_comma.wren
@@ -5,3 +5,8 @@ var list = [
 
 IO.print(list[0]) // expect: a
 IO.print(list[1]) // expect: b
+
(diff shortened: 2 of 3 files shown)


bjorn:
Right, I'm also used to this from Lua and consider it very useful.

munificent:
This has always felt dirty to me in languages that allow it. I don't know exactly why, but it offends my grammatical sensibilities or something. (But then, I'm also someone who cares about the difference between en dashes and em dashes, so I'm probably more punctuation-scrupulous than most.)

But, apparently, Lua, PHP, C#, Java, Ruby, JS, and Python support it so I guess it's the thing to do.

This could use a few more tests just to make sure we didn't loosen it up in weird ways. We want to make sure it still disallows:

```
[,]
[1,,]
[1,,2]
// And the same for maps.
```

Thanks for doing this!

MarcoLizza:
I second  @munificent.

Since the early days while programming in C it always mined my grammar consciousness to put a trailing comma in (for example) the last item of an enumeration. It would be still valid, but I simply don't get the point of doing this, since I want to be in full control of what I'm doing.

( and, yes, probably I'm over-scrupulous on the visual quality of my code, too... ;) )

preble:
I like having the flexibility to use a trailing comma. Its biggest selling point for me is that it keeps the source control diff/blame for your code less noisy – you don't have to change two lines if you just want to add or remove an item to/from an array/map. (I think this is basically what @lluchs is saying in the initial description but I wanted to emphasize the source control aspect.)

lluchs:
Is it really necessary to have a separate file for each of these tiny failing test cases? I tried something like this:

```dart
Meta.eval("[,]") // expect runtime error: Could not compile source code.   
Meta.eval("[1,,]") // expect runtime error: Could not compile source code. 
Meta.eval("[1,,2]") // expect runtime error: Could not compile source code.
```

...but the test script doesn't seem to register these correctly. I get `Expected runtime error on line 3 but was on line 1`. I think it stops executing after the first error (unlike the REPL)?

bjorn:
@lluchs When `Meta.eval` raises a runtime error when the compilation fails, you may need to wrap this in a `Fiber` and use `Fiber.try` on it (a method that is apparently not documented, but it catches such errors).

minirop:
@bjorn `Fiver.try` is *documented* in "Error Handling".

bjorn:
@minirop Puh, I had searched for it at https://munificent.github.io/wren/fibers.html and http://munificent.github.io/wren/core/fiber.html ... anyway of course it should simply still be added in the latter page at least.

lluchs:
Thanks, that worked. I also rebased the branch onto current master.

munificent:
Using `Meta.eval` is pretty clever, but if it's OK with you, I prefer to keep the tests simple, even if it means making a bunch of small files for compile error tests.

Using as few language and library features as possible in the tests makes it easier to track down bugs when tests are failing. It also makes it easier to make low-level changes to the VM or create new language implementations because you can start getting some tests passing even when few features work correctly.

munificent:
I think GitHub is a bit confused because I rebased these, but this is merged in now. See:

d7eed087745e967a05863f602d8495930a7babf3, b20c9a1b387b6fccef2fcd8fa93306cde1b96c8c, e1d35966438b7729f1f671a423695cceef0af0c8

Thanks!

Issue #266, opened 2015-05-16
Running set example errors


hij1nx:
Freshly pulled from github, after running make...

```
./wren example/set.wren
WhereSequence does not implement '+(_)'.
[main line 91] in &(_)
[main line 110] in (script)
true
```

bjorn:
Hrm, I didn't notice this example when I changed `Sequence.where` to be lazy rather that eagerly returning the computed list, which apparently the example Set class still relies on.

The example can be fixed by adding `.toList()` at the end of each call to `where` and the call to `map`.

bjorn:
Note that this example is now quite bad since implementing a set should be much more efficient based on the `Map` class instead of the `List` class. So while it could be fixed up to run, it probably really needs a complete rewrite.

munificent:
+1 for rewriting it to use a map.

zeckalpha:
Yeah, I wrote that before Maps existed. Feel free to remove the file entirely, too.

munificent:
Done!

PR #268, opened 2015-05-20
Run examples as tests. Would have prevented #266

zeckalpha:
Just a thought.

### DIFF
--- a/script/test.py
+++ b/script/test.py
@@ -44,10 +44,17 @@ skipped = defaultdict(int)
 num_skipped = 0
 
 
-def walk(dir, callback):
-  """ Walks [dir], and executes [callback] on each file. """
+def walk(dir, callback, ignored=None):
+  """
+  Walks [dir], and executes [callback] on each file unless it is [ignored].
+  """
+
+  if not ignored:
+    ignored = []
+  ignored += [".",".."]
+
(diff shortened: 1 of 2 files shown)


script/test.py
```diff
@@ -242,10 +244,15 @@ def run_test(path):
...
     print('')
 
+def run_example(path):
+  return run_test(path, example=True)
 
 for dir in ['core', 'io', 'language', 'limit', 'meta']:
   walk(join(WREN_DIR, 'test', dir), run_test)
 
+walk(join(WREN_DIR, 'example'), run_example)
+walk(join(WREN_DIR, 'example', 'import-module'), run_example)
```
munificent: This shouldn't be needed. `walk()` does a recursive walk.

munificent:
Good idea!

script/test.py
zeckalpha: Hmm. I didn't think it did given this line. https://github.com/munificent/wren/pull/268/files#diff-1d8e229374da562b542ae6b3bbeccc1bR250

I'll check and update as needed.

zeckalpha:
Updated. Cleaned it up a bit.

munificent:
I dig it. Thanks!

Issue #269, opened 2015-05-22
Improved VM performance for Windows/VC++


BarabasGitHub:
I read your article about Pratt Parsers and somehow stumbled upon Wren, which looks really nice! I briefly went through the implementation and saw the mention of computed gotos. I also read this article http://eli.thegreenplace.net/2012/07/12/computed-goto-for-efficient-dispatch-tables/ last week. And it got me thinking if it wasn't somehow possible to achieve the same on windows with VC++. 
After thinking about it on and off for a week I made some pseudo computed goto implementation for the Wren VM. Some quick tests showed 10-12% improvement on the benchmarks. It works by implementing the goto table as a switch case (in a macro), which contains the correct jumps back to the instructions.

However I had to include the 'wren_opcodes.h' in the switch case which is in a macro. And that's not possible. So my solution was to wrap the whole 'wren_opcodes.h' into a macro. It's not very pretty and I can understand if you're not happy about that. (I'm not =D) But I don't know of a better solution. I've never programmed in C before, only C++, so also macros are a bit new to me.

I'm not sure how the generated code compares to the actual computed goto implementation, since I haven't got Linux/GCC at hand here.

You can check out the code on my fork. Let me know if you're interested.

lluchs:
I tested on Linux/Clang and it seems to work pretty well (baseline is computed goto):
```
binary_trees - wren            .......... 0.38s 0.2054  92.28% relative to baseline
delta_blue - wren              .......... 0.17s 0.0851  91.47% relative to baseline
fib - wren                     .......... 0.46s 0.2803  90.74% relative to baseline
for - wren                     .......... 0.15s 0.0962 104.88% relative to baseline
method_call - wren             .......... 0.23s 0.0055  94.14% relative to baseline
map_numeric - wren             .......... 0.60s 0.0018 101.43% relative to baseline
map_string - wren              .......... 0.14s 0.0011 107.44% relative to baseline
string_equals - wren           .......... 0.46s 0.0029 106.16% relative to baseline
```

As comparison, here's how the old fallback performed:
```
binary_trees - wren            .......... 0.42s 0.2361  82.88% relative to baseline
delta_blue - wren              .......... 0.19s 0.0943  79.13% relative to baseline
fib - wren                     .......... 0.54s 0.2093  76.87% relative to baseline
for - wren                     .......... 0.25s 0.0987  65.64% relative to baseline
method_call - wren             .......... 0.31s 0.0816  69.85% relative to baseline
[...]

PR #272, opened 2015-05-23
Pseudo Computed Goto

BarabasGitHub:
Since it is not possible to use computed gotos in Visual Studio/VC++ I have simulated the dispatch table with a switch case containing goto statements.


munificent:
Interesting!

How does this affect the performance? Can you run the benchmarks with and without this change and show us what the numbers look like?

BarabasGitHub:
There are some numbers in the discussion on issue #269
If you want I can try to run the benchmarks with the old and new approach later this week.

BarabasGitHub:
Sorry I'm a bit late, but some things came up. In any case I ran the benchmark script, with the original to generate a base-line and then with the new implementation.

Original:
```
binary_trees - wren            .......... 0.49s 0.0092 no baseline
delta_blue - wren              .......... 0.23s 0.0036 no baseline
fib - wren                     .......... 0.62s 0.0027 no baseline
for - wren                     .......... 0.31s 0.0017 no baseline
method_call - wren             .......... 0.41s 0.0100 no baseline
map_numeric - wren             .......... 0.80s 0.0133 no baseline
map_string - wren              .......... 0.20s 0.0024 no baseline
string_equals - wren           .......... 0.68s 0.0045 no baseline
```
New:
```
binary_trees - wren            .......... 0.42s 0.0048 118.27% relative to baseline
delta_blue - wren              .......... 0.19s 0.0013 121.51% relative to baseline
fib - wren                     .......... 0.55s 0.0033 113.09% relative to baseline
for - wren                     .......... 0.26s 0.0030 122.18% relative to baseline
method_call - wren             .......... 0.29s 0.0018 138.31% relative to baseline
map_numeric - wren             .......... 0.69s 0.0042 115.32% relative to baseline
map_string - wren              .......... 0.18s 0.0016 107.73% relative to baseline
string_equals - wren           .......... 0.54s 0.0027 125.09% relative to baseline
```

munificent:
OK, I'm sold. :)

I wish we didn't have to turn wren_opcodes.h into a giant macro, but I can't think of any other solution.

I'll add some other comments at specific points, but would you mind also rebasing this patch against the latest master? I think some stuff has moved under you, making this hard to merge in.

Thank you for doing so much work on this!

src/vm/wren_common.h
```diff
@@ -37,8 +37,8 @@
 // Defaults to on on supported compilers.
 #ifndef WREN_COMPUTED_GOTO
   #ifdef _MSC_VER
-    // No computed gotos in Visual Studio.
-    #define WREN_COMPUTED_GOTO 0
+    // No real computed gotos in Visual Studio. Use the pseudo version.
+    #define WREN_COMPUTED_GOTO 2
```
munificent: Instead of numeric values, let's do separate defines, I'm thinking:

```c
WREN_DISPATCH_SWITCH
WREN_DISPATCH_COMPUTED_GOTO
WREN_DISPATCH_JUMP_TABLES
```

`SWITCH` is the 0 case here, `COMPUTED_GOTO` is 1, and `JUMP_TABLES` is 2.

src/vm/wren_opcodes.h
```diff
@@ -7,186 +7,190 @@
...
-
-// This pseudo-instruction indicates the end of the bytecode. It should
-// always be preceded by a `CODE_RETURN`, so is never actually executed.
-OPCODE(END)
\ No newline at end of file
+// The whole file is one big macro, because it is later used inside
+// a '#define' where you cannot use an '#include'.
+#define PLACE_OPCODES                                                           \
+                                                                                \
+/* Load the constant at index [arg]. */                                         \
```
munificent: Instead of translating this all to `/* */` comments, it's probably simpler to just pull all of the comments out of the macro definition. More like:

```c
// CONSTANT
// Load the constant at index [arg].
//
//
// NULL
// Push null onto the stack.
//
// ...
#define PLACE_OPCODES \
    OPCODE(CONSTANT) \
    OPCODE(NULL) \
    ...
```

src/vm/wren_common.h
BarabasGitHub: I think that will indeed be more clear.

src/vm/wren_opcodes.h
BarabasGitHub: That might indeed be less horrible.

BarabasGitHub:
I also really dislike the wren_opcodes macro. I wish it could be different.

Anyway, I tried to rebase the goto branch onto the new master branch. But somewhere along the line I did something that made my master branch end up being the same as the goto branch. I think I might have pushed it to the wrong branch or so... AND I fixed it! 

I didn't change your suggestions. Should I do that?

paulsapps:
Did you try what a function pointer table to declspec(naked) functions looks like with MSVC? Since you can get the address of the function thats like taking the address of a label, then the "naked" function call should become a jmp?

BarabasGitHub:
I don't think that will work. It says on msdn: "The *naked* attribute
affects only the nature of the compiler's code generation for the
function's prolog and epilog sequences. It does not affect the code that is
generated for calling such functions. Thus, the *naked* attribute is not
considered part of the function's type, and function pointers cannot have
the *naked* attribute." (
https://msdn.microsoft.com/en-us/library/21d5kd3a.aspx)

paulsapps:
I saw something else related to this today "__assume" https://msdn.microsoft.com/en-us/library/1b3fsfxw.aspx apparently you could use this on the switch value to stop it generating the extract checks. Might be worth checking if you ever get time.

munificent:
Ugh, I'm so sorry, but I left this open so long that I'm not able to merge it in now since so many changes have happened since. :(

PR #276, opened 2015-06-01
Remove LOOP opcode and make jump offsets signed

NomAnor:
Removing the LOOP opcode and using a signed offset in the jump opcode makes the list of opcodes cleaner. I made the offset in all jump opcodes signed to have them uniform.

If the range is too short I would extend the argument.

### DIFF
--- a/src/vm/wren_compiler.c
+++ b/src/vm/wren_compiler.c
@@ -975,11 +975,12 @@ static int emit(Compiler* compiler, int byte)
   return compiler->bytecode.count - 1;
 }
 
+
 // Emits one 16-bit argument, which will be written big endian.
 static void emitShort(Compiler* compiler, int arg)
 {
-  emit(compiler, (arg >> 8) & 0xff);
-  emit(compiler, arg & 0xff);
+  emit(compiler, ((uint16_t)arg >> 8) & 0xff);
+  emit(compiler, (uint16_t)arg & 0xff);
 }
 
 // Emits one bytecode instruction followed by a 8-bit argument. Returns the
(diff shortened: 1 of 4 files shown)


minirop:
no need to extend the argument, if longer jump become necessary, another opcode could be used. The JVM has `jsr_w` and `goto_w` opcodes and [the `wide` opcode for some others](https://docs.oracle.com/javase/specs/jvms/se8/html/jvms-6.html#jvms-6.5.wide).

munificent:
I like the idea behind this, but I ran into two problems:

1. It regresses performance a bit. I'm not sure why, but the benchmarks get about 5% slower on my machine with this.

2. It relies on undefined behavior. As far as I know, the C standard doesn't guarantee the results of casting a negative value to an unsigned type. This is a big part of the reason Wren has a separate opcode for LOOP: it ensures all of the numbers stored in the bytecode are unsigned.

Any ideas on if these can be addressed? Have you tried running the benchmarks on your machine?

bjorn:
> As far as I know, the C standard doesn't guarantee the results of casting a negative value to an unsigned type.

Actually in this direction it seems to be well-defined. It's the unsigned to signed case that's left up the implementation:

> 6.3.1.3 Signed and unsigned integers
>
> 1. When a value with integer type is converted to another integer type other than _Bool, if the value can be represented by the new type, it is unchanged.
> 2. Otherwise, if the new type is unsigned, the value is converted by repeatedly adding or subtracting one more than the maximum value that can be represented in the new type until the value is in the range of the new type.
> 3. Otherwise, the new type is signed and the value cannot be represented in it; either the result is implementation-defined or an implementation-defined signal is raised.

Found this on [StackOverflow](http://stackoverflow.com/questions/50605/signed-to-unsigned-conversion-in-c-is-it-always-safe).

munificent:
Ah, right. In either case, this patch goes in both directions. In the compiler, it goes signed -> unsigned, and in the VM it goes back from unsigned -> signed.

munificent:
Sorry, I've been a bad open source maintainer. :(

I'm going to close this out since I don't think we'll be able to move forward with it.

PR #277, opened 2015-06-01
Refactor the common code in the interpreter loop

NomAnor:
I saw your todo comment in the interpreter loop so I put all the duplicated code in define statements. All the continuation characters are a nuisance when editing but now there's only one place to edit.

I added a new compile switch WREN_PERFORMANCE (the name could be better). When activated, every opcode gets a seperate code block. This increases the code size but it gives better performance (probably from better branch prediction).

### DIFF
--- a/src/vm/wren_common.h
+++ b/src/vm/wren_common.h
@@ -29,6 +29,13 @@
 #define WREN_NAN_TAGGING 1
 #endif
 
+// Generates fater code but increases the code size.
+//
+// Defaults to on.
+#ifndef WREN_PERFORMANCE
+#define WREN_PERFORMANCE 1
+#endif
+
 // If true, the VM's interpreter loop uses computed gotos. See this for more:
 // http://gcc.gnu.org/onlinedocs/gcc-3.1.1/gcc/Labels-as-Values.html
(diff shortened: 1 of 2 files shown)


minirop:
you say you refactored the code, buy why duplicating the `#if WREN_PERFORMANCE LOAD_LOCAL_CODE(X); #endif` all over the place ? you only *win* 2 CPU instructions (2 load and 1 sub VS 1 load, right?).

NomAnor:
Your're right that's unnecessary. For the call code I'm not so sure. With gcc 5.1 and -O3 the seperated code for the call opcodes gives me around +10% in the fib and method_call benchmark compared to gcc 5.1 and -O3 without the seperation. Maybe others can run the benchmarks?

munificent:
Doing a giant `#define` is a neat idea to workaround the `STORE_FRAME` and `LOAD_FRAME` limitations, but I think it would be too hard to maintain. I spent some time poking around at this this morning and managed to come up with a simpler way to share code between the `CODE_CALL` and `CODE_SUPER` cases: just use a `goto`: https://github.com/munificent/wren/commit/1a87f52e9d2d3cbb4fa15b0c465de177b6dbfb8b

PR #279, opened 2015-06-04
Fix for memory leaks

NomAnor:
Valgrind's memcheck showed me some memory leaks when running the wren cli so I tried to fix them.

The first is in the cli itself. It's not a serious one because the memory gets freed a soon a the programm exits. The memory allocated for the root directory path is not freed.

In the wren compiler the constants buffer data's ownership is not transfered to the new function but the constans are copied so the buffer's data must be freed.

The last one is debatable. According to the C Standard `realloc(NULL, size)` is the same as `malloc(size)`. When the memory for the internal data structures is allocated the previous memory (which defaults to NULL) is deallocated by calling `DEALLOCATE(vm, NULL)` this will invoke `realloc(NULL, 0)` which will behave like `malloc(0)` which can return (implementation defined) a pointer to a zero sized object which must not be dereferenced, but freed. By using the DEALLOCATE macro with a NULL ptr these objects leak (they are size 0 so not really a memory leak). I changed the default reallocate function to one which does the right thing.

### DIFF
--- a/src/cli/io.c
+++ b/src/cli/io.c
@@ -46,6 +46,10 @@ char* readFile(const char* path)
 
 void setRootDirectory(const char* path)
 {
+  if (rootDirectory != NULL)
+  {
+    free((void*)rootDirectory);
+  }
   rootDirectory = path;
 }
 
--- a/src/cli/vm.c
+++ b/src/cli/vm.c
@@ -48,6 +48,7 @@ void runFile(WrenBindForeignMethodFn bindForeign, const char* path)
 
   wrenFreeVM(vm);
   free(source);
+  setRootDirectory(NULL);
 
   // Exit with an error code if the script failed.
   if (result == WREN_RESULT_COMPILE_ERROR) exit(65); // EX_DATAERR.
--- a/src/vm/wren_compiler.c
+++ b/src/vm/wren_compiler.c
@@ -1280,6 +1280,7 @@ static void freeCompiler(Compiler* compiler)
 {
   wrenByteBufferClear(compiler->parser->vm, &compiler->bytecode);
   wrenIntBufferClear(compiler->parser->vm, &compiler->debugSourceLines);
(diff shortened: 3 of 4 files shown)


BarabasGitHub:
I think it's safe to do free(NULL). So you probably don't have to check for that.

NomAnor:
It is safe, but the checks are necessary because `realloc(NULL, 0)` is not equivalent to `free(NULL)`.

There are 4 possible invocations for the wren reallocation function:

  * `ptr == NULL && size == 0`: Do nothing because the function is asked to deallocate a NULL pointer, thats not the behaviour of realloc (it's equivalent to `malloc(0)`, which is implementation defined)
  * `ptr != NULL && size == 0`: Deallocate memory. It's implementation defined wether realloc only frees the memory or reallocates a new zero sized object.
  * `ptr == NULL && size > 0`: Allocate memory, handled by realloc
  * `ptr != NULL && size > 0`: Reallocate memory, handled by realloc

See http://en.cppreference.com/w/c/memory/realloc, same for malloc

> If new_size is zero, the behavior is implementation defined (null pointer may be returned (in which case the old memory block is not freed), or some non-null pointer may be returned that may not be used to access storage).

BarabasGitHub:
Ah, I see. So whenever the size == 0 free has to be called instead of
realloc. I understand. =)

src/cli/io.c
```diff
@@ -46,6 +46,10 @@ char* readFile(const char* path)
 
 void setRootDirectory(const char* path)
 {
+  if (rootDirectory != NULL)
+  {
+    free((void*)rootDirectory);
+  }
```
bjorn: Regardless, here an unnecessary NULL check is done before calling free.

munificent:
This is awesome! I really appreciate you doing this. I've tried to get Valgrind running on my machine before but was never successful.

I made a few changes to your patch but otherwise merged it in. Thanks!

If you want, send me a PR to add yourself to the AUTHORS file. :grinning:

PR #283, opened 2015-07-10
Make constructors just methods.

munificent:
* Eliminate "new" reserved word.
* Allow "this" before a method definition to define a constructor.
* Only create a default constructor for classes that don't define one.


bjorn:
Overall I like the change, but I have a few concerns:

* What made you pick `this`? It does not intuitively remind me of anything related to construction, so I wonder what your reasoning was.

* Why does the default `new` method come with parenthesis? Isn't that just forcing unnecessary syntax clutter everywhere?

I have a small proposal (a modification of the example you wrote to the mailing list), which I fully expect to be shot down cause of some oversight, but here it is:

```dart
class Vector {
  new(x, y) {
    _x = x
    _y = y
  }

  new polar(theta, radius) {
    _x = theta.cos * radius
    _y = theta.sin * radius
  }
}

var a = Vector(1, 2)
var b = Vector.polar(Num.pi / 4, 1)
```
So here I'm keeping the `new` keyword, but giving it the same meaning of the `this` keyword. In addition, I'm allowing the method name to be empty, in which case it will be called when the class itself is called as a function. What I'm thinking is, that when you force the parenthesis for the nameless constructor (so `new {}` is invalid), there will not be ambiguities in the code.

To me, the `new` keyword reads more intuitive and the resulting code would be less verbose.

bjorn:
Hmm, allowing for empty method names could also solve the `.call` verbosity for calling `Fn` instances. I really hope I'm not missing something here.

minirop:
if I understand correctly, the `this` keyword is about the same as saying "`this function is static, call the allocator before calling it`"?

Like @bjorn, I'm not really fond of `this` (probably my C++/JS/... background does not help), so maybe find another keyword (*constructor*? *init*?) or something a little more cryptic like a sigil (`$new`, `@new`).

lluchs:
Using `this` may also be confusing for people coming from Ruby where methods defined with `self` are static.

I like @bjorn's idea of empty method names for function calls. Elixir also does this.

munificent:
> What made you pick `this`? It does not intuitively remind me of anything related to construction, so I wonder what your reasoning was.

I needed some keyword, and `this` was already reserved. The way I was thinking of it is that a constructor's job is to create a new object, so it makes a "new this". Another way to think of it is that it lets you define a static method, but one that *does* have `this` inside of it.

I was hoping to avoid a new reserved word. And, in particular, for familiarity's sake I try to make the reserved words in Wren familiar from other popular OOP languages. Alas, none of those have a keyword for constructors we can use.

You may be right that it's confusing, though. We could use `new` as a keyword to indicate this, but that makes it harder to also use `new` as the method name. It's possible to do "contextual keywords" where an identifier is treated like a reserved word in some places, but I'd like to avoid that since it's a bit subtle for users.

I would be up for another keyword, I just didn't have any ideas.

> Why does the default new method come with parenthesis? Isn't that just forcing unnecessary syntax clutter everywhere?

There's an open issue with Wren currently about whether or not the default constructor that comes with a class should have parentheses or not. Likewise, there's sort of a style question about whether other constructors should have parentheses.

Especially if constructors are methods on the class, they look like getters without the `()`. Maybe I'm wrong, but I feel that could be confusing with things like `Time.now` or `Guid.new`. Though, now that I look at those, they seem fine.

It would be straightforward to extend this to allow constructors without parentheses. Should I do that?

> I'm allowing the method name to be empty, in which case it will be called when the class itself is called as a function.

Alas, that doesn't work. I understand not allowing "bare" function invocations is an annoying part of the syntax, but it follows naturally from other syntax choices. Wren is similar to Ruby and Smalltalk, and they also require a named method (`.call` and `value`) to invoke functions/blocks.

In particular, with your suggestion, if you store a reference to a class in a field or local variable, and then try to invoke it through that, you're back to the same ambiguities around invoking functions.

> if I understand correctly, the this keyword is about the same as saying "this function is static, call the allocator before calling it"?

Yeah, sort of. You could also think of it as, "this method is static (it's called on the class), but it's body is an instance method, so allocate an object first".

> constructor? init?

I would do `constructor` but it's such a long word. :( `init` would work but I feel that's a common name for methods so I don't want to take that away from people by reserving it.

> Using this may also be confusing for people coming from Ruby where methods defined with self are static.

Good point. We could do `ctor`, but I worry that's too cryptic. Any other ideas?

bjorn:
> In particular, with your suggestion, if you store a reference to a class in a field or local variable, and then try to invoke it through that, you're back to the same ambiguities around invoking functions.

Arg! That's what I meant with thing I said about some overight earlier. :-)

PR #284, opened 2015-07-10
added scientific notation

minirop:
Hi there, here is my attempt to allow scientific notation of numbers.

### DIFF
--- a/src/vm/wren_compiler.c
+++ b/src/vm/wren_compiler.c
@@ -608,7 +608,6 @@ static void readHexNumber(Parser* parser)
 // Finishes lexing a number literal.
 static void readNumber(Parser* parser)
 {
-  // TODO: scientific, etc.
   while (isDigit(peekChar(parser))) nextChar(parser);
 
   // See if it has a floating point. Make sure there is a digit after the "."
@@ -616,6 +615,22 @@ static void readNumber(Parser* parser)
   if (peekChar(parser) == '.' && isDigit(peekNextChar(parser)))
   {
     nextChar(parser);
+    while (isDigit(peekChar(parser))) nextChar(parser);
+  }
+  
+  // See if the number is in scientific notation
+  if (peekChar(parser) == 'e' || peekChar(parser) == 'E')
+  {
+    nextChar(parser);
+    
+    // if the exponant is negative
+    if(peekChar(parser) == '-') nextChar(parser);
+    
+    if(!isDigit(peekChar(parser)))
+    {
+      lexError(parser, "Unterminated scientific notation.");
+    }
+    
     while (isDigit(peekChar(parser))) nextChar(parser);
   }
 
--- a/test/language/number/scientific_literals.wren
+++ b/test/language/number/scientific_literals.wren
@@ -1,0 +1,9 @@
+var x = 2.55e2
+
+IO.print(x) // expect: 255
+IO.print(x + 1) // expect: 256
+IO.print(x == 255) // expect: true
+IO.print(2.55e-2 is Num) // expect: true
+IO.print(x is Num) // expect: true
+IO.print(-2.55e2) // expect: -255
+IO.print(-25500e-2) // expect: -255
--- a/test/language/number/scientific_missing_fractional_part.wren
+++ b/test/language/number/scientific_missing_fractional_part.wren
@@ -1,0 +1,1 @@
+var x = 1.e // expect runtime error: Num does not implement 'e'.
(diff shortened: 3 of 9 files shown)


src/vm/wren_compiler.c
```diff
@@ -586,6 +586,21 @@ static void readNumber(Parser* parser)
...
   }
+  
+  // See if the number is in scientific notation
+  if (peekChar(parser) == 'e' || peekChar(parser) == 'E')
+  {
+    nextChar(parser);
+    
+    // if the exponant is negative
+    if(peekChar(parser) == '-')
+      nextChar(parser);
```
munificent: I hate to make style nits, but can you change this to:

```
if (peekChar(parser) == '-') nextChar(parser);
```

src/vm/wren_compiler.c
```diff
@@ -586,6 +586,21 @@ static void readNumber(Parser* parser)
...
+    
+    if(!isDigit(peekChar(parser)))
+      lexError(parser, "Unterminated scientific notation.");
```
munificent: And this to:

```c
if (!isDigit(peekChar(parser)))
{
  lexError(parser, "Unterminated scientific notation.");
}
```

My style rules here are:

1. Always have a space after `if` (and other control flow keywords).
2. Only omit the `{}` on an `if` if the entire thing fits on one line.

Thanks!

test/language/number/scientific_literals.wren
```diff
@@ -0,0 +1,9 @@
+var x = 2.55e2
+
+IO.print(x) // expect: 255
+IO.print(x + 1) // expect: 256
+IO.print(x == 255) // expect: true
+IO.print(2.55e-2 is Num) // expect: true
+IO.print(x is Num) // expect: true
+IO.print(-2.55e2) // expect: -255
+IO.print(-25500e-2) // expect: -255
```
munificent: These tests are great!

Can you add a couple of error case tests too? You'll probably need to make each one a separate tiny file. I'm thinking things like:

```
1e
1.e
1.2e
1.2e3e
1.2e3.4
```

Also, two tests like "literal_too_large.wren" to test how it handles an overly long exponent and an overly long mantissa.

I know that's a lot of corner cases, but I think it's really really important to have very detailed test coverage. It's one of the things we can do to give people confidence in embedding Wren in their programs.

munificent:
This is awesome! Thanks for doing this! I have a couple of comments, but otherwise this looks swell. :)

test/language/number/scientific_literals.wren
minirop: For the very long exponent and mantissa, I don't really know how to write the tests, since it is the resulting number that counts. If `9999....9999` results in "too large literal", adding `e-100` to it will no longer trigger the error.

minirop:
I've updated the PR with your comments. I've just left out the overly mantissa/exponent (see my other comment).
You (or another linux user) have to "fix" the tests since on windows it's not as flawless (and I'm not really sure what "expect" message I have to put).

test/language/number/scientific_missing_fractional_part.wren
```diff
@@ -0,0 +1 @@
+var x = 1.e // expect error
```
munificent: It looks like this test isn't failing at compile time. It's parsing this as `(1).e`, which makes sense. That allows things like `123.abs` to work so I think the behavior is right here.

You should just be able to change this to expect a runtime error to occur.

minirop: I've just updated the file (I didn't understood that "expect error" was only for compile time errors), but the test still fails. And I have no clue why. The test does pass on my machine.

minirop:
Finally! I found what was failing. It was a '\r' in the test file... :sweat: 
To prevent this, you could remove them on line 79 of *test.py*: `self.runtime_error_message = match.group(1).replace('\r', '')`

test/language/number/scientific_literals.wren
munificent: Yeah, that's fine. I don't think we need to worry about this then.

munificent:
This looks great. I'd be happy to merge this in. Before I do, can you sync up (either merge or rebase) against the latest master? I tried to do it locally but wasn't able to get Git into a happy place. The easiest way I could figure out to get your changes on master was to cherry-pick, but I don't want to lose your commit authorship.

minirop:
here we go!

munificent:
Awesome, thank you! You rock!

PR #286, opened 2015-07-10
Added string_equals.py to benchmarks

BarabasGitHub:
It was missing so I thought I'd add it, only took me five minutes. =)


munificent:
Rad, thanks!

PR #292, opened 2015-08-13
Update VS project files to build with libuv and the timer module

Rohansi:
`shutil.rmtree` was failing to remove the .git directory because it's full of readonly files, so `rmtree_onerror` was added to try and get around that.

`scripts/libuv.py` isn't run on build right now because it currently does not check if libuv exists and has been built.

### DIFF
--- a/script/libuv.py
+++ b/script/libuv.py
@@ -22,25 +22,15 @@ def ensure_dir(dir):
 
   os.makedirs(dir)
 
-def rmtree_onerror(func, path, exc_info):
-  """
-  Error handler for ``shutil.rmtree``.
-
-  If the error is due to an access error (read only file)
-  it attempts to add write permission and then retries.
-
-  If the error is for another reason it re-raises the error.
-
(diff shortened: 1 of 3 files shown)


script/libuv.py
```diff
@@ -30,7 +49,7 @@ def download_libuv():
   # version number in this script changes.
   if os.path.isdir(LIB_UV_DIR):
     print("Cleaning output directory...")
-    shutil.rmtree(LIB_UV_DIR)
+    shutil.rmtree(LIB_UV_DIR, onerror=rmtree_onerror)
```
munificent: Oh, Windows. Always gotta make things difficult!

Instead of this, how do you feel about just detecting the OS and on Windows calling:

```python
subprocess.check_call(['cmd', '/c', 'rd', '/s', '/q', path])
```

instead of `shutil.rmtree()`?

That way, Linux and Mac are left alone.

munificent:
Thanks for working on this!

munificent:
Woo, thanks! One step closer to being able to merge libuv into master. :)

PR #293, opened 2015-08-13
Fix some bugs in the libuv timer module

Rohansi:
* `timerGetSource` was returning a string without setting its null terminator
* Timer handles were being freed when libuv still had access to them

With this and #292, Wren builds and runs all the timer tests successfully on Windows.

### DIFF
--- a/src/module/timer.c
+++ b/src/module/timer.c
@@ -70,8 +70,7 @@ char* timerGetSource()
 {
   size_t length = strlen(timerLibSource);
   char* copy = (char*)malloc(length + 1);
-  strncpy(copy, timerLibSource, length);
-  copy[length] = '\0';
+  strncpy(copy, timerLibSource, length + 1);
   
   return copy;
 }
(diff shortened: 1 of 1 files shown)


src/module/timer.c
```diff
@@ -28,11 +28,19 @@ static const char* timerLibSource =
...
+}
+
 // Called by libuv when the timer has completed.
 static void timerCallback(uv_timer_t* handle)
 {
   WrenValue* fiber = (WrenValue*)handle->data;
-  free(handle);
+  
+  // Tell libuv that we don't need the timer anymore.
+  uv_close((uv_handle_t*)handle, timerCloseCallback);
```
munificent: Oh, great catch. Can you tell I'm a noob with libuv? :)

src/module/timer.c
```diff
@@ -63,6 +71,7 @@ char* timerGetSource()
   size_t length = strlen(timerLibSource);
   char* copy = malloc(length + 1);
   strncpy(copy, timerLibSource, length);
```
munificent: Let's just do `length + 1` and then we don't need to explicitly terminate.

src/module/timer.c
Rohansi: I'm surprised you missed it! Without this I get an access violation on exit.

munificent:
:metal:

PR #298, opened 2015-08-19
Override default constructors for primitive types. [Fixes #296]

### LINKED ISSUE #296: Built-in types shouldn't expose default constructors
Since String, Range, etc. don't define constructors, the compiler gives them a .new(). They do bad things when you call them.

For Range, we should probably just give it a constructor. For the other types, we should make them throw runtime errors, or just not define them.

gsmaverick:


src/vm/wren_core.c
```diff
@@ -262,6 +267,11 @@ DEF_PRIMITIVE(class_toString)
 
 DEF_PRIMITIVE(fiber_new)
 {
+  RETURN_ERROR("Fiber metaclass does not implement 'new()'.");
+}
```
munificent: We don't need this one.

src/vm/wren_core.c
```diff
@@ -472,6 +482,11 @@ DEF_PRIMITIVE(fiber_yield1)
 
 DEF_PRIMITIVE(fn_new)
 {
+  RETURN_ERROR("Fn metaclass does not implement 'new()'.");
+}
```
munificent: Or this one.

munificent:
Two minor requests, but this looks great, thanks!

src/vm/wren_core.c
gsmaverick: Hmm, maybe I'm missing something but the following behaviour doesn't seem correct:

```
\\/"-
 \_/   wren v0.0.0
> var g = Fn.new()
> g.call()
Segmentation fault: 11
```

It seems the default constructor is still being created for `Fn`. Now I'm not sure if the solution I have in this PR is the correct one or if maybe there's a problem with the default constructor logic in the compiler?

gsmaverick: Actually now that I look at the code a bit I think I see the disconnect. It's the compiler that decides whether to create a default constructor but when the source for the core code is parsed the [Fn class](https://github.com/munificent/wren/blob/eda3f89266276878ea4c17ca46d635b5c3d43938/src/vm/wren_core.c#L17) is empty and does not yet define its constructor. From the compiler's perspective it sees a class without any constructors and creates a default constructor for it anyways.

munificent: > From the compiler's perspective it sees a class without any constructors and creates a default constructor for it anyways.

Ah, right. Because primitives are injected in the class after the fact unlike foreign methods.

One option would be to put the fake constructors that abort in the .wren code (using `Fiber.abort()`) instead of as a primitive.

It does feel like we may need a more general-purpose solution, though. For example, `Sequence` is basically an abstract base class. It's designed to be extended but not constructed. There isn't really any good way to express that right now.

This touches on some areas I've been thinking a lot about lately (including stuff like mixins), but I don't know if I have any good solutions yet.

minirop: one possible solution could be to create those class completely in C (by doing what `classDefinition()` does).

munificent: That would work for some of the classes, but not the ones that do need to define some methods in Wren, like String.

munificent:
I took the nuclear option to fixing this instead—no more default constructors at all! :)

Getting rid of them makes a few tiny class definitions a bit more verbose, but it lets you easily create abstract base classes that can be inherited but not constructed. It also simplifies the VM a bit and fixes this bug.

PR #299, opened 2015-08-19
updated comments where "this" was still used for constructors

minirop:
There was those 2 places were "this" was still used instead of "construct".


munificent:
Nicely done. I really appreciate clean-up stuff like this. Our future selves will thank us.

PR #300, opened 2015-09-06
Correct platform.system() call

jesskay:
An "arch" somehow ended up in the platform.system() call, which takes no arguments.


munificent:
Whoops, thanks!

PR #301, opened 2015-09-13
Fixes libuv build for systems where python is python3

krypt-n:
Fixed by using python2 when executing the libuv scripts.

### DIFF
--- a/script/libuv.py
+++ b/script/libuv.py
@@ -14,6 +14,15 @@ import sys
 LIB_UV_VERSION = "v1.6.1"
 LIB_UV_DIR = "deps/libuv"
 
+def python2_binary():
+    """Tries to find a python 2 executable"""
+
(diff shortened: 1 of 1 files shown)


munificent:
This breaks users that don't have a `python2` executable (which includes me and I suspect many others). :(

Maybe it could check for the presence of `python2` first and use it and fall back to `python` if it's not defined?

bjorn:
Ideally the scripts would support both Python 2 and 3, but I guess that'd be kinda hard.

krypt-n:
Oh, I assumed that python2 is standard in python packages and the distributions I looked at confirmed that. I'll try finding a workaround

munificent:
Thanks!

krypt-n:
Updated, now uses the default python executable if the major version is 2.

munificent:
Rad, thanks!

PR #302, opened 2015-09-19
Allow empty method names

lluchs:
This is a proposal to get a nicer function call syntax into wren. It is inspired by Elixir and turned out to be pretty easy to implement.

With this pull request, it's possible to define methods like this:

``` dart
class EmptyMethodName {
  construct new() {}
  () {
    System.print("Hello World")
  }

  (a) {
    System.print(a)
  }
}
```

...and then call them like this:

``` dart
var fn = EmptyMethodName.new()
fn.()
fn.("foo")
```

The additional dot avoids ambiguities with getters:

``` dart
class test {
  construct new() {}
  fn { Fn.new { System.print("ok") } }
}
test.new().fn.()
```

### DIFF
--- a/src/vm/wren_compiler.c
+++ b/src/vm/wren_compiler.c
@@ -1629,8 +1629,17 @@ static Signature signatureFromToken(Compiler* compiler, SignatureType type)
   
   // Get the token for the method name.
   Token* token = &compiler->parser->previous;
-  signature.name = token->start;
-  signature.length = token->length;
+  // fn.() or () {}
+  if (token->type == TOKEN_DOT || token->type == TOKEN_LINE)
+  {
+    signature.name = "";
+    signature.length = 0;
+  }
+  else
+  {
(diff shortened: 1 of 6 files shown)


IsaacWoods:
I agree something is needed to disambiguate between getters and nameless methods, but I really don't like the look of that dot... Otherwise, nice idea!

munificent:
This is pretty interesting. I didn't realize Elixir allowed that syntax. I think I briefly considered the same thing for Wren but I was afraid it would look too unusual to users.

How do you feel about bringing this up on the mailing list? I'd like to see what others think. My feeling is that `.call()` is a little more approachable even though it's longer, but I'm not that strongly attached to it.

I do feel we should do one or the other and not both.

underscorediscovery:
This pull seems to be out of the loop with the latest Wren, since .call exists and is in use, so I'm gonna close this.

PR #303, opened 2015-09-19
Templates

munificent:
I've been a lot about how annoying it is to print a little formatted string in Wren right now. I originally had the overloads in `IO.print(...)` so you could do:

```dart
IO.print("Hi, ", name, "! You are ", year - birth, " years old.")
```

That sort of works, but it's still kind of ugly and feels hokey. Also, it only works for `print()`. There's no easy way to just create a formatted string.

Something like `printf()`-style format strings aren't a good fit for Wren because we don't have varargs. Also, I never liked the separation between the formatted value and its position in the string.

All of this leads towards string interpolation—a way to insert expressions in the middle of a string literal that get evaluated and the stringified result is placed into the string.

Stuff like:

```
"Hi, #{name}! You are #{year - birth} years old."   # Ruby
"Hi, ${name}! You are ${year - birth} years old."   # Dart
"Hi, %(name)! You are %(year - birth) years old."   # Python
"Hi, {{name}}! You are {{year - birth}} years old." # Mustache et. al.
```

It's a nice feature, but it feels kind of special-casey to me. You'll note that even languages that have interpolation rarely use it for things like web templating. *However*, there is a way to really kick it up a notch: let user code control the interpolation.

This is what [tagged template strings](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/template_strings) in ES6 do. I've wanted to do something similar for several years. Yesterday, I realized that adding this to Wren would actually be pretty easy. So that's what this branch is. I'd like your feedback on it.

## Syntax

Any string literal can contain template expressions, delimited by `%(` and `)`:

```
"Hi, %(name)! You are %(year - birth) years old."
```

Any expression is allowed inside the parentheses, including other template strings:

```
"%( "%( "inner" )" ) "
```

### Why `%`?

I considered `#`, `$`, and `%`. I think the latter is the most intuitive once we introduce format strings (below).

### Why `()`?

I considered `()`, `{}`, and `{{}}`. The latter is more verbose. `{}` is more common, but is also a bit more visually noisy. Given that `{}` is used for things like maps, I think it could look weird in some cases:

```
"interpolate an empty map: %{{}}"
```

I like that `()` is visually clean and it does roughly convey the right impression regarding the grammar. The contents of it are any expression, which is the same as what's allowed inside `()` when it's used for grouping elsewhere.

This naturally extends to possibly supporting unparenthesized templates for simple primary expressions like names:

```
"point = %x, %y, %z"
```

I'm not *sure* if I want to support that (it might not play nice with format strings), but it might be cool.

Also, I just think `%()` looks nice.

## Semantics

From the user's perspective, when they write a template string, they just get a string object back with all of the template expressions evaluated and inserted in the result:

```dart
System.print("The sum is %(3 + 4)") // "The sum is 7".
```

What's going on under the hood is kind of interesting, and leads nicely to tagged templates below.

String templates are just syntax sugar. Given this string template:

```
"Hi, %(name)! You are %(year - birth) years old."
```

The compiler sees:

```dart
String.template([
  "Hi, ",
  Fn.new { name },
  "! You are ",
  Fn.new { year - birth },
  " years old."
])
```

The template becomes a list of parts. The text sections become string literals. Each template expression becomes a function that evaluates the expression and returns the result. Then this list is passed to `String.template()`.

That method does what you expect:

```dart
static template(parts) {
  var result = ""
  for (part in parts) {
    if (part is String) {
      result = result + part
    } else {
      result = result + part.call().toString
    }
  }

  return result
}
```

Since functions are closures in Wren, template expressions can use surrounding variables, `this`, the whole works.

## Tagged templates

This desugaring leads to the natural question, "Can I pass the list of parts to something other than `String.template()`"? The answer will be "Yes!".

I'm still figuring out the syntax here, but it will be something along the lines of:

```dart
Foo "Some %(template) string"
```

This gets desugared to:

```dart
Foo.template(["Some ", Fn.new { template }, " string"])
```

Now `Foo` has complete control over how those template expressions are handled. This is important because it lets you do things like:

```dart
Html "<body>%(content)</body>"
```

Then the `Html` class could, for example, ensure that the interpolated expression gets HTML sanitized before it gets placed in the resulting string.

A custom template handler could even evaluate the template parts later or multiple times, making it useful for things like data binding and other "template-y" use cases.

## Format strings

I have some rough ideas for one more small addition. Even in languages with string interpolation, doing things like custom numeric formats or padding can be a chore because you have to make it a named method call in the expression:

```
"Some hex number %(someNumber.toHexString)"
```

The format specifiers in `printf()` are actually pretty nice for this. So what I'm thinking is we can expand the syntax to allow a "format string" between the `%` and `(`:

```
"Some hex number %x(someNumber)"
```

That gets passed to the template processor too. So the above desugars to something like:

```dart
String.template([
  "Some hex number ",
  ["x", Fn.new { someNumber }]
])
```

Then the template processor can handle that as it sees fit.

## Pros

  * Much easier to create nice formatted strings out of output. It's a really handy bread-and-butter feature for little scripts and command-line programs. Likewise, for `printf()`-style debugging. Once you have interpolation, you find yourself using it *all the time*.

  * Can be used anywhere instead of baking it into certain printing APIs.

  * Because template expressions are parsed at compile time, we get all of the compile time grammar and name validation you expect.

    Unlike, say, `.format()` in Python, template expressions in Wren should be safe and fast.

## Cons

  * A consequence of the above pro is that template strings cannot be created at runtime. It's not a backdoor to eval. This is good for performance, simplify, and safety, but may be a limitation in some cases. For example, a Wren program could not load a template string from a file and then evaluate it.

  * One interesting consequence is that tokenizing is no longer [regular][]. This makes it harder to parse Wren using some simple parser generators, and may make things like syntax highlighters a little more annoying to create.

    [regular]: https://en.wikipedia.org/wiki/Regular_language

  * You have to escape `%` in strings. I thought about having a separate syntax for "template" strings and only allow %() there, but I want to avoid a proliferation of different string types. I'd rather have one string syntax that supports everything. I picked `%` because I think it's fairly uncommon so escaping should be rare.

  * One more feature to add.

## Status

Right now, I just have the basic string template syntax working. The actual code change in the VM is pretty small: 50 lines of non-comment code.

I don't think tagged templates or format strings will add much to that. (Actually implementing various numeric format strings for numbers might, though.)

I updated the benchmarks and tests to take advantage of this, and I think it really is a readability win.

What do you think?


underscorediscovery:
Everything above sits well in my view, with my understanding of wren: Clean, simple, flexible. 

String interpolation is _way_ better than vararg style formatters, 
because the value isn't separated from the natural expression being formed.

`"This %method %is clear even %(with complex expr)` 
vs
`"This %d isn't % longer", method, "especially when")`

I've been using compile time only string interpolation a few years, yet I haven't felt it missing from runtime. I think that's because I view runtime (and especially) templating engines a bit more specific use case. Most of the ones listed like mustache don't usually deal with arbitrary expressions as much as simple context -> values, and helper functions. The facilitation mentioned above would account for a runtime convenience sharing the same code. 

Feeding a context of values + functions with a string and it will run through it seems like it could be a module (if not a std library feature) anyway.

For interest sake, which you may know, 
[haxe string interpolation](http://haxe.org/manual/lf-string-interpolation.html) is bound to `'` single quote strings (where haxe has `"` as well), and `$variable` and `${expr}` are used. It's compile time only and any expression is valid.

bjorn:
> I never liked the separation between the formatted value and its position in the string.

I think the main point here is localization. Translations will commonly swap around the position placeholders. This is why Qt uses numbered placeholders, for example. Though I personally think named ones would be even better. But it can't be direct code, otherwise you can't swap the string for a translated version.

munificent:
> I've been using compile time only string interpolation a few years, yet I haven't felt it missing from runtime. I think that's because I view runtime (and especially) templating engines a bit more specific use case. 

Yeah, I feel the same way. There is still a use case for pure runtime templating (localization like @bjorn brings up is another big one), but for most uses, I think compile time is fine, and I like the compile-time syntax checking it gives. I figure the parser is already in the middle of parsing things at compile time anyway, so that's a good time to parse format strings to, instead of deferring it until later.

> Feeding a context of values + functions with a string and it will run through it seems like it could be a module (if not a std library feature) anyway.

Yes, I think that may be worth doing eventually too. With this, I'm going for sort of the 80% common use case while still getting simple enough to stick in the core VM. (I try to keep the VM really small, whereas the modules that come with the CLI aren't really size limited.)

> is bound to ' single quote strings (where haxe has " as well), and $variable and ${expr} are used.

Single quotes are an interesting idea. My hunch is that most languages that use a different kind of string for interpolation do so because they added it later and didn't want to break compatibility. Since we don't have that problem with Wren, I figured sticking with one kind of string made sense.

I considered `$` as the magic character. That's what Dart uses too. But it's often a valid identifier character too, which leads to a weird edge case with things like `"some string $foo$bar"`. I went with `%` partially to avoid that.

> I think the main point here is localization. Translations will commonly swap around the position placeholders.

Yup. This definitely isn't a good solution for localization. But then, with that case, you're often loading the template strings from a separate data file too, and this doesn't work for that either.

You *could* try to use this with localization by wrapping the string in a function:

```
var english = Fn.new {|fields| "Hi %(fields["first"]) %(fields["last"])!" }
var serious = Fn.new {|fields| "Greetings, %(fields["last"]), %(fields["first"])." }
```

Then pass that function around as a "template". But my hunch is that would be a bit of a chore.

munificent:
I landed simpler interpolation support without custom templates so we're going to leave it at that for now.

PR #305, opened 2015-09-24
Modified libuv.py file to be able to work with Python 2.6

ppearson:
So once again, some old software I'm afraid, and I'd understand if you prefer not to take this change, but Python 2.6 is still used a fair bit too.

This change makes two changes:
1. In python2_binary(), sys.version_info.major was being used, which is 2.7+ only. I've converted back to always using list accessor.
2. In run(), subprocess.check_output() was being used. I've changed the code to check the version to see if this is supported, but it's now more code, so possibly this isn't the best fix... The semantics are different in terms of what errors it raises.

I haven't been able to test this on a version of Python > 2.6 however, so...

### DIFF
--- a/util/libuv.py
+++ b/util/libuv.py
@@ -17,11 +17,11 @@ LIB_UV_DIR = "deps/libuv"
 def python2_binary():
     """Tries to find a python 2 executable"""
 
-    if sys.version_info.major == 2:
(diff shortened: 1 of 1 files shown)


munificent:
Awesome, thanks! Woo compatibility!

PR #310, opened 2015-10-19
Remove Recursive Mark from GC

### LINKED ISSUE #280: Deeply nested objects can result in a stack overflow in the garbage collector
Due to the garbage collector using recursion on the C stack to mark the objects, deeply nested objects can result in a stack overflow. On my system, the following example results in one:

```
var head

for (i in 1..400000) {
  head = { "next" : head }
}

IO.print("done")
```

gdb backtrace:

```
Program received signal SIGSEGV, Segmentation fault.
0x000000000040a92c in wrenMarkObj ()
(gdb) bt
#0  0x000000000040a92c in wrenMarkObj ()
#1  0x000000000040ab78 in wrenMarkObj ()
#2  0x000000000040ab85 in wrenMarkObj ()
#3  0x000000000040ab85 in wrenMarkObj ()
...
```

Moving to a tri-color garbage collector would fix this issue; additionally it would also act as the first step towards the garbage collector being able to work incrementally.

iwillspeak:
The previous GC implementation used a recursive mark method. This can
result in stack overflows when attempting to mark deeply nested objects (#280).

This is an initial look at removing the recursive mark method. I've added a
simple `GCState` object which keeps track of a list of 'gray' objects. Once
we have finished the initial mark of all the root objects we iterate through
the gray list, adding each objects referenced objects to the gray list.

This resolves #280

### DIFF
--- a/src/vm/wren_value.c
+++ b/src/vm/wren_value.c
@@ -37,7 +37,7 @@ DEFINE_BUFFER(Method, Method);
 static void initObj(WrenVM* vm, Obj* obj, ObjType type, ObjClass* classObj)
 {
   obj->type = type;
-  obj->marked = false;
+  obj->isDark = false;
   obj->classObj = classObj;
   obj->next = vm->first;
   vm->first = obj;
@@ -858,21 +858,21 @@ ObjUpvalue* wrenNewUpvalue(WrenVM* vm, Value* value)
 static void markClass(WrenVM* vm, ObjClass* classObj)
 {
   // The metaclass.
-  wrenMarkObj(vm, (Obj*)classObj->obj.classObj);
+  wrenGrayObj(vm, (Obj*)classObj->obj.classObj);
 
   // The superclass.
-  wrenMarkObj(vm, (Obj*)classObj->superclass);
+  wrenGrayObj(vm, (Obj*)classObj->superclass);
 
   // Method function objects.
   for (int i = 0; i < classObj->methods.count; i++)
   {
     if (classObj->methods.data[i].type == METHOD_BLOCK)
     {
-      wrenMarkObj(vm, classObj->methods.data[i].fn.obj);
+      wrenGrayObj(vm, classObj->methods.data[i].fn.obj);
     }
   }
 
-  wrenMarkObj(vm, (Obj*)classObj->name);
+  wrenGrayObj(vm, (Obj*)classObj->name);
 
   // Keep track of how much memory is still in use.
   vm->bytesAllocated += sizeof(ObjClass);
@@ -882,12 +882,12 @@ static void markClass(WrenVM* vm, ObjClass* classObj)
 static void markClosure(WrenVM* vm, ObjClosure* closure)
 {
   // Mark the function.
-  wrenMarkObj(vm, (Obj*)closure->fn);
+  wrenGrayObj(vm, (Obj*)closure->fn);
 
   // Mark the upvalues.
   for (int i = 0; i < closure->fn->numUpvalues; i++)
   {
-    wrenMarkObj(vm, (Obj*)closure->upvalues[i]);
+    wrenGrayObj(vm, (Obj*)closure->upvalues[i]);
   }
 
   // Keep track of how much memory is still in use.
@@ -900,7 +900,7 @@ static void markFiber(WrenVM* vm, ObjFiber* fiber)
   // Stack functions.
   for (int i = 0; i < fiber->numFrames; i++)
   {
-    wrenMarkObj(vm, fiber->frames[i].fn);
+    wrenGrayObj(vm, fiber->frames[i].fn);
   }
 
   // Stack variables.
@@ -913,12 +913,12 @@ static void markFiber(WrenVM* vm, ObjFiber* fiber)
   ObjUpvalue* upvalue = fiber->openUpvalues;
   while (upvalue != NULL)
   {
-    wrenMarkObj(vm, (Obj*)upvalue);
+    wrenGrayObj(vm, (Obj*)upvalue);
     upvalue = upvalue->next;
   }
 
   // The caller.
-  wrenMarkObj(vm, (Obj*)fiber->caller);
+  wrenGrayObj(vm, (Obj*)fiber->caller);
   wrenMarkValue(vm, fiber->error);
 
   // Keep track of how much memory is still in use.
@@ -954,7 +954,7 @@ static void markForeign(WrenVM* vm, ObjForeign* foreign)
 
 static void markInstance(WrenVM* vm, ObjInstance* instance)
 {
-  wrenMarkObj(vm, (Obj*)instance->obj.classObj);
+  wrenGrayObj(vm, (Obj*)instance->obj.classObj);
 
   // Mark the fields.
   for (int i = 0; i < instance->obj.classObj->numFields; i++)
@@ -1002,7 +1002,7 @@ static void markModule(WrenVM* vm, ObjModule* module)
     wrenMarkValue(vm, module->variables.data[i]);
   }
 
-  wrenMarkObj(vm, (Obj*)module->name);
+  wrenGrayObj(vm, (Obj*)module->name);
 
   // Keep track of how much memory is still in use.
   vm->bytesAllocated += sizeof(ObjModule);
@@ -1030,20 +1030,9 @@ static void markUpvalue(WrenVM* vm, ObjUpvalue* upvalue)
   vm->bytesAllocated += sizeof(ObjUpvalue);
 }
 
-void wrenMarkObj(WrenVM* vm, Obj* obj)
+static void darkenObject(WrenVM* vm, Obj* obj)
 {
-  if (obj == NULL) return;
-
-  // Stop if the object is already marked so we don't get stuck in a cycle.
-  if (obj->marked) return;
-
-  // It's been reached.
-  obj->marked = true;
-
 #if WREN_DEBUG_TRACE_MEMORY
-  static int indent = 0;
-  indent++;
-  for (int i = 0; i < indent; i++) printf("  ");
   printf("mark ");
   wrenDumpValue(OBJ_VAL(obj));
   printf(" @ %p\n", obj);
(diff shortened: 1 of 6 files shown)


src/vm/wren_value.c
```diff
@@ -1021,25 +1021,17 @@ static void markString(WrenVM* vm, ObjString* string)
...
-  if (obj == NULL) return;
-
-  // Stop if the object is already marked so we don't get stuck in a cycle.
-  if (obj->marked) return;
-
-  // It's been reached.
-  obj->marked = true;
-
 #if WREN_DEBUG_TRACE_MEMORY
   static int indent = 0;
```
iwillspeak: Seen as we don't visit each object recursively any more this indented trace output doesn't really work. I can't really work out how to solve this at the moment i'm suggesting just removing the indentation completely from the tracing.

iwillspeak:
This change adds another pointer to the `Obj` struct. I'm thinking that if we moved to a doubly-linked list for the currently allocated objects we could remove each object from the old list as we marked it and add it to a new one stored as part of the `GCState`. Once the `expandMarks` phase is complete all we would need to do is free all of the objects left in the old list and update the `WrenVM` structure to point at the new list. I'm not sure if that change would be worth the effort though at this stage.

iwillspeak:
Another option is to remove the `marked` field and rely on a non-null `gcNext` pointer to determine marked objects.

src/vm/wren_value.c
munificent: Yeah, getting rid of the indentation is totally fine.

src/vm/wren_value.c
```diff
@@ -1071,17 +1063,43 @@ void wrenMarkObj(WrenVM* vm, Obj* obj)
...
+    expandMark(vm, obj, state);
+  }
+}
+
+void wrenMarkObj(WrenVM* vm, Obj* obj, GCState* state)
+{
+  if (obj == NULL) return;
+
+  // Stop if the object is already marked so we don't get stuck in a cycle.
+  if (obj->marked) return;
```
munificent: How about renaming this field `isBlack` to make the tri-color abstraction more explicit?

src/vm/wren_value.h
```diff
@@ -599,6 +602,12 @@ typedef union
   double num;
 } DoubleBits;
 
+// A struct to store objects for garbage collection
+typedef struct
+{
+  Obj* first;
+} GCState;
```
munificent: I think we can get rid of this. Either:

1. Just pass around a `Obj**`.
2. Add a `Obj* gray` field to WrenVM. That would avoid having to pass it around everywhere. It's a bit weird since the field is only ever used during collection, but it will probably be handy later when we move to incremental.

src/vm/wren_value.h
```diff
@@ -114,6 +114,9 @@ typedef struct sObj
 
   // The next object in the linked list of all currently allocated objects.
   struct sObj* next;
+
+  // The next object in the GC gray set.
+  struct sObj* gcNext;
```
munificent: It would be nice to avoid adding another field to every object if possible. The simplest solution is probably to store the gray set as an explicit collection. We can make an `Obj*` Buffer and add and remove gray objects to that.

What do you think?

munificent:
This looks really nice!

iwillspeak:
Ok. So I've got rid of the `GCState` object and renamed some of the variables from your suggestion. I've not moved the gray set to the main vm yet as I'm not sure if we want to be allocating/resizing buffers during a collection if this is just supposed to be a temporary GC implementation.

This is what the benchmark of the current changes looks like against `master`:

```
Switched to branch 'master'
Your branch is up-to-date with 'origin/master'.
generating baseline
binary_trees - wren            .......... 0.21s 0.0032 no baseline
delta_blue - wren              .......... 0.13s 0.0038 no baseline
fib - wren                     .......... 0.21s 0.0018 no baseline
fibers - wren                  .......... 0.25s 0.0041 no baseline
for - wren                     .......... 0.08s 0.0020 no baseline
method_call - wren             .......... 0.11s 0.0015 no baseline
map_numeric - wren             .......... 0.34s 0.0047 no baseline
map_string - wren              .......... 0.12s 0.0066 no baseline
string_equals - wren           .......... 0.19s 0.0009 no baseline
Switched to branch 'iterative-gc'
Your branch is up-to-date with 'origin/iterative-gc'.
binary_trees - wren            .......... 0.22s 0.0024  96.89% relative to baseline
delta_blue - wren              .......... 0.12s 0.0024 102.12% relative to baseline
fib - wren                     .......... 0.21s 0.0016 101.57% relative to baseline
fibers - wren                  .......... 0.25s 0.0056  98.71% relative to baseline
for - wren                     .......... 0.08s 0.0005  98.96% relative to baseline
method_call - wren             .......... 0.11s 0.0012 100.52% relative to baseline
map_numeric - wren             .......... 0.33s 0.0045 100.45% relative to baseline
map_string - wren              .......... 0.11s 0.0140 106.32% relative to baseline
string_equals - wren           .......... 0.19s 0.0075 102.06% relative to baseline
```

Looks like some things are a bit slower. Guessing it's all the pointer shuffling we end up doing in `wrenExpandMarks`. Looks pretty good overall though.

iwillspeak:
I think I understand how to move the gray set into the main `WrenVM` object. It should mean we can get rid of the `next`, `gcNext` and `isBlack` fields from the object.

As objects are allocated we add them to the end of an array, resizing when required. When we GC we walk two pointers `black` and `gray` through the array until there are no more gray objects (`black == gray`). We can then collect all of the remaining objects in the array. Each time we mark some thing we switch it with the object at the `gray` pointer and increment the `gray` pointer. That would mean the only time we need to reallocate the array would be when we were also allocating a new object.

If you're happy with the current linked list implementation I'll squash the commits down, if not I can experiment with what I've described above this evening (UK time).

iwillspeak:
I had a hack about with storing the gray set as a collection on the `WrenVM` structure. It turned out to be a bit more complex than I first thought. We don't have enough information to do the expansion of the gray set in place so we would need to copy live objects into a new buffer in the GC sweep and switch the pointer over when we have freed up all of the unreachable objects.

munificent:
> It should mean we can get rid of the next, gcNext and isBlack fields from the object.

We'll still need `next` because without that, unreferenced objects are completely lost and unreachable. We wouldn't be able to `free()` them.

We also still need `isBlack` (or `isDark` since it's true for gray objects too). We need a fast way to tell if an object is already in the work list so we don't get stuck in cycles and I think an explicit mark bit is the best way to do that.

I do think we can get rid of `gcNext`, which I'd like to do. Making all objects larger even not during GC is something I hope to avoid. (And, in fact, now that I think about it, at some point we could be really clever by stuffing the mark bit in the low bit of the `next` pointer.)

> As objects are allocated we add them to the end of an array, resizing when required.

Ah, interesting! Maybe this is just because it's what I'm used to, but I think I like using a linked list in the objects for the main collection of objects. Though, you're right, we *could* use an array of pointers for that.

> Each time we mark some thing we switch it with the object at the gray pointer and increment the gray pointer.

Clever! I could be wrong, but I don't think it will work, though. :( When we are graying the roots, we don't know their index in the array, and we don't want to do an O(n) seek to find them.

What I have in mind is something much simpler relative to the current code:

1. We add a buffer of Obj* to WrenVM. It will only be used during collection and will be empty otherwise. This is the stack for the depth-first traversal of the marked objects.

2. We separate out marking into steps, like you have: marking and expanding. (I think it may be clearer to rename these to "gray" and "blacken" respectively.)

3. Marking just pushes the object onto that stack.

4. Traversing walks an object's fields and marks them.

4. Mark all the roots. So now the stack is populated with the work list.

5. Until the work list is empty, we pop the last object and traverse it.

I *think* that would do a straightforward replacement of the old recursive code with a non-recursive form. It does do dynamic allocation during collection, but I'm OK with that. The recursive code was effectively doing the same thing too since it grows the C stack. This just gives us more control.

In practice, I think the work list will be pretty small. It's maximum size is the length of the deepest non-cyclical object reference chain. And, outside of pathological code, that tends to be pretty small.

Oh, to try to make things a bit easier, I landed a couple of small patches on master: the test runner will timeout, and I added a benchmark that invokes the GC so we can do timing comparisons. (Most of the existing benchmarks use little enough memory that the GC never runs.)

What do you think?

iwillspeak:
> Clever! I could be wrong, but I don't think it will work, though. :( When we are graying the roots, we don't know their index in the array, and we don't want to do an O(n) seek to find them.

Yep. Unfortunately we don't have enough information to pull this off.

Thanks for the more detailed explanation. I've implemented a simple centralised stack as you suggested. Seems to work pretty well so far, and it's actually a lot simpler than many of my earlier attempts to get this working.

I guess moving to a centralised array of pointers for the allocated objects isn't really needed.

iwillspeak:
Here's the latest set of benchmarks, after rebasing to pick up your latest changes on master:

```
Switched to branch 'master'
Your branch is up-to-date with 'origin/master'.
generating baseline
binary_trees - wren            .......... 0.22s 0.0215 no baseline
binary_trees_gc - wren         .......... 0.28s 0.0176 no baseline
delta_blue - wren              .......... 0.13s 0.0040 no baseline
fib - wren                     .......... 0.22s 0.0124 no baseline
fibers - wren                  .......... 0.26s 0.0099 no baseline
for - wren                     .......... 0.08s 0.0066 no baseline
method_call - wren             .......... 0.11s 0.0035 no baseline
map_numeric - wren             .......... 0.33s 0.0058 no baseline
map_string - wren              .......... 0.12s 0.0053 no baseline
string_equals - wren           .......... 0.19s 0.0061 no baseline
Switched to branch 'iterative-gc'
Your branch is up-to-date with 'origin/iterative-gc'.
binary_trees - wren            .......... 0.22s 0.0045  99.94% relative to baseline
binary_trees_gc - wren         .......... 0.29s 0.0047  97.69% relative to baseline
delta_blue - wren              .......... 0.13s 0.0047  99.80% relative to baseline
fib - wren                     .......... 0.22s 0.0021 100.31% relative to baseline
fibers - wren                  .......... 0.26s 0.0068 101.44% relative to baseline
for - wren                     .......... 0.08s 0.0049 101.39% relative to baseline
method_call - wren             .......... 0.11s 0.0027  99.71% relative to baseline
map_numeric - wren             .......... 0.33s 0.0048 100.34% relative to baseline
map_string - wren              .......... 0.12s 0.0098 103.63% relative to baseline
string_equals - wren           .......... 0.19s 0.0075 103.91% relative to baseline
```

iwillspeak:
Your new benchmark is looking good. I'm quite happy with how this has turned out. I've squashed the commits down to a single commit if you're ready to merge this?

iwillspeak:
fixes #280

munificent:
This is merged in now!

I'm not sure why GitHub doesn't automatically recognize that. <shrug>

I had to make a few tweaks to it, but just minor stuff. The main issues was that calling `wrenReallocate()` to grow the gray set can itself recursively trigger a GC call. So now it directly calls the allocate function.

One really nice thing about this change is that it effectively does "tail call elimination" when traversing references. The deeply nested GC test, for example, only uses a gray stack of size 64 because it pops each link off the stack *before* traversing into it. So it's not just that this change uses the heap for the gray stack. It actually uses way less memory in most cases.

Great job!

iwillspeak:
Nice. Thanks. I think GitHub is confused because I rebased yesterday to get the new GC test, but you merged in the older version of the pull request.

PR #312, opened 2015-11-18
Update Visual Studio projects

Rohansi:
The Visual Studio projects were out of sync (and the CLI one was removed!) and this updates them so we can compile in VS again.

There were some issues in the code that I'm not 100% sure how to handle:
* `uv_fs_t::bufs` is accessed directly but that doesn't work in VC++, I have no idea what the proper way to do this in libuv so I just used what compiled
* `S_IRUSR` is not defined on Windows, and I don't know what it would map to. Replaced with 0 for now because I don't think it's used unless `O_CREAT` is given. [This might be useful.](http://stackoverflow.com/a/593017/1056845)
* The macros used in `modules.c` were causing syntax errors because of the nested `{}` in the object initializers. Removing them seems to work but I haven't tested on other compilers, let's see what the CI says about it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,10 +23,6 @@ xcuserdata/
 !wren/
 !wren_lib/
 
-# Visual Studio intermediate files.
-Debug/
-Release/
-
 # Visual Studio cache files.
 ipch/
 *.aps
--- a/.gitignore
+++ b/.gitignore
@@ -19,6 +19,14 @@ test/benchmark/baseline.txt
 xcuserdata/
 *.xccheckout
 
+# Allow Visual Studio project files.
+!wren/
+!wren_lib/
+
+# Visual Studio intermediate files.
+Debug/
+Release/
+
 # Visual Studio cache files.
 ipch/
 *.aps
--- a/src/module/io.c
+++ b/src/module/io.c
@@ -9,6 +9,10 @@
 
 #include <stdio.h>
 
+#ifdef _WIN32
+  #include <fcntl.h>
+#endif
+
 static const int stdinDescriptor = 0;
 
 // Handle to Stdin.onData_(). Called when libuv provides data on stdin.
@@ -111,7 +115,7 @@ void fileOpen(WrenVM* vm)
   uv_fs_t* request = createRequest(wrenGetArgumentValue(vm, 2));
   
   // TODO: Allow controlling flags and modes.
-  uv_fs_open(getLoop(), request, path, O_RDONLY, S_IRUSR, openCallback);
+  uv_fs_open(getLoop(), request, path, O_RDONLY, 0, openCallback);
 }
 
 // Called by libuv when the stat call for size completes.
@@ -171,7 +175,7 @@ static void fileReadBytesCallback(uv_fs_t* request)
 {
   if (handleRequestError(request)) return;
 
-  uv_buf_t buffer = request->bufs[0];
+  uv_buf_t buffer = request->fs.info.bufs[0];
   WrenValue* fiber = freeRequest(request);
 
   // TODO: Having to copy the bytes here is a drag. It would be good if Wren's
--- a/src/module/io.c
+++ b/src/module/io.c
@@ -8,16 +8,13 @@
 #include "wren.h"
 
 #include <stdio.h>
+#include <fcntl.h>
 
-#ifdef _WIN32
-  #include <fcntl.h>
-#endif
-
-typedef struct sFileReqData
+typedef struct sFileRequestData
 {
   WrenValue* fiber;
   uv_buf_t buffer;
-} FileReqData;
+} FileRequestData;
 
 static const int stdinDescriptor = 0;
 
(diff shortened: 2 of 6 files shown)


Rohansi:
Ugh VC++, sort yourself out! Probably going to need a macro for `{}` that uses `{0}` if using VC++.

Rohansi:
OK, I had no idea VC++ was *this* bad at C. However, `uv_fs_t::buf` isn't a public member according to [libuv's documentation](http://docs.libuv.org/en/v1.x/fs.html#public-members), so we shouldn't be using it anyway.

Maybe storing a pointer to it in `uv_fs_t::data` would be a proper way? It's there for user-defined data, after all.

Rohansi:
Hmm, no, that's already being used to store the fiber.

Maybe we need to expand the amount allocated by `wrenAllocateForeign`? I'm not sure how that works but it's used to store the file descriptor so it should live long enough, but what happens if two reads are done in parallel for the same `File`? It could be overwritten!

Expanding what's stored in `uv_fs_t::data` might be the best option.

.gitignore
```diff
@@ -19,6 +19,14 @@ test/benchmark/baseline.txt
 xcuserdata/
 *.xccheckout
 
+# Allow Visual Studio project files.
+!wren/
+!wren_lib/
+
+# Visual Studio intermediate files.
+Debug/
+Release/
```
munificent: Would you be OK with tweaking the VS project to put these under `build/debug` and `build/release`? That matches the other platforms and makes the docs that talk about build outputs correct across Mac/Linux/Windows.

.gitignore
```diff
@@ -19,6 +19,14 @@ test/benchmark/baseline.txt
 xcuserdata/
 *.xccheckout
 
+# Allow Visual Studio project files.
+!wren/
+!wren_lib/
```
munificent: Are these needed? These aren't at the top level anyway, are they?

src/cli/modules.c
```diff
@@ -67,11 +67,19 @@ typedef struct
   ClassRegistry classes[MAX_CLASSES_PER_MODULE];
 } ModuleRegistry;
 
+// Visual C++ doesn't support initializing arrays with {} but we need to use it
+// for compilers that actually follow the standard.
+#ifndef _MSC_VER
+  #define EMPTY {}
+#else
+  #define EMPTY {0}
+#endif
```
munificent: Aw, man, VC++ is always a pain. I tried a different approach to this code that I think avoids this problem. Can you merge in 69a5c1d017ca185e6466eb848cc76cff55aa60b2 and see if that helps?

src/module/io.c
```diff
@@ -77,8 +88,13 @@ static bool handleRequestError(uv_fs_t* request)
 // Allocates a new request that resumes [fiber] when it completes.
 uv_fs_t* createRequest(WrenValue* fiber)
 {
-  uv_fs_t* request = (uv_fs_t*)malloc(sizeof(uv_fs_t));
-  request->data = fiber;
+  char* buffer = (char*)malloc(sizeof(uv_fs_t) + sizeof(FileReqData));
```
munificent: Instead of a single concatenated alloc, I think it's clearer to allocate the `FileReqData` separately (and free it below).

src/module/io.c
```diff
@@ -9,6 +9,16 @@
 
 #include <stdio.h>
 
+#ifdef _WIN32
+  #include <fcntl.h>
+#endif
```
munificent: I think this can be a regular unguarded `#include` up with the others. It was probably a mistake to omit it on the other platforms. :)

src/module/io.c
```diff
@@ -9,6 +9,16 @@
...
+
+typedef struct sFileReqData
+{
+  WrenValue* fiber;
+  uv_buf_t buffer;
+} FileReqData;
```
munificent: Nit: How about `FileRequestData`? I try to avoid most acronyms.

munificent:
Thanks for doing all the work on this! The `.bufs` thing is annoying. I wish I'd realized I was doing the wrong thing there.

.gitignore
Rohansi: The current gitignore ignores `wren` in build outputs so the first one might be required. There is probably a better solution to this, though.

.gitignore
Rohansi: Yeah, I can do that.

`wren.exe` and `wren*.lib` get sent to `build/vs/Debug` right now. Those should be in `bin/` and `lib/` right? Are debug and release separated (`bin/debug/`)?

src/cli/modules.c
Rohansi: That works! I'll drop my changes there.

.gitignore
Rohansi: Also, none of the entries in gitignore are top level only. I suppose fixing that would be the preferred way to get rid of these entries.

The CLI project was accidentally removed from the repository when you moved it to another directory because anything with the name `wren` was ignored. Including the `wren` folder containing the CLI project.

munificent:
Awesome, thanks! :metal:

I went ahead a tweaked the .gitignore file too. I was accidentally using a trailing `/` when I meant to use a leading one in a few places. :confused:

PR #319, opened 2016-02-08
Update wrenInterpret doc/site/embedding-api

4d47:
Was reading [embedding-api](http://wren.io/embedding-api.html#executing-wren-code) and found the doc for `wrenInterpret` out of sync with 3b09b35b593d6298b1615e883b3ceab367675aab.


munificent:
Awesome, thanks! There's still lots of work to do to get the embedding docs fleshed out, but this is an important step!

PR #320, opened 2016-02-12
Fix two small code samples in doc

4d47:
I'm pretty much done reading manual so there shouldn't be any more of those :-)


munificent:
Awesome, thanks!

Issue #325, opened 2016-02-23
Relative import?


Qard:
I'm trying to figure out a way to do a node.js-style relative import, but `loadModuleFn` doesn't seem to be quite enough on its own. I can adjust the base directory before the import, but can't figure out how to restore it again after the module fiber runs. I need something to hook into there for sequential import statements to be rooted at the correct level. Any ideas?

munificent:
> I'm trying to figure out a way to do a node.js-style relative import, but loadModuleFn doesn't seem to be quite enough on its own.

Right, it's context free. You know the import string, but not where it appears.

This is sort of by design, though it may not ultimately be the right design. I like the idea of an import saying *what* you want to import more so than *where you want to go from the current module to find it*. Being context-free in that way makes it easier to rearrange files, support package managers, etc.

It's, more or less, how Java, C#, and C/C++ tend to work, if you wave your hands around include paths, classpaths, .cvproj files, etc.

But it can also be tedious and verbose in the presence of a lot of nesting, and it's not clear how the CLI should locate the "root" that absolute imports are imported relative from. (Right now, it's the directory where the main script appears, but that's not ideal.)

The import system in Wren is really rudimentary right now. I wanted to get enough in place to start experimenting. And, more important to me at the time, start playing with the actual runtime semantics of modules and scope. But it definitely needs a lot of work and probably a lot of experimentation.

One option to try might be to have a syntax that means "relative import". Perhaps a leading dot in the import string. Then, when the VM sees that, it prepends the containing module's path. Though, once we do that, we'll want to be able to back out of directories too. We'll likely end up with something like [Python's syntax for relative imports](https://docs.python.org/2.5/whatsnew/pep-328.html).

I'm not sure if that's a good thing or a bad thing. What do you think?

Qard:
Yeah, I had actually been thinking ./ for relative and bare to search a wren_modules folder, like node. The relative search doesn't currently work though. :(

PR #327, opened 2016-02-23
node-style import

Qard:
This is the beginnings of a node-style import system I'm working on. At this stage, it's usable enough simply supporting a wren_modules folder in the app root. I'd like to get relative import supported also though, for fully node-like nested dependency resolution. See: #325.

I'm hoping that, if I can get nested dependency resolution working, I can build a proper package manager for wren.

### DIFF
--- a/src/cli/vm.c
+++ b/src/cli/vm.c
@@ -86,11 +86,8 @@ static char* wrenFilePath(const char* name)
 //
 // Returns it if found, or NULL if the module could not be found. Exits if the
 // module was found but could not be read.
-static char* readModule(WrenVM* vm, const char* module)
+static char* loadModule(WrenVM* vm, const char* module)
 {
-  char* source = readBuiltInModule(module);
-  if (source != NULL) return source;
-  
   // First try to load the module with a ".wren" extension.
   char* modulePath = wrenFilePath(module);
   char* moduleContents = readFile(modulePath);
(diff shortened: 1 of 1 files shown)


munificent:
Ah, very interesting! Having more powerful import resolution so that we can eventually do a package manager is definitely on the radar. Thanks for hacking on this!

However, I'm not personally a fan of node's nested dependency approach. I wrote the [package manager for Dart](https://pub.dartlang.org/) and it's based on Bundler's model, which [I think is a much better way to do things](https://www.dartlang.org/tools/pub/versioning.html). Even the npm folks seem to be realizing that [nested dependencies aren't all they're cracked up to be](https://nodejs.org/en/blog/npm/peer-dependencies/).

For lots of reasons, I think a flattened dependency graph works better, though it also has its own challenges around constraint resolution. Most package managers I've seen seem to either use Bundler's model or be going in the direction of doing so.

The interesting question is how much support do we need to bake into the Wren CLI in order to support something like Bundler. Should it just walk up directories looking for a lockfile that it can parse? What do you think?

Qard:
I agree that nested dependencies are complicated, and peer dependencies was definitely a mistake.

Personally, I like the current model of npm, supporting nested dependencies, but trying to flatten as much as possible. I'm good with whatever people are comfortable with though. :smile_cat:

Another thing I want to consider for package management is how to support native code. The mruby way of bundling the modules directly into the build is probably better for supporting embedded environments than the node way of loading dynamic libraries. That'd probably add more complexity to supporting nested packages.

src/cli/vm.c
```diff
@@ -116,6 +113,33 @@ static char* readModule(WrenVM* vm, const char* module)
...
+// module was found but could not be read.
+static char* readModule(WrenVM* vm, const char* module)
+{
+  char* source = readBuiltInModule(module);
+  if (source != NULL) return source;
+  
+  source = loadModule(vm, module);
+  if (source != NULL) return source;
+  
+  // If searching in the current path fails, look
```
paulsapps: in wren_modules/ ?

src/cli/vm.c
```diff
@@ -116,6 +113,33 @@ static char* readModule(WrenVM* vm, const char* module)
...
+  char* source = readBuiltInModule(module);
+  if (source != NULL) return source;
+  
+  source = loadModule(vm, module);
+  if (source != NULL) return source;
+  
+  // If searching in the current path fails, look 
+  size_t moduleLength = strlen(module);
+  size_t extModuleLength = moduleLength + 13;
+  char* extModule = (char*)malloc(extModuleLength + 1);
```
paulsapps: where does this get freed?

Qard: Umm...nowhere? 😨

src/cli/vm.c
Qard: Yes. I guess I accidentally the rest of the comment. 😉

underscorediscovery:
This PR appears to be out of phase with the latest Wren development, so I'm gonna close this. Discussions about the Wren CLI will soon be on their own repo, where we can pick up the discussion over there.

Issue #317, opened 2015-12-21
How should method collisions be handled?


munificent:
Right now, Wren lets you define the same method twice in the same class:

```
class Foo {
  foo() {}
  foo() {}
}
```

There's no error here and the last one wins. This seems less than optimal. :(

Detecting this case is kind of a pain, but I think we want to do it before we get stuck with *having* to allow duplicate definitions. We'll probably just need a buffer or map of signature strings in ClassCompiler for the methods that have been defined so far, and check in there for before a method is defined.

bjorn:
Raising some kind of redefinition error seems helpful to me.

minirop:
a buffer of int could do it. containing the index of the signature in the methodNames array.

munificent:
> a buffer of int could do it. containing the index of the signature in the methodNames array.

Ah, right. I forgot we keep the signatures around already. Great idea!

PR #328, opened 2016-02-24
Forbid duplicate methods

Qard:
Here's a possible fix for #317. Not sure if you'll want all that extra state in ClassCompiler though. I'm new to the codebase, so I'm not yet sure if there's a better way.

Probably needs a bit of cleanup too.

### DIFF
--- a/src/vm/wren_compiler.c
+++ b/src/vm/wren_compiler.c
@@ -290,8 +290,11 @@ typedef struct
 // Bookkeeping information for compiling a class definition.
 typedef struct
 {
+  ObjString* name;
+  
   // Symbol table for the fields of the class.
   SymbolTable fields;
+  SymbolTable methods;
 
   // True if the class being compiled is a foreign class.
   bool isForeign;
@@ -3116,6 +3119,18 @@ static bool method(Compiler* compiler, Variable classVariable)
   char fullSignature[MAX_METHOD_SIGNATURE];
   int length;
   signatureToString(&signature, fullSignature, &length);
+  
+  // Check for duplicate methods
+  StringBuffer* methods = &compiler->enclosingClass->methods;
+  int symbol = wrenSymbolTableFind(methods,
+    fullSignature, length);
+  if (symbol != -1)
+  {
+    error(compiler, "Already defined method '%s' for class %s in module '%s'.",
+      fullSignature, &compiler->enclosingClass->name->value,
+      compiler->parser->module->name);
+  }
+  wrenSymbolTableAdd(compiler->parser->vm, methods, fullSignature, length);
 
   if (isForeign)
   {
@@ -3162,6 +3177,9 @@ static void classDefinition(Compiler* compiler, bool isForeign)
   classVariable.scope = compiler->scopeDepth == -1 ? SCOPE_MODULE : SCOPE_LOCAL;
   classVariable.index = declareNamedVariable(compiler);
   
+  ObjString* className = AS_STRING(wrenNewString(compiler->parser->vm, 
+    compiler->parser->previous.start, compiler->parser->previous.length));
+  
   // Make a string constant for the name.
   emitConstant(compiler, wrenNewString(compiler->parser->vm,
       compiler->parser->previous.start, compiler->parser->previous.length));
@@ -3200,12 +3218,14 @@ static void classDefinition(Compiler* compiler, bool isForeign)
 
   ClassCompiler classCompiler;
   classCompiler.isForeign = isForeign;
+  classCompiler.name = className;
 
   // Set up a symbol table for the class's fields. We'll initially compile
   // them to slots starting at zero. When the method is bound to the class, the
   // bytecode will be adjusted by [wrenBindMethod] to take inherited fields
   // into account.
   wrenSymbolTableInit(&classCompiler.fields);
+  wrenSymbolTableInit(&classCompiler.methods);
   compiler->enclosingClass = &classCompiler;
 
   // Compile the method definitions.
@@ -3230,6 +3250,7 @@ static void classDefinition(Compiler* compiler, bool isForeign)
   }
   
   wrenSymbolTableClear(compiler->parser->vm, &classCompiler.fields);
+  wrenSymbolTableClear(compiler->parser->vm, &classCompiler.methods);
   compiler->enclosingClass = NULL;
   popScope(compiler);
 }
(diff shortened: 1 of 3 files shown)


Qard:
Eep. Doesn't seem to differentiate between statics and methods yet...I'll dig into some more on Thursday. :disappointed:

munificent:
Thanks for hacking on this! There's a more memory efficient way to handle this. We don't need to create a full symbol table which maps names to symbols. To find duplicates, all we need is the numeric symbols.

Instead, give ClassCompiler and IntBuffer that just contains the numeric symbols of the methods we've defined so far. (Or, I suppose, two of them, one for instance and one for static methods.)

This should simplify the code too, since you won't have to deal with creating a heap-allocated string for the method signature.

Qard:
Okay, I got it working now for static and non-static methods, and switched to using IntBuffer, as suggested. Still needs tests. Could also use some review.

I made a generic `wren##name##BufferFind` function for this. Not sure if you'd prefer specialization, since it's currently only used for this particular use of IntBuffer. Seems like it might be useful for other things in the future though. Also, I called it *Find due to it basically being a copy/paste of wrenSymbolTableFind, but it seems to me less a "find" and more an "index of" so the naming feels a little wrong to me. Thoughts?

src/vm/wren_compiler.c
```diff
@@ -3215,6 +3234,9 @@ static void classDefinition(Compiler* compiler, bool isForeign)
   classVariable.scope = compiler->scopeDepth == -1 ? SCOPE_MODULE : SCOPE_LOCAL;
   classVariable.index = declareNamedVariable(compiler);
   
+  ObjString* className = AS_STRING(wrenNewString(compiler->parser->vm, 
+    compiler->parser->previous.start, compiler->parser->previous.length));
```
munificent: We create a string for the class name a few lines down in the call to `emitConstant()`, so we should just reuse that object. We can probably just store another pointer straight to it in ClassCompiler, though it feels a little iffy to me to have a pointer to a GC object that we don't trace. It will do the right thing since there will always be another pointer to the same string coming from the constant table, but it feels weird. It's probably enough to just leave a comment on className explaining this.

src/vm/wren_utils.h
```diff
@@ -64,6 +65,15 @@ typedef struct {
...
       wren##name##BufferFill(vm, buffer, data, 1); \
+    } \
+    int wren##name##BufferFind(name##Buffer* buffer, type data) \
+    { \
+      for (int i = 0; i < buffer->count; i++) \
+      { \
+        if (memcmp(&buffer->data[i], &data, sizeof(data)) == 0) return i; \
+      } \
+      \
+      return -1; \
```
munificent: Since this is only used in one place, at least right now, let's just make a little standalone static function in wren_compiler.c just for the use case we need. The Buffer stuff isn't really like an OOP class where every "method" that works on it needs to be part of it. It's fine to have unique functions that poke at buffers of specific kinds.

src/vm/wren_compiler.c
```diff
@@ -3202,6 +3206,21 @@ static bool method(Compiler* compiler, Variable classVariable)
...
+  IntBuffer* methods = isStatic ? &compiler->enclosingClass->staticMethods
+                                : &compiler->enclosingClass->instanceMethods;
+  
+  int symbol = wrenIntBufferFind(methods, methodSymbol);
+  if (symbol != -1)
+  {
+    const char* methodType = isStatic ? "static method" : "method";
+    error(compiler, "Already defined %s '%s' for class %s in module '%s'.",
+      methodType, fullSignature, &compiler->enclosingClass->name->value,
+      compiler->parser->module->name);
```
munificent: Compile errors always include the module name, so you can leave that off here.

munificent:
Added a couple of comments, but overall this looks rad! Of course, we definitely want tests before we land it. :) I'm thinking:

* Test that you get an error with two instance methods with the same signature.
* Same for static methods.
* No error if a static and instance method have the same signature.

We could conceivably test that you don't get an error with two methods with the same name but different signatures like `foo(_)` and `foo(_,_)`, but that's overkill. We rely on that behavior all over the place, so *nothing* would work if that was an error. :)

Let me know if you want me to help out with the tests.

src/vm/wren_compiler.c
Qard: Yeah, I was thinking that too, but I figured it was safest to just make a separate string since I don't yet fully understand what happens with that value in the VM.

Can you provide a little more detail on exactly what would happen so I can comment it accurately?

src/vm/wren_utils.h
Qard: Ok, I'll do that. :smile_cat:

src/vm/wren_compiler.c
Qard: Ok, I basically copy/pasted most of this out of [an error in vm.c](https://github.com/munificent/wren/blob/master/src/vm/wren_vm.c#L339) which _is_ passing in the module. I hadn't noticed the compiler error function was already doing that.

Qard:
I'll take another pass at this tomorrow night. I'd like to dig into the tests too, so I can familiarize myself with how they are laid out. I'd be happy to receive any guidance you can provide though. :smile_cat:

src/vm/wren_compiler.c
munificent: Yeah, runtime errors are a little different. :) Compile errors handle it automatically because at compile time we know the most recently consumed token and can infer source position based on that.

munificent:
> I'd like to dig into the tests too, so I can familiarize myself with how they are laid out.

Hopefully it's pretty straightforward, though now that I think about it, we could use a little README explaining the format for asserts (`expect error`, `expect: ...`, etc.).

> I'd be happy to receive any guidance you can provide though.

You probably aren't the only one who would benefit from this, so I'll see if I can find a bit of time to write a quick doc on it. If I don't get to that in time, feel free to bug me and I'll help you out.

Thanks!

Qard:
Still haven't got back to this. I should have some time to work on it this weekend though. I haven't forgotten it!

Qard:
Finally got around to updating this PR and adding tests. Seems to all be working. Not sure what the appveyor failure is about though--looks unrelated to me.

munificent:
Awesome, thanks! If you haven't already, feel free to send me a patch to add yourself to the AUTHORS file. You're part of Wren now! :)

Qard:
Yay! :tada:

PR #331, opened 2016-03-03
wrenNewVM NULL check

MADGameStudio:
My first attempt to embed VM in project ended with crash. Fortunately I'm not so easily scared and managed to found out that despite comment in "wren.h" no check for null is performed. I took the liberty of fixing that. 
Btw. nice project, keep up the good work!


munificent:
Aww, I'm sorry you ran into a crash. :( The docs for `wrenNewVM()` were totally bogus.

Thanks for fixing this! I tweaked the code a bit and went ahead and added a test too. Thanks for contributing to Wren! :bird:

Issue #265, opened 2015-05-15
Provide access to compile/parse errors generated during wrenInterpret()


preble:
Wren presently outputs parse errors to stderr. I would like to see an API to access the text of these errors programmatically, either by querying for them after a failed `wrenInterpret()`, or (better?) via callback supplied in `WrenConfiguration`. If the user is able to supply their own Wren code (as I would like them to be able to do) we will need a way to show them the errors.

The callback approach might work as a way to report errors during `wrenCall()` as well?

munificent:
Yes! This has always been my intention, I just haven't gotten to it. A callback in `WrenConfiguration` is exactly what I had in mind.

If you'd like to take a stab at implementing that, I'd be happy to guide you through it. I'm not sure if the varargs stuff will make it harder, but we should be able to figure it out.

preble:
I'd be happy to take a shot at it, thanks!

Would this callback be specifically for `wrenInterpret()`, or should it be more general – for runtime errors from `wrenCall()`, too? Only difference I can think of at the moment is that it'd be great for `wrenInterpret()`'s callback to provide line/column information via a struct. With `wrenCall()` that kind of information isn't applicable.

munificent:
> With wrenCall() that kind of information isn't applicable.

Right. On the other hand, callstack information is.

How about we start with one and then expand it to handle both compile and runtime errors over time? It may be that we end up with two separate callbacks, one for compile errors and one for runtime. If possible, though, it's good to try to keep it simple.

preble:
The varargs/formatting bit turned into a bit of a rabbit hole. The approach I was starting to take was to build out a string structure (like `String` but with capacity too), and then functions to allocate, free, and vsnprintf to it (reallocating as needed). All of this in the interests of being able to pass the formatted null-terminated C string to the callback. 

As I type it out here it doesn't sound that bad, but when I was writing it it felt overengineered. Thoughts? Is there a simpler way that I'm not thinking of? I'm trying to steer clear of anything unsafe or too constrained (like limiting output length).

munificent:
> The varargs/formatting bit turned into a bit of a rabbit hole.

Yeah, that's part of why I kicked it down the road so far.

> As I type it out here it doesn't sound that bad, but when I was writing it it felt overengineered.

It would definitely be nice to not have to reimplement much of stdlib streams inside Wren. After some digging around, it turns out there is a simple solution. Woo!

Alas, it's C99-specific, so a little hackery will be required to please MSVC.

[...]

PR #333, opened 2016-03-04
Allow embedders to decide where their run-time errors are logged.

### LINKED ISSUE #332: Embedding API request: `config.errorFn`
Embedders currently can set the `writeFn` for their VMs using a `WrenConfiguration` object.
I would like a similar process for wren runtime error output.

My particular use case is an in-browser wren editor with on-screen debugging, but I expect this feature could be useful for other types of applications.

I have a solution on my end that involves configuring emscripten, but this seems useful enough to merit a pull request.

I'm not really competent in C, but I think this could be achieved by having all the `printf`s [here](https://github.com/munificent/wren/blob/b7ed774da306460120eff9b5e3d9eeec9342c962/src/vm/wren_debug.c) store to some string buffer, passing that to a `config.errorFn`.

I'll start there, and see where that takes me.

ppvk:
Fixes #332 
Please let me know if there's something wrong with my C (especially wren_debug.c  Lines 42-53).

### DIFF
--- a/src/include/wren.h
+++ b/src/include/wren.h
@@ -58,6 +58,9 @@ typedef WrenForeignMethodFn (*WrenBindForeignMethodFn)(WrenVM* vm,
 // Displays a string of text to the user.
 typedef void (*WrenWriteFn)(WrenVM* vm, const char* text);
 
+// Displays a runtime error to the user.
+typedef void (*WrenErrorFn)(WrenVM* vm, const char* text);
+
 typedef struct
 {
   // The callback invoked when the foreign object is created.
@@ -126,6 +129,11 @@ typedef struct
   //
   // If this is `NULL`, Wren discards any printed text.
   WrenWriteFn writeFn;
+
+  // The callback Wren uses to display text when runtime errors occur.
+  //
+  // If this is `NULL`, Wren printf's the error.
+  WrenErrorFn errorFn;
 
   // The number of bytes Wren will allocate before triggering the first garbage
   // collection.
--- a/src/include/wren.h
+++ b/src/include/wren.h
@@ -58,7 +58,7 @@ typedef WrenForeignMethodFn (*WrenBindForeignMethodFn)(WrenVM* vm,
 // Displays a string of text to the user.
 typedef void (*WrenWriteFn)(WrenVM* vm, const char* text);
 
-// Displays a runtime error to the user.
+// Processes a runtime error.
 typedef void (*WrenErrorFn)(WrenVM* vm, const char* text);
 
 typedef struct
@@ -130,9 +130,9 @@ typedef struct
   // If this is `NULL`, Wren discards any printed text.
   WrenWriteFn writeFn;
 
-  // The callback Wren uses to display text when runtime errors occur.
+  // The callback Wren uses to display runtime errors.
   //
-  // If this is `NULL`, Wren printf's the error.
+  // If this is `NULL`, Wren outputs the error to the stderr stream.
   WrenErrorFn errorFn;
 
   // The number of bytes Wren will allocate before triggering the first garbage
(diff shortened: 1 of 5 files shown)


ppvk:
Currently investigating the failed tests.

src/include/wren.h
```diff
@@ -58,6 +58,9 @@ typedef WrenForeignMethodFn (*WrenBindForeignMethodFn)(WrenVM* vm,
 // Displays a string of text to the user.
 typedef void (*WrenWriteFn)(WrenVM* vm, const char* text);
 
+// Processes a runtime error.
+typedef void (*WrenErrorFn)(WrenVM* vm, const char* text);
```
munificent: We can use the same WrenWriteFn for this since it's the same type signature.

src/vm/wren_debug.c
```diff
@@ -2,17 +2,31 @@
 
 #include "wren_debug.h"
 
-void wrenDebugPrintStackTrace(ObjFiber* fiber)
+void wrenDebugPrintStackTrace(WrenVM* vm, ObjFiber* fiber)
 {
   if (IS_STRING(fiber->error))
   {
-    fprintf(stderr, "%s\n", AS_CSTRING(fiber->error));
+    if ( vm->config.errorFn != NULL)
```
munificent: Instead of checking for `NULL` every time you use this, how about just initializing it in `wrenInitConfiguration()` to a function that calls `fprintf(stderr, ...)`?

src/vm/wren_debug.c
```diff
@@ -22,16 +36,27 @@ void wrenDebugPrintStackTrace(ObjFiber* fiber)
...
     // detail of what part of the core module is written in C and what is Wren.
     if (fn->module->name == NULL) continue;
-    
+
     // -1 because IP has advanced past the instruction that it just executed.
     int line = fn->debug->sourceLines.data[frame->ip - fn->code.data - 1];
-    fprintf(stderr, "[%s line %d] in %s\n",
+
+    char buffer[1024];
+    snprintf(buffer, sizeof buffer, "[%s line %d] in %s\n",
```
munificent: Ah, this is the annoying part. Microsoft has really poor support for `snprintf()`. :sob: 

http://stackoverflow.com/questions/2915672/snprintf-and-visual-studio-2010

Libuv has its own emulation of it:

```c
#if defined(_MSC_VER) && _MSC_VER < 1900

#include <stdio.h>
#include <stdarg.h>

/* Emulate snprintf() on MSVC<2015, _snprintf() doesn't zero-terminate the buffer
 * on overflow...
 */
int snprintf(char* buf, size_t len, const char* fmt, ...) {
  int n;
  va_list ap;
  va_start(ap, fmt);

  n = _vscprintf(fmt, ap);
  vsnprintf_s(buf, len, _TRUNCATE, fmt, ap);

  va_end(ap);
  return n;
}

#endif
```

We could probably take that and put it into wren_utils.h/.c. Actually, a cleaner solution might be to add a function in wren_utils that both patches Windows support and handles allocating a buffer of the right size. You do that by calling `snprintf()` twice. The first time it will return the number of bytes it needs to fit the entire string. Then you allocate a buffer of that size and call it again with that buffer.

munificent:
Thanks for hacking on this! It's been on the TODO list forever. It will be great having it finally done.

munificent:
Ah, and now I see there's another pull request for the same thing. I'm not sure if that shows either good timing or bad. The other one handles Windows' broken `snprintf()`, so I'm going to close this one out, but I do appreciate it!

ppvk:
No worries, I've been discussing this with @preble in #265. I'll see about taking your suggestions here and creating a pull request on his pull request.

PR #336, opened 2016-03-08
Adds callback for error messages

preble:
For #265, adds `WrenConfiguration.errorFn`, similar to `writeFn`, which gives embedders access to parse/compile/runtime error messages.

Only caveat is that I haven't been able to test the MSVC code; it was taken from a post that @munificent [linked](https://social.msdn.microsoft.com/Forums/vstudio/en-US/2b339bdf-7ab1-4a08-bf7e-e9293801455b/c99-standard-snprintf-function-?forum=vcgeneral).

Room for improvement: this should be added to the API tests, but haven't had time to figure that area out.

A big hat tip to @ppvk for bringing some activity back to this issue!

### DIFF
--- a/src/vm/wren_vm.c
+++ b/src/vm/wren_vm.c
@@ -1628,17 +1628,30 @@ void wrenPrintError(WrenVM* vm, const char* format, ...)
 {
   va_list args;
   
+#if defined( _MSC_VER )
+  // MSVC doesn't have snprintf():
+  #define WREN_STRING_SIZE(format, args) _vscprintf(format, args)
+  #define WREN_STRING_PRINT(buffer, size, format, args) _vsnprintf_s(buffer, size, _TRUNCATE, format, args)
+#else
+  // Use C99's snprintf():
+  #define WREN_STRING_SIZE(format, args) vsnprintf(NULL, 0, format, args)
+  #define WREN_STRING_PRINT(buffer, size, format, args) vsnprintf(buffer, size, format, args)
+#endif
+
   // Allocate `buffer` to hold the output:
   va_start(args, format);
-  int sizeNeeded = vsnprintf(NULL, 0, format, args);
-  int sizeAvailable = sizeNeeded + 1; // +1 for the null terminator
-  char *buffer = malloc(sizeAvailable);
+  int sizeNeeded = WREN_STRING_SIZE(format, args);
+  int sizeAllocated = sizeNeeded + 1; // +1 for the null terminator
+  char *buffer = malloc(sizeAllocated);
   va_end(args);
-  
+
   // Write the format and args into `buffer`:
   va_start(args, format);
-  vsnprintf(buffer, sizeAvailable, format, args);
+  WREN_STRING_PRINT(buffer, sizeAllocated, format, args);
   va_end(args);
+  
+  #undef WREN_STRING_SIZE
+  #undef WREN_STRING_PRINT
   
   // Output `buffer`:
   if (vm->config.errorFn != NULL)
--- a/src/vm/wren_vm.c
+++ b/src/vm/wren_vm.c
@@ -1642,7 +1642,7 @@ void wrenPrintError(WrenVM* vm, const char* format, ...)
   va_start(args, format);
   int sizeNeeded = WREN_STRING_SIZE(format, args);
   int sizeAllocated = sizeNeeded + 1; // +1 for the null terminator
-  char *buffer = malloc(sizeAllocated);
+  char *buffer = (char *)malloc(sizeAllocated);
   va_end(args);
 
   // Write the format and args into `buffer`:
--- a/src/vm/wren_vm.c
+++ b/src/vm/wren_vm.c
@@ -1,5 +1,6 @@
 #include <stdarg.h>
 #include <string.h>
+#include <stdio.h>
 
 #include "wren.h"
 #include "wren_common.h"
@@ -17,7 +18,6 @@
 
 #if WREN_DEBUG_TRACE_MEMORY || WREN_DEBUG_TRACE_GC
   #include <time.h>
-  #include <stdio.h>
 #endif
 
 // The behavior of realloc() when the size is 0 is implementation defined. It
@@ -42,6 +42,7 @@ void wrenInitConfiguration(WrenConfiguration* config)
   config->bindForeignMethodFn = NULL;
   config->bindForeignClassFn = NULL;
   config->writeFn = NULL;
+  config->errorFn = NULL;
   config->initialHeapSize = 1024 * 1024 * 10;
   config->minHeapSize = 1024 * 1024;
   config->heapGrowthPercent = 50;
@@ -387,7 +388,7 @@ static void runtimeError(WrenVM* vm)
   }
 
   // If we got here, nothing caught the error, so show the stack trace.
-  wrenDebugPrintStackTrace(vm->fiber);
+  wrenDebugPrintStackTrace(vm, vm->fiber);
   vm->fiber = NULL;
 }
 
@@ -1620,4 +1621,34 @@ void wrenGetVariable(WrenVM* vm, const char* module, const char* name,
   ASSERT(variableSlot != -1, "Could not find variable.");
   
   setSlot(vm, slot, moduleObj->variables.data[variableSlot]);
+}
+
+// Prints to WrenVM.errorFn if provided, otherwise prints to stderr.
+void wrenPrintError(WrenVM* vm, const char* format, ...)
+{
+  va_list args;
+  
+  // Allocate `buffer` to hold the output:
+  va_start(args, format);
+  int sizeNeeded = vsnprintf(NULL, 0, format, args);
+  int sizeAvailable = sizeNeeded + 1; // +1 for the null terminator
+  char *buffer = malloc(sizeAvailable);
+  va_end(args);
+  
+  // Write the format and args into `buffer`:
+  va_start(args, format);
+  vsnprintf(buffer, sizeAvailable, format, args);
+  va_end(args);
+  
+  // Output `buffer`:
+  if (vm->config.errorFn != NULL)
+  {
+    vm->config.errorFn(vm, buffer);
+  }
+  else
+  {
+    fprintf(stderr, "%s", buffer);
+  }
+  
+  free(buffer);
 }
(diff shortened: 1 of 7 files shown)


src/vm/wren_vm.c
```diff
@@ -1,5 +1,6 @@
 #include <stdarg.h>
 #include <string.h>
+#include <stdio.h>
```
preble: I'm not sure if this works with MSVC.

src/vm/wren_vm.c
```diff
@@ -1621,3 +1622,46 @@ void wrenGetVariable(WrenVM* vm, const char* module, const char* name,
...
+  // Use C99's snprintf():
+  #define WREN_STRING_SIZE(format, args) vsnprintf(NULL, 0, format, args)
+  #define WREN_STRING_PRINT(buffer, size, format, args) vsnprintf(buffer, size, format, args)
+#endif
+
+  // Allocate `buffer` to hold the output:
+  va_start(args, format);
+  int sizeNeeded = WREN_STRING_SIZE(format, args);
+  int sizeAllocated = sizeNeeded + 1; // +1 for the null terminator
+  char *buffer = (char *)malloc(sizeAllocated);
```
munificent: Lets use the reallocate function provided in the config for this, so it goes through the user-provided allocator.

src/vm/wren_vm.c
```diff
@@ -1621,3 +1622,46 @@ void wrenGetVariable(WrenVM* vm, const char* module, const char* name,
...
+  if (vm->config.errorFn != NULL)
+  {
+    vm->config.errorFn(vm, buffer);
+  }
+  else
+  {
+    fprintf(stderr, "%s", buffer);
+  }
+  
+  free(buffer);
```
munificent: This too.

src/vm/wren_vm.c
```diff
@@ -1621,3 +1622,46 @@ void wrenGetVariable(WrenVM* vm, const char* module, const char* name,
   
   setSlot(vm, slot, moduleObj->variables.data[variableSlot]);
 }
+
+// Prints to WrenVM.errorFn if provided, otherwise prints to stderr.
+void wrenPrintError(WrenVM* vm, const char* format, ...)
+{
+  va_list args;
+  
+#if defined( _MSC_VER )
```
munificent: It does, finally, as of VS 2015, so you could change this to:

```
#if defined(_MSC_VER) && _MSC_VER < 1900
```

Though I suppose it's fine either way.

src/vm/wren_vm.h
```diff
@@ -125,6 +125,10 @@ struct WrenVM
 //   [oldSize] will be zero. It should return NULL.
 void* wrenReallocate(WrenVM* vm, void* memory, size_t oldSize, size_t newSize);
 
+// Prints an error to the configuration's `errorFn`, if non-NULL.
+// Otherwise, prints to stderr.
+void wrenPrintError(WrenVM* vm, const char* format, ...);
```
munificent: How about we move this function to wren_utils.h/.c? wren_vm.c is getting a little cluttered, and this function is fairly separate from the rest of the VM.

src/vm/wren_vm.c
```diff
@@ -1621,3 +1622,46 @@ void wrenGetVariable(WrenVM* vm, const char* module, const char* name,
...
+  #undef WREN_STRING_PRINT
+  
+  // Output `buffer`:
+  if (vm->config.errorFn != NULL)
+  {
+    vm->config.errorFn(vm, buffer);
+  }
+  else
+  {
+    fprintf(stderr, "%s", buffer);
```
munificent: How about we make `wrenInitConfiguration()` just set `errorFn` to point to a little stub function that does this? That way we don't need to check for `NULL` here.

munificent:
This is groovy! I added a few comments, but overall this looks great.

ppvk:
@preble I'm going to try my hand at some of @munificent's comments, both from my closed pull request and this one, and I'll send a pull request your way shortly.

src/vm/wren_vm.c
preble: I think this is a good call.

- [ ] Use vsnprintf if available – check `#if defined(_MSC_VER) && _MSC_VER < 1900`

src/vm/wren_vm.h
preble: :+1: 

- [ ] Move `wrenPrintError` to `wren_utils`

src/vm/wren_vm.c
preble: Could you elaborate a bit on your thinking here? I ask because I'm not the biggest fan of adding a function for this (either to wren_vm.c, or perhaps wren_utils.h/.c); it seems like a natural behavior for `wrenPrintError()`. On the other hand, doing this would make it easier for the user to silence errors by setting it to `NULL`, assuming the check remains, minus the `else { fprintf(stderr, ...) }`.

src/vm/wren_vm.c
preble: Ah, smart.

- [ ] Use wren's built in reallocate function.

preble:
@munificent I think this is just about ready. There is a discrepancy to resolve, however. As currently implemented on this branch if `errorFn` is set to `NULL`, Wren will crash on printing an error. My original [comment](https://github.com/munificent/wren/pull/336/files#diff-fd5ee83f9065d1e5926d1e98ffb147c8R133) from yesterday is still intact, so either that comment should be changed or we should adjust `wrenPrintError` to check for `NULL`.

I was about to just change the comment to match the code, but I noticed that `reallocateFn`'s comment suggests that it gracefully falls back:

> If `NULL`, defaults to a built-in function that uses `realloc` and `free`.

Except this isn't true – if `reallocateFn` is set to `NULL` Wren will crash. My guess is that this comment simply needs to be updated.

I think that `errorFn` is less like `reallocateFn` and more like `writeFn`, which prints nothing if set to `NULL`. So to me checking for `NULL` in `wrenPrintError()` and gracefully failing over to stderr seems sensible.

Thoughts?

munificent:
> Except this isn't true – if reallocateFn is set to NULL Wren will crash. My guess is that this comment simply needs to be updated.

By "defaults to", the doc means "if you call `wrenInitConfiguration()` and then don't set it to anything else." Explicitly setting it to `NULL` would not be the "default", though I suppose the docs could be clearer about how this works.

For reallocateFn, setting an explicit NULL would give you an unusable VM, and it should probably assert that. With writeFn, using an explicit NULL could be useful to mean "do not print".

munificent:
OK, I spent a little more time looking at this and I ran into two snags:

1. `vsnprintf()` isn't supported in standard C++98. It's a C++11 addition. Wren is designed to compile cleanly as either standard C99 or C++98. That ensures it works on most platforms, including older versions of VS on Windows.

2. The code doesn't cleanly route everything through `wrenPrintError()`. In `lexError()` and `error()` in wren_compiler.c, it's still directly calling `fprintf()` and `vfprintf()` for the rest of the error message.

Both of this boil down to the fact that format strings are really handy in C, but aren't well supported once you move outside of talking directly to stdout and stderr. Fortunately, we don't use many formatting features when displaying runtime and compile errors, so I think we can simplify our problem. I've got some ideas to try to make this PR easier to land but I'll need a little time to see if they pan out. Let me hack on things on my end for a while and then I'll ping you.

ppvk:
Sounds good to me.

preble:
:+1: Thanks, @munificent.

munificent:
OK, I made some progress in this commit: https://github.com/munificent/wren/commit/22a95265f2b502cfb3432db3043ec4405f77cc76

Now, there are only a few places that print to stderr:

```c
  fprintf(stderr, "[%s line %d] %s\n", parser->module->name->value, line,
          message);

    fprintf(stderr, "%s\n", AS_CSTRING(fiber->error));

    fprintf(stderr, "[error object]\n");

    fprintf(stderr, "[%s line %d] in %s\n",
            fn->module->name->value, line, fn->debug->name);
```

They are all pretty similar. They have an module name and line number, then a single string of error message. So my thought is we could make `errorFn` take the module name (`const char*`) and line number as separate parameters, and let the host be responsible for formatting those into a single string.

All of the other formatting (well, except for adding "in " to a runtime error message) is done ahead of time by a call to `sprintf()` and `vsprintf()`. For those, we know in advance the largest possible buffer size, so we don't need to use `snprintf()` or dynamically allocate a buffer.

What do you think?

ppvk:
I think that sounds reasonable. Part of the reason for having customisable error output is to have interactive debuggers and whatnot, and this would remove the need to parse an error string for the line and module information.

ppvk:
I'll spend some time tonight reconciling this pull request and the latest master.

ppvk:
Actually, i'm going to have to put that off till around Monday, unless anyone else would like to do it.

ppvk:
@munificent, were you thinking that `printError` would be the 'default' option for a WrenConfiguration error function, or do you think we could throw a conditional in `printError` that would call the user configured error function? This would be like what @preble suggested earlier for `wrenPrintError`.

munificent:
> this would remove the need to parse an error string for the line and module information.

Yes, that's exactly what I had in mind. It directly gives the host more structured information.

> were you thinking that printError would be the 'default' option for a WrenConfiguration error function, or do you think we could throw a conditional in printError that would call the user configured error function?

Right now, at least, Wren has no default function for `writeFn`. If you don't provide one, it's `NULL` and `System.print()` checks for that and doesn't print anything if it is. I was thinking we'd do the same thing for `errorFn`. It's maybe not the nicest out-of-the-box experience for a new embedder, but it has the nice feature of making the VM not depend on `<stdio.h>`, which I'd like.

munificent:
OK, @ppk's version of this is merged in now (https://github.com/munificent/wren/commit/c5855a6fe97cab72f94be40ae7eedc6457c44e72), which I think takes into account everything in this PR too? Thanks for hacking on this!

preble:
@munificent Cool, nice work @ppvk! I like the callback having the line number and module as separate parameters much better, too.

If I am not mistaken that handles compiler errors, which is a great step, but runtime errors are still `fprintf(stderr, ...)`?

munificent:
> If I am not mistaken that handles compiler errors, which is a great step, but runtime errors are still fprintf(stderr, ...)?

Yeah, I noticed that this morning. It should be straightforward to route those through `errorFn` too.

ppvk:
I'll work on this a little later today.

ppvk:
Oops, looks like you beat me to it @munificent.

munificent:
:grin:

PR #337, opened 2016-03-08
Fix windows build

mattn:


util/wren.mk
```diff
@@ -103,9 +103,13 @@ else
 	SHARED_LIB_FLAGS := -Wl,-soname,$@.so
 	SHARED_EXT := so
 
+ifeq (,$(findstring Windows,$(OS)))
```
munificent: This doesn't seem to be doing the right thing on Travis. Would `ifeq ($(OS),mingw32)` work here like some of the other Windows-specific tests use?

mattn: In default, %OS% is set as Windows_NT. AFAIK, I don't know the toolchain or make that set %OS% as mingw32.

munificent: >  I don't know the toolchain or make that set %OS% as mingw32.

`OS` isn't coming from the environment. It's defined by the makefile on line 81. :)

mattn: >OS isn't coming from the environment. It's defined by the makefile on line 81. :)

Did you make sure? See my screenshot. If you re-define in Makefile, it's not right way, I think.

munificent: I'm sure the makefile defines `OS`. I didn't realize at the time that collides with an existing Windows environment variable. The goal was never to use a value coming from the environment. The makefile wants to control the value of `OS`. We could give it another name if it's helpful.

mattn: Okay, I don't have strong objection. :)
What you want? Change the part to below?
```
ifeq ($(OS),mingw32)
```

munificent: Exactly right. :)

mattn: squashed. please take a look.

munificent:
Awesome, thanks! :bird:

mattn:
Thank you too.

PR #344, opened 2016-03-15
io.c fixes to build in Visual Studio

RallyTronics:
Visual Studio has no definitions for the following:

S_IRUSR, S_IWUSR, O_SYNC, S_ISREG and S_ISDIR

I was able to provide substitutes for all but O_SYNC which i set to 0

I also fixed a type in one of the comments

### DIFF
--- a/src/module/io.c
+++ b/src/module/io.c
@@ -10,6 +10,15 @@
 #include <stdio.h>
 #include <fcntl.h>
 
+#if defined(WIN32) || defined(WIN64)
+#include <sys\stat.h>
+#define S_IRUSR _S_IREAD 
+#define S_IWUSR _S_IWRITE
+#define O_SYNC 0
+#define S_ISREG(m) (((m) & S_IFMT) == S_IFREG)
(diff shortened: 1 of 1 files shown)


munificent:
Super groovy, thanks! Hopefully this will get Windows back in shape.

RallyTronics:
I still need to clean up the Visual Studio project, unfortunately it makes assumptions on where libuv is located.  Unless you think otherwise I will change it to assume that libuv is present at the same directory level as wren and issue another pull request.

munificent:
> I still need to clean up the Visual Studio project, unfortunately it makes assumptions on where libuv is located.

Thanks for looking at that! I don't have a Windows machine, so I rely on contributors to keep the VS project in good shape.

> Unless you think otherwise I will change it to assume that libuv is present at the same directory level as wren

For what it's worth, on Mac and Linux, the makefile downloads it into `deps/libuv`. Putting it there on Windows would be consistent and also makes sure it's correctly .gitignored.

Another PR would be rad. :)

PR #347, opened 2016-04-15
Fix small typo in method call documentation

ianmackenzie:


munificent:
Thanks!

ianmackenzie:
Well, GitHub does make this sort of thing pretty easy =)

bjorn:
To be entirely correct, this wasn't a typo but actually an outdated snippet. At some point the position argument was the second one. :-)

PR #354, opened 2016-05-03
Fix typos and re-word explanation of gc heapGrowthPercentage

ElFeesho:
Tried to reduce total number of words.


munificent:
Very nice, thanks!

PR #364, opened 2016-05-19
Add support for retrieving list values from a slot

dradtke:
This PR adds support for retrieving list information from a slot. Rather
than take the whole list, two different methods are provided for
retrieving a) the length of the list, and b) a specific value in the
list.


munificent:
Sorry it took sooo long for me to get to this. It's fantastic! I made some tweaks but otherwise it's in. Thank you for writing it, and tests!

PR #366, opened 2016-05-21
fix range docs to introduction

underscorediscovery:
Is that the link that was intended?


munificent:
Yes, perfect thanks! I must have missed that when I rearranged the top level docs. Thanks!

PR #375, opened 2016-07-02
Allow builds to be verbose

bfontaine:
It’s impossible to disable `@`s in Makefiles to get a verbose build. Adding them conditionally lets users choose if they want a verbose build by using the `VERBOSE` variable.

Normal build:

    make

Verbose build:

    make VERBOSE=1

I stole this idea from [this SO thread](http://stackoverflow.com/q/3861911/735926). Note we could also remove both the variable, the `printf`s and the `@`s and get a verbose build by default.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,73 +2,79 @@
 # compile Wren itself, it invokes util/wren.mk for the various configurations
 # that Wren can be built with.
 
+# Allows one to enable verbose builds with VERBOSE=1
+V := @
+ifeq ($(VERBOSE),1)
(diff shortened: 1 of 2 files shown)


bfontaine:
The Travis CI failure is irrelevant to this PR; it was [already failing before](https://travis-ci.org/munificent/wren/jobs/140560208).

paulsapps:
Created Pr #376 for that

munificent:
Brilliant. I love it. :)

PR #376, opened 2016-07-02
Fix build, return false instead of NULL

paulsapps:


paulsapps:
Note: AppVeyor build is failing because libuv isn't pulled down - since this is configured via the website instead of a yml file I can't fix this.

munificent:
Thank you! Yeah, I still need to poke at around at the AppVeyor build.

PR #379, opened 2016-07-11
Num properties

Nelarius:
Instead of using magic number literals when searching for the smallest or largest number in a list, I figured it might be nice to have some of the underlying `double` properties exposed via the `Num` class.

### DIFF
--- a/src/vm/wren_core.c
+++ b/src/vm/wren_core.c
@@ -1,5 +1,6 @@
 #include <ctype.h>
 #include <errno.h>
+#include <float.h>
 #include <math.h>
 #include <string.h>
 #include <time.h>
@@ -666,6 +667,21 @@ DEF_PRIMITIVE(num_sign)
   }
 }
 
+DEF_PRIMITIVE(num_highest)
+{
+  RETURN_NUM(DBL_MAX);
+}
+
+DEF_PRIMITIVE(num_lowest)
+{
+  RETURN_NUM(-DBL_MAX);
+}
+
+DEF_PRIMITIVE(num_epsilon)
+{
+  RETURN_NUM(DBL_EPSILON);
+}
+
 DEF_PRIMITIVE(num_toString)
 {
   RETURN_VAL(wrenNumToString(vm, AS_NUM(args[0])));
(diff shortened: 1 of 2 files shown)


src/vm/wren_core.c
```diff
@@ -666,6 +667,21 @@ DEF_PRIMITIVE(num_sign)
...
 }
 
+DEF_PRIMITIVE(num_highest)
+{
+  RETURN_NUM(DBL_MAX);
+}
+
+DEF_PRIMITIVE(num_lowest)
+{
+  RETURN_NUM(-DBL_MAX);
```
minirop: Why not DBL_MIN? they have different values in the header.

> DBL_MIN 2.22507385850720138309e-308L
> DBL_MAX 1.79769313486231570815e+308L

(or are they identical, and it's just floating point _error_?)

Nelarius: My idea was for `highest` to correspond to the "largest" positive number, and `lowest` to correspond to the "largest" negative number. Note that `DBL_MIN` corresponds to the smallest positive normalized number (the negative sign is in the exponent).

minirop: Oh right! :confounded:

munificent:
Sorry I got so behind on PRs. :( Too many projects, not enough hours.

I think exposing these constants is a great idea. I'm not sure if `-DBL_MAX` should be exposed separately. (I'm not even sure if that evaluates correctly? Floating point behavior near boundaries confuses me.)

How about we just define getters for `DBL_MAX` and `DBL_MIN`? Users can always negate them if they want. Naming is a little tricky. Java calls them `MAX_VALUE` and `MIN_VALUE`, respectively. Dart calls them `MAX_FINITE` and `MIN_POSITIVE`... which is kind of weird. Ruby just calls them `MAX` and `MIN`.

What do you think about `largest` and `smallest`?

Nelarius:
As far as I know, the floating point number is symmetrical, and since `DBL_MAX` should be exactly representable, then `-DBL_MAX` should be as well. But yeah, I'm not 100% sure on that, and my only backing claim is that it works on my machine™. I guess the `lowest` property is redundant anyway :smiley:

`largest` and `smallest` works for me! I'll get around to revising the pull request a bit later.

munificent:
Groovy, thanks!

PR #380, opened 2016-07-12
missed bytecode name refactoring

minirop:
A missed "rename". Visible only when WREN_DEBUG_TRACE_INSTRUCTIONS is 1.


munificent:
Nice catch, thanks!

PR #381, opened 2016-07-14
Implement String indexOf(needle, startIndex)

underscorediscovery:
The ability to offset the search inside the string is critical for many string operations.
A good example is a string split (or replace) method (implemented in wren, not in core). 
Without the ability to consume the string without a copy, the alternative is to duplicate 
the remaining part, generating both new ranges and new strings. This is bad for garbage, 
and on larger strings like data files can get prohibitively expensive.

In my case, parsing/splitting a dense 92KB file (19582 lines) with several thousand split calls, by line, and then by token per line, dropped from ~750ms down to 14ms running time alone (for newlines, for further parsing it was around 70ms on top which didn't change much as the lines being split weren't as long).

I added some basic tests, and ran the benchmarks before and after (see below).

**Notes**
There are some things that I think could be improved in the implemenation detail later.

For example, it might be nice (and consistent) to have startIndex allow negative values, which would behave like ranges do and use the end as a marker. I also tried to keep the code as similar to what it was at first to make sure the differences are clear.

Feedback welcome!

Things I wasn't sure of specifically:
- The `..indexOf_with_startIndex` naming
- Whether or not it should be a separate function, or handled inside existing `indexOf` with arg checks. If so, how to query for unspecified args? IS_NULL returned false for those...

**benchmark** 

```
api_call - wren                .......... 0.10s 0.0028 101.67% relative to baseline
api_foreign_method - wren      .......... 0.45s 0.0097 100.73% relative to baseline
binary_trees - wren            .......... 0.36s 0.0060 100.34% relative to baseline
binary_trees_gc - wren         .......... 1.14s 0.0085 101.32% relative to baseline
delta_blue - wren              .......... 0.27s 0.0027 100.93% relative to baseline
fib - wren                     .......... 0.32s 0.0047  97.67% relative to baseline
fibers - wren                  .......... 0.06s 0.0022  99.09% relative to baseline
for - wren                     .......... 0.11s 0.0022 100.56% relative to baseline
method_call - wren             .......... 0.22s 0.0043 104.13% relative to baseline
map_numeric - wren             .......... 0.37s 0.0034 100.27% relative to baseline
map_string - wren              .......... 0.15s 0.0016  99.58% relative to baseline
string_equals - wren           .......... 0.30s 0.0074 103.72% relative to baseline
```

### DIFF
--- a/src/vm/wren_core.c
+++ b/src/vm/wren_core.c
@@ -869,7 +869,7 @@ DEF_PRIMITIVE(string_contains)
   ObjString* string = AS_STRING(args[0]);
   ObjString* search = AS_STRING(args[1]);
 
-  RETURN_BOOL(wrenStringFind(string, search) != UINT32_MAX);
+  RETURN_BOOL(wrenStringFind(string, search, 0) != UINT32_MAX);
 }
 
 DEF_PRIMITIVE(string_endsWith)
@@ -893,7 +893,19 @@ DEF_PRIMITIVE(string_indexOf)
   ObjString* string = AS_STRING(args[0]);
   ObjString* search = AS_STRING(args[1]);
 
-  uint32_t index = wrenStringFind(string, search);
+  uint32_t index = wrenStringFind(string, search, 0);
+  RETURN_NUM(index == UINT32_MAX ? -1 : (int)index);
+}
+
+DEF_PRIMITIVE(string_indexOf_with_startIndex)
+{
+  if (!validateString(vm, args[1], "Argument")) return false;
+
+  ObjString* string = AS_STRING(args[0]);
+  ObjString* search = AS_STRING(args[1]);
+  uint32_t startIndex = AS_NUM(args[2]);
+
+  uint32_t index = wrenStringFind(string, search, startIndex);
   RETURN_NUM(index == UINT32_MAX ? -1 : (int)index);
 }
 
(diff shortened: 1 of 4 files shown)


src/vm/wren_core.c
```diff
@@ -893,7 +893,19 @@ DEF_PRIMITIVE(string_indexOf)
   ObjString* string = AS_STRING(args[0]);
   ObjString* search = AS_STRING(args[1]);
 
-  uint32_t index = wrenStringFind(string, search);
+  uint32_t index = wrenStringFind(string, search, 0);
+  RETURN_NUM(index == UINT32_MAX ? -1 : (int)index);
+}
+
+DEF_PRIMITIVE(string_indexOf_with_startIndex)
```
munificent: Maybe just `string_indexOf_from`?

munificent:
This is really really great, especially the rationale. (Although, in this case, I'm an easy sell since the implementation is simple and I think we definitely should of indexOf() with a start.)

> Whether or not it should be a separate function, or handled inside existing indexOf with arg checks. If so, how to query for unspecified args?

You did it the right way. In Wren, methods with different numbers of arguments are distinct methods.

I made a few minor tweaks, like allowing negative indexes like you suggest and beefed up the tests and docs a bit, but it's in now. Thank you! :metal:

underscorediscovery:
Thanks for the notes, and happy to help!

Issue #377, opened 2016-07-05
Add List constructor to create list of given size


munificent:
Sometimes, you need a large list with a given number of elements. Right now, the only way to do that is to repeatedly add elements to it. That's kind of a chore.

We should probably add two constructors:

```
class List {
  construct new(size) // Create list with [size] elements, all null.
  construct filled(size, value) // Create list with [size] elements, all [value].
}
```

sdegutis:
@munificent Meh. How often do you need to prepare to put elements in it way before putting them in it? And doesn't that encourage a kind of race condition where you try accessing it before it's there? It seems like the only real gain from this is performance by avoiding repeated heap allocation, which could be avoided by allocating more than is expected (memory is cheap) in giant chunks. Although I doubt the bytecode compiler could be made to anticipate this by trying to find the array's usage, so it would indiscriminately use more memory in general.

minirop:
@sdegutis it prevents you from writing this kind of ugly code:
```
var x = []
while (dadada) {
    var c = lalala
    while (x.count <= c) {
        x.add(0);
    }
    x[c] += 1
}
```
since it would become:
```
var x = List.new(max_size)
while (dadada) {
    var c = lalala
    x[c] += 1
}
```

mooxen:
@sdegutis The presence of a race condition is interesting, I guess the same argument could be made about the in built slicing method. I dont imagine it does introduce one because you dont stritcly have a reference to the object until its constructed. Maybe I've missed something.

@minirop that code could be improved a little more with a constructor that takes a block argument that genereates its elements but perhaps that encroaches on simply having some sort of list comprehension. Im not sure what the consensus on that is.

sdegutis:
If anything, I'd recommend only having the `filled` variant. It could be optimized with memmap too maybe.

munificent:
> How often do you need to prepare to put elements in it way before putting them in it?

Not too often, which is why we've gotten away with not having this for a while. :)

I ran into it when I was writing a little 2D array class for a game:

```
/// A fixed-size 2D array of values.
class Table {
  construct filled(width, height, fill) {
    _width = width
    _height = height

    // TODO: Better list constructor.
    _cells = []
    for (i in 0...(width * height)) {
      _cells.add(fill)
    }
  }

  [x, y] {
    return _cells[y * _width + x]
  }

  [x, y]=(value) {
    return _cells[y * _width + x] = value
  }

  width { _width }
  height { _height }
}
```

[...]

PR #382, opened 2016-07-17
New list constructors

mooxen:
this is an implementation of the enhancement suggested in  #377.
add 2 constructors to the List class

- One that takes a size, and defaults them to null List.new(_)
- One that takes a size and sets them to a provide value List.filled(_,_)


munificent:
Awesome! I made some changes, but this is in now. Thanks!

In particular, I was convinced by the discussion on #377 to see if we can list with just `List.filled(_)` at first, so I took out the other constructor. If we change our minds, it's easy to get it back. :)

PR #384, opened 2016-08-04
Implemented getting and setting env vars.

cacciatc:
PR to address: https://github.com/munificent/wren/issues/385


cacciatc:
Using clang locally so I didn't run into the gcc fun there with the CI builds...

paulsapps:
```error: implicit declaration of function ‘setenv’ [-Werror=implicit-function-declaration]``` == missing header

**Comment by cacciatc** [comment]
It all makes sense now.

setenv is not included in c99, but getenv is included. Since I was only originally interested in getenv I removed the setenv shenanigans.

**Comment by munificent** [comment]
Sorry for the very very very long delay. Too many projects, not enough time. :(

An API for accessing environment variables sounds great. I don't think we want to hang them off the core VM's `System` class (despite the name that makes it sound great) because that couples the VM more to the underlying OS that I'd like to.

We try to keep the core module itself relatively minimal so that Wren works well as an embedded language.

Instead, I think environment variables fit better into the [`Process` class in the `os` module](https://github.com/munificent/wren/blob/master/src/module/os.wren). How about moving this there?

Also, it's probably a more pleasant API if the user gets back something like a Map for the environment variables instead of a string they have to parse. I would probably shunt the unparsed string over to Wren and then have a Wren implementation of something like `Process.environmentVariables` that splits it out into a map.

What do you think?

**Comment by IngwiePhoenix** [comment]
Don't forget - there is `putenv()` too!

If it was me, I would make a class/module just called `env` and have it's `[key] { ... }` operator overloaded to access or even set environment variables.

```
if(env.DEBUG) {
  // Assuming debug is 0, this would come through. NULL wouldn't, i guess.
}
```

Or a method to just check for a variable to exist, like `env.has(string)`. This would be just my idea, though.

underscorediscovery:
As this PR doesn't fit quite right yet and hasn't been updated, I'm closing it. The CLI will be in it's own repo soon, where we can reassess the environment variable API from there.

PR #391, opened 2016-10-09
Improve Travis Config to Build & Test All Configurations

iwillspeak:
I think this pull request is now complete. Wren now successfully builds in 32 and 64 bit on OS X and Linux on Travis. I've also added in builds which disable NaN tagging which run on the same platforms.

munificent/wren/#167

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,13 @@
 language: c
+os:
+ - osx
+ - linux
 compiler:
   - gcc
   - clang
+env:
+  - WREN_OPTIONS="" CI_ARCHS="ci_32 ci_64"
+  - WREN_OPTIONS="-DWREN_NAN_TAGGING=0" CI_ARCHS="ci_64"
 # Travis VMs are 64-bit but we compile both for 32 and 64 bit. To enable the
 # 32-bit builds to work, we need gcc-multilib.
 addons:
@@ -10,4 +16,4 @@ addons:
     - gcc-multilib
     - g++-multilib
 sudo: false # Enable container-based builds.
-script: make all && make test
+script: make WREN_CFLAGS=${WREN_OPTIONS} ${CI_ARCHS}
--- a/Makefile
+++ b/Makefile
@@ -44,6 +44,28 @@ all: debug release
 	$(V) $(MAKE) -f util/wren.mk MODE=debug ARCH=64
 	$(V) $(MAKE) -f util/wren.mk MODE=debug LANG=cpp ARCH=64
 
+ci: ci_32 ci_64
+
+ci_32:
+	$(V) $(MAKE) -f util/wren.mk MODE=debug   LANG=c   ARCH=32 vm cli test
+	$(V) ./util/test.py --suffix=d-32 $(suite)
+	$(V) $(MAKE) -f util/wren.mk MODE=debug   LANG=cpp ARCH=32 vm cli test
+	$(V) ./util/test.py --suffix=d-cpp-32 $(suite)
+	$(V) $(MAKE) -f util/wren.mk MODE=release LANG=c   ARCH=32 vm cli test
+	$(V) ./util/test.py --suffix=-32 $(suite)
+	$(V) $(MAKE) -f util/wren.mk MODE=release LANG=cpp ARCH=32 vm cli test
+	$(V) ./util/test.py --suffix=-cpp-32 $(suite)
+
+ci_64:
+	$(V) $(MAKE) -f util/wren.mk MODE=debug   LANG=c   ARCH=64 vm cli test
+	$(V) ./util/test.py --suffix=d-64 $(suite)
+	$(V) $(MAKE) -f util/wren.mk MODE=debug   LANG=cpp ARCH=64 vm cli test
+	$(V) ./util/test.py --suffix=d-cpp-64 $(suite)
+	$(V) $(MAKE) -f util/wren.mk MODE=release LANG=c   ARCH=64 vm cli test
+	$(V) ./util/test.py --suffix=-64 $(suite)
+	$(V) $(MAKE) -f util/wren.mk MODE=release LANG=cpp ARCH=64 vm cli test
+	$(V) ./util/test.py --suffix=-cpp-64 $(suite)
+
 # Remove all build outputs and intermediate files. Does not remove downloaded
 # dependencies. Use cleanall for that.
 clean:
@@ -80,4 +102,4 @@ gh-pages: docs
 amalgamation: src/include/wren.h src/vm/*.h src/vm/*.c
 	./util/generate_amalgamation.py > build/wren.c
 
-.PHONY: all amalgamation builtin clean debug docs gh-pages release test vm watchdocs
+.PHONY: all amalgamation builtin clean debug docs gh-pages release test vm watchdocs ci ci_32 ci_64
--- a/example/syntax.wren
+++ b/example/syntax.wren
@@ -129,8 +129,10 @@ class Literals is SyntaxExample {
     3.14159
     1.0
     -12.34
-    0xdeadbeef
-    0x1234567890ABCDEF
+    0x1000000
+// Literals larger than 0x1000000 are only valid in 64 bit builds
+//    0xdeadbeef
+//    0x1234567890ABCDEF
   }
 
   strings {
--- a/src/vm/wren_core.c
+++ b/src/vm/wren_core.c
@@ -618,7 +618,9 @@ DEF_PRIMITIVE(num_bangeq)
 DEF_PRIMITIVE(num_bitwiseNot)
 {
   // Bitwise operators always work on 32-bit unsigned ints.
-  RETURN_NUM(~(uint32_t)AS_NUM(args[0]));
+  uint64_t wideVal = AS_NUM(args[0]);
+  uint32_t val = wideVal & 0xFFFFFFFF;
+  RETURN_NUM(~(uint32_t)val);
 }
 
 DEF_PRIMITIVE(num_dotDot)
--- a/src/vm/wren_value.c
+++ b/src/vm/wren_value.c
@@ -742,9 +742,16 @@ Value wrenNumToString(WrenVM* vm, double value)
   // Corner case: If the value is NaN, different versions of libc produce
   // different outputs (some will format it signed and some won't). To get
   // reliable output, handle that ourselves.
-  if (value != value) return CONST_STRING(vm, "nan");
-  if (value == INFINITY) return CONST_STRING(vm, "infinity");
-  if (value == -INFINITY) return CONST_STRING(vm, "-infinity");
+  switch (fpclassify(value))
+  {
+  case FP_INFINITE:
+    if (signbit(value))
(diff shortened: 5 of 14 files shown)


iwillspeak:
Ok. I'm pretty happy with this now. There's still a couple of issues though:

 * Gcc on Linux seems to fail to compile libuv when `WREN_NAN_TAGGING` is disabled for some reason.
 * A few of the tests fail when run with a 32 bit Wren:
    1. There's a literal that's too big in `example/syntax.wren:133`
    2. The `test/language/deeply_nested_gc.wren` test takes a _very_ long time to complete on 32 bit builds. I'm guessing that this is to do with poor allocation or gc performance but I've not been able to investigate it well enough yet.

iwillspeak:
GC slowdown seems to be in the GC, not in reallocation: 
screen shot 2016-10-09 at 16 59 18

iwillspeak:
Looks like there's a sign overflow or underflow in the logic for choosing the next collection point on 32 bit. The 64 bit build makes it through the run with 2 gcs whereas the 32 bit one seems to collect at each allocation.

64:

```-- gc --
GC 104857645 before, 104831820 after (25825 collected), next at 157247730. Took 0.033s.
-- gc --
GC 118473276 before, 118473484 after (18446744073709551408 collected), next at 177710226. Took 0.046s.
done
```

32:

```
-- gc --
GC 104857785 before, 104835472 after (22313 collected), next at 28404189. Took 0.050s.
-- gc --
GC 104835500 before, 104835728 after (4294967068 collected), next at 28404573. Took 0.041s.
-- gc --
GC 104835984 before, 104835756 after (228 collected), next at 28404615. Took 0.039s.
-- gc --
GC 104835784 before, 104836012 after (4294967068 collected), next at 28404999. Took 0.047s.
-- gc --
GC 104836268 before, 104836040 after (228 collected), next at 28405041. Took 0.040s.
-- gc --
GC 104836068 before, 104836296 after (4294967068 collected), next at 28405425. Took 0.031s.
-- gc --
GC 104836552 before, 104836324 after (228 collected), next at 28405467. Took 0.038s.
-- gc --
GC 104836352 before, 104836580 after (4294967068 collected), next at 28405851. Took 0.040s.
-- gc --
GC 104836836 before, 104836608 after (228 collected), next at 28405893. Took 0.045s.
-- gc --
GC 104836636 before, 104836864 after (4294967068 collected), next at 28406277. Took 0.040s.
...
```

paulsapps:
Seems to be lots of test failures on travis..

iwillspeak:
Yes. There's two main groups of failures. The first is lots of failures seem to be due to precisision issues when NAN_TAGGING is disabled on 32 bit builds. The second is that libuv doesn't compile when NAN_TAGGING is disabeld on linux for 32bit gcc builds.

I'm looking into what can be done about the precision issues, but i'm stumped as to why that one linux configuration fails to build libuv, especially seen as it builds fine when we don't disable WREN_NAN_TAGGING.

minirop:
Why it happens, it's because there is an assert that compare a `size_t` to a `ssize_t` (but why only gcc x86 linux? dunno, maybe the only one with "treat warning as error" on?).
```
struct uv_fs_s { // the type of "read_req"
  ssize_t result; // other fields omitted
};
size_t iovcount;
ASSERT(read_req.result == sizeof(test_buf) * iovcount);
```
and a few line above, there is : `ASSERT((size_t)write_req.result == sizeof(test_buf) * iovcount);`

by the way, the erronous line has been corrected with libuv 1.9.0, a cast as been added.

minirop:
Sorry, when I checked, I have check the wrong line :-1: the master is still missing the cast.

iwillspeak:
Yep, it's https://github.com/libuv/libuv/blob/229b3a4cc150aebd6561e6bd43076eafa7a03756/test/test-fs.c#L2604 that's causing the error. I've tried digging into whats wrong with the build when WREN_NAN_TAGGING is disabled. It looks like method dispatch is broken quite badly. Just trying to run the resulting build as a repl fails:

```bash
$ bin/wrend-32 
\\/"-
 \_/   wren v0.1.0
> [src/vm/wren_vm.h:210] This code should not be reached in wrenGetClassInline()
Abort trap: 6
```

iwillspeak:
I've [opened a PR](https://github.com/libuv/libuv/pull/1089) to fixup the libuv compilation error.

iwillspeak:
Ok, so i've got rid of the 32 builds with NAN_TAGGING disabled and things are looking a lot healthier. Unfortunately i'm not able to reproduce the precision issues that travis seems to see with `cos.wren` and `sin.wren` locally with a VM. I am however able to reproduce the leaked handles from the GCC build and the apparent inadvertent cast to 32 bit integer on clang builds in `bitwise_not.wren`.

To fix the issue with sin and cos failing to return _exactly_ the correct value I'm wondering if we should have an expect statement which takes an epsilon to compare with?

iwillspeak:
Clang on linux seems to be generating an overflow check when converting values to 32 bit for [the not function](https://github.com/munificent/wren/blob/2ebe86c64156b0b69c96af692b64b364e1f5e727/src/vm/wren_core.c#L621):

```c
RETURN_NUM(~(uint32_t)AS_NUM(args[0]));
``

```
// Function prolog
   0x0805e8e0 <+0>:	push   %ebp
   0x0805e8e1 <+1>:	mov    %esp,%ebp
   0x0805e8e3 <+3>:	push   %ebx
   0x0805e8e4 <+4>:	sub    $0x24,%esp

// get the function's address to to IP-relative loads
   0x0805e8e7 <+7>:	call   0x805e8ec <prim_num_bitwiseNot+12>
   0x0805e8ec <+12>:	pop    %eax

// store the address of some constant, at some point relative to the IP, in eacx for later.
   0x0805e8ed <+13>:	add    $0x32714,%eax

// Load the parameters to the funciton into registers
   0x0805e8f3 <+19>:	mov    0xc(%ebp),%ecx
   0x0805e8f6 <+22>:	mov    0x8(%ebp),%edx

// Spill the regsiters
   0x0805e8f9 <+25>:	mov    %edx,-0x8(%ebp)
   0x0805e8fc <+28>:	mov    %ecx,-0xc(%ebp)
   0x0805e8ff <+31>:	mov    %eax,-0x1c(%ebp)


=> 0x0805e902 <+34>:	mov    -0xc(%ebp),%eax
   0x0805e905 <+37>:	mov    (%eax),%ecx
   0x0805e907 <+39>:	mov    0x4(%eax),%eax
   0x0805e90a <+42>:	mov    %esp,%edx
   0x0805e90c <+44>:	mov    %eax,0x4(%edx)
   0x0805e90f <+47>:	mov    %ecx,(%edx)

// call wrenValueToNum and store the return on the stack
   0x0805e911 <+49>:	mov    -0x1c(%ebp),%ebx
   0x0805e914 <+52>:	call   0x8061f30 <wrenValueToNum>
   0x0805e919 <+57>:	fstpl  -0x18(%ebp)

/// move the return value into an XMM register
   0x0805e91c <+60>:	movsd  -0x18(%ebp),%xmm0

// Get the IP-relative stuff back from the stack into EAX
   0x0805e921 <+65>:	mov    -0x1c(%ebp),%eax

// load something from there into xmm1
   0x0805e924 <+68>:	movsd  -0xfef0(%eax),%xmm1

// copy the value retuned from `valueToNum` into xmm2 as well
   0x0805e92c <+76>:	movaps %xmm0,%xmm2

// subtract the mystery contant from it
   0x0805e92f <+79>:	subsd  %xmm1,%xmm2

// store the result in ecx, but with the sign bit flipped
   0x0805e933 <+83>:	cvttsd2si %xmm2,%ecx
   0x0805e937 <+87>:	xor    $0x80000000,%ecx

// compare the mystery constant with the `valueToNum` result
   0x0805e93d <+93>:	ucomisd %xmm1,%xmm0

// Move the double from `valuetoNum` to edx
   0x0805e941 <+97>:	cvttsd2si %xmm0,%edx

// overwrite the integer represetion of the return of `valtoNum` if CF is set
   0x0805e945 <+101>:	cmovb  %edx,%ecx

// not whatever integer representation we have
   0x0805e948 <+104>:	not    %ecx

// Move it back to fp registers, spill things generally get ready to call
   0x0805e94a <+106>:	movd   %ecx,%xmm0
   0x0805e94e <+110>:	movsd  -0xfee0(%eax),%xmm1
   0x0805e956 <+118>:	movaps %xmm1,%xmm2
   0x0805e959 <+121>:	por    %xmm2,%xmm0
   0x0805e95d <+125>:	subsd  %xmm1,%xmm0
   0x0805e961 <+129>:	mov    %esp,%ecx
   0x0805e963 <+131>:	movsd  %xmm0,(%ecx)
   0x0805e967 <+135>:	mov    %eax,%ebx

// convert bcak to a value
   0x0805e969 <+137>:	call   0x8061f10 <wrenNumToValue>
   0x0805e96e <+142>:	mov    -0xc(%ebp),%ecx
   0x0805e971 <+145>:	mov    %edx,0x4(%ecx)

// more stack switcheroo to return the value
   0x0805e974 <+148>:	mov    %eax,(%ecx)
   0x0805e976 <+150>:	mov    $0x1,%al
   0x0805e978 <+152>:	and    $0x1,%al
   0x0805e97a <+154>:	movzbl %al,%eax
   0x0805e97d <+157>:	add    $0x24,%esp
   0x0805e980 <+160>:	pop    %ebx
   0x0805e981 <+161>:	pop    %ebp
   0x0805e982 <+162>:	ret
```

I'm not 100% sure of my interpretation of the assembly output here, or if this is correct behavior or not for the code.

**Comment by iwillspeak** [comment]
I can reproduce the last failure with the following minimal c example:

```c
#include <stdio.h>
#include <stdint.h>

int main(int argc, const char* argv[]) {
	double foo = -4294967293;
	printf("%.14g\n", foo);
	foo = ~(uint32_t)foo;
	printf("%.14g\n", foo);
	return 0;
}
```

on my laptop it prints: 

```
$ ./foo 
-4294967293
4294967292
```

However on `ubutnu/trusty64` it prints the following:

```
$ ./foo 
-4294967293
2147483647
```

If i enable optimisation however the output is the same as above. This appears to be because the compiler has removed the not instruction & casting entirely and just pre-calculated the result however.

iwillspeak:
It looks like the `prim_num_bitwiseNot` is triggering undefined behaviour when we give it numbers that overflow [according to StackOverflow](http://stackoverflow.com/a/25600780/1353098). Both behaviours are acceptable and within the standard.

iwillspeak:
Ok. So i've fixed the bitwise not test by making the overflow behaviour of it a bit more explicit in the code. This seems to stop Clang from emitting runtime checks for overflow and brings back the expected result on all platforms.

This should be good to merge now. Is it OK as is, or shall I sqash it down to fewer commits?

munificent:
This is so *so* awesome. I am a bad person for getting distracted by other projects and taking so long to get to this. It's going to take me a little while to wrap my head around it all, but it looks great so far.

Thank you!

munificent:
This is the bestest thing ever! I tweaked a few things here and there but it's all in now. You rock! 🤘

PR #392, opened 2016-10-09
Fix Bug in GC Reallocation Calculation

iwillspeak:
Updates the way we calculate thenext GC to make sure that we're not
already past the threshold. This was causing endless garbage collections
on 32 bit builds in `test/language/deeply_nested_gc.wren`.

This bug is partly responsible for 32 bit build failures on #391


munificent:
Awesome!

PR #393, opened 2016-10-24
Implement basic sandbox functionality

bytemaster:
When WREN_SANDBOX is defined the VM can be configured to limit the
number of ops that could be executed before the VM returns a runtime
error.

### DIFF
--- a/src/include/wren.h
+++ b/src/include/wren.h
@@ -206,6 +206,12 @@ typedef struct
   //
   // If zero, defaults to 50.
   int heapGrowthPercent;
+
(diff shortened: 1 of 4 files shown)


munificent:
Sorry for the very very long delay.

This patch looks nice, but I have a general worry that it won't be useful to enough Wren users to be worth the complexity and maintenance costs. Every compile flag we add doubles the number of supported configurations and that can get out of hand quickly.

If it's OK I think for now I'll leave this out (though it's really useful as a proof of concept!) and then we'll wait and see if other users come forward and ask for it too.

Issue #362, opened 2016-05-18
Invoking a foreign method using wrenCall() returns an invalid slot array


Nelarius:
I ran into the following issue. After calling a foreign method from C, the slot API can't be queried for the return type of the method -- because `callForeign` has set it to `NULL`.

Here's a minimal example demonstrating the issue.

``` c
void returnsTwo(WrenVM* vm) {
    wrenSetSlotDouble(vm, 0, 2.0);
}

WrenForeignMethodFn bind(WrenVM* vm,
    const char* module,
    const char* className,
    bool isStatic,
    const char* signature
    ) {
    return returnsTwo;
}

int main() {
    WrenConfiguration configuration;
    wrenInitConfiguration(&configuration);
    configuration.bindForeignMethodFn = bind;

    WrenVM* vm = wrenNewVM(&configuration);
    wrenInterpret(vm,
        "class Dummy {\n"
        "  foreign static foreignReturnsTwo()\n"
        "  static returnsTwo() {\n"
        "    return 2.0\n"
        "  }\n"
        "}\n"
    );

    wrenEnsureSlots(vm, 1);

    wrenGetVariable(vm, "main", "Dummy", 0);
    WrenValue* returnsTwo = wrenMakeCallHandle(vm, "returnsTwo()");

    wrenCall(vm, returnsTwo);

    // this is okay -- type and res are as expected
    WrenType type = wrenGetSlotType(vm, 0);
    double res = wrenGetSlotDouble(vm, 0);

    wrenGetVariable(vm, "main", "Dummy", 0);
    WrenValue* foreignReturnsTwo = wrenMakeCallHandle(vm, "foreignReturnsTwo()");

    wrenCall(vm, foreignReturnsTwo);

    // wrenGetSlotType segfaults because vm->apiStack is null
    type = wrenGetSlotType(vm, 0);

    wrenFreeVM(vm);

    return 0;
}
```

Is this as designed? The side effect of this is that if I call a regular Wren method which happens to call a foreign method, the resulting API stack will be NULL, and I can't tell if the method returns something via `wrenGetSlotType`.

munificent:
Ah, this is an interesting wrinkle! I don't think I'd considered this case, but it's totally valid. This sounds like a bug to me.

ahrbe1:
Any news on this bug? Seems that this rather seriously affects the ability to embed wren. Is this a fundamental design problem in wren or merely an implementation issue? Thanks.

bytemaster:
[...]

PR #394, opened 2016-10-26
Fix bug #362 calling Foreign methods

bytemaster:
I assume that invoking callForeign that the apiStack should be restored
to it's original value, rather than reset to NULL.


aggsol:
*bump* Is this integrated any time soon?

munificent:
Sorry I got so behind on stuff! The book sucked away all my free time, and I'm still trying to load balance them.

I think this fix is right, but I would love to have a test for it. If you wouldn't mind writing one, that would be great. If not, I can take a stab at it.

bytemaster:
No time to write test, but can confirm that it is working in my own usage.

munificent:
I'm always worried we'll regress untested code, but for now we should at least have the fix in. Thanks!

PR #396, opened 2016-10-31
Num features: log and pow(_)

underscorediscovery:
Adds two frequently useful functions to Num following the existing conventions of implementation and usage pattern.

Also remembered to update AUTHORS this time (forgot before :)).


munificent:
👍

PR #398, opened 2016-11-21
'make bytecode' generates 'lib/libwren.bc'

ppvk:
I needed to generate a bytecode file for my emscripten-based port for some experiments, and added a bit to the make files to do so.

This was useful to me, perhaps it will be useful for other embedders.

If there is a better way to do this, let me know. 
I just copied the shared lib section and changed the extension.


iwillspeak:
I don't quite understand this, is the `.bc` file the same as a standard shared library but just with a different extension? If so couldn't you achieve the same thing by manipulating the `SHARED_EXT` variable instead?

ppvk:
Turns out that emscripten builds the same file whether the extension is '.a', '.so', or '.bc', meaning this pull request is misinformed and unnecessary. Oops!

I met my project's needs by just renaming the output of `make static` to 'wren.bc.'

PR #405, opened 2017-01-20
added C++ guards in wren.h

foobit:


bjorn:
I think the idea is that when you use Wren in C++, you'd also compile Wren itself as C++, so the `extern "C"` is not necessary. At least the README.md says that Wren compiles as C++98 as well. I'm not exactly sure if there is a particular reason to support this, though.

foobit:
I compile Wren separately as a static library.  In my build pipeline, it is built in parallel with all my other 3rd party libraries.  The external C only libraries I use (zlib, libpng, freetype2, glew, etc), all have C++ guards.  

I do not have to provide any special build settings and allow the compiler to build based on extensions.

Without the guards, C++ will assume the public functions are mangled names and cause link errors.

As a workaround, I'd have to wrap extern "C" every time I include  a Wren header:

`extern "C" { #include "wren.h" }`

aggsol:
Compiling Wren as a C++ static library will create the proper names though.

paulsapps:
The other C libs you mentioned do actually only build as C though, where as wren compiles as CPP.

munificent:
Like others note, you shouldn't need to `extern "C"` Wren's header if you want to use it from C++. You can just compile it as C++ instead and the names will be consistently mangled. Because Wren uses `.c` as a file extension, you may have to tell your build system that you explicitly want to compile it as C++, but that should be fairly straightforward.

We wouldn't want to always use `extern "C"` when compiling wren while `__cplusplus` is defined because that would break users who *do* intend to compile it as C++.

For what it's worth, Lua is set up the same way.

Are you running into problems getting your build system to compile it as C++?

minirop:
You could do (or even wren could  provide it, like lua) "wren.hpp" where its content is:
```
#pragma once
extern "C" {
#include "wren.h"
}
```

munificent:
> You could do (or even wren could provide it, like lua) "wren.hpp" where its content is:

Oh, that's clever. :) I'd be happy to do that for Wren. Would that work for you, @foobit?

bjorn:
> For what it's worth, Lua is set up the same way.

I'm left wondering though, what is the advantage of doing this?

foobit:
The `wren.hpp` is a perfect solution which keeps Wren open to both options.

munificent:
> I'm left wondering though, what is the advantage of doing this?

It's not too hard to get the code to compile as either C or C++ and supporting the latter lets users integrate it into their C++ apps without having to deal with two sets of build flags, `extern "C"`, etc. It just gives you a simpler, more homogenous build.

munificent:
OK, "wren.hpp" is in now: https://github.com/munificent/wren/commit/417ed7aeede528a2f646741a0c4d75beb6d52d64

PR #406, opened 2017-01-20
User data associated with a WrenVM

foobit:
I use Wren embedded with C++.  I needed a way to associate a C++ object with each VM so I can have state access during callbacks.

**New API functions**
`void wrenSetUserData(WrenVM* vm, void* userData);`
`void* wrenGetUserData(WrenVM* vm);`

You can also set the user data during VM initialization via the `WrenConfiguration` struct:
`config.userData = this;`

**Breaking API changes**
Added additional vm parameter to `WrenErrorFn` (The only callback that did not specify the VM).
`typedef void (*WrenErrorFn)(
    WrenErrorType type, const char* module, int line, const char* message, WrenVM* vm);`

### DIFF
--- a/src/cli/vm.c
+++ b/src/cli/vm.c
@@ -155,7 +155,7 @@ static void write(WrenVM* vm, const char* text)
 }
 
 static void reportError(WrenErrorType type,
-                        const char* module, int line, const char* message)
+                        const char* module, int line, const char* message, WrenVM* vm)
 {
   switch (type)
   {
--- a/src/include/wren.h
+++ b/src/include/wren.h
@@ -96,7 +96,7 @@ typedef enum
 // Each of those has the module and line where the method or function is
 // defined and [message] is the name of the method or function.
 typedef void (*WrenErrorFn)(
-    WrenErrorType type, const char* module, int line, const char* message);
+    WrenErrorType type, const char* module, int line, const char* message, WrenVM* vm);
 
 typedef struct
 {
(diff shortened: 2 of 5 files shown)


ayosec:
What do you think on adding a finalizer function associated to the user data?

Something like:

```c
void wrenSetUserData(WrenVM* vm, void* userData, WmUserDataFinalizerFn finalizer)
{
	vm->config.userData = userData; 
	vm->config.userDataFinalizerFn = finalizer; 
}
```

Thus, we can execute a code to release the object in the `userData` when the VM is terminated:

```c
void wrenFreeVM(WrenVM* vm)
{
	ASSERT(vm->methodNames.count > 0, "VM appears to have already been freed.");

	if(vm->config.userDataFinalizerFn != NULL && vm->config.userData != NULL) {
		vm->config.userDataFinalizerFn(vm->config.userData);
	}

	// ...
```

foobit:
I like the finalizer addition.  It provides additional flexibility.

minirop:
wouldn't it be easier to do that association on your side? like instead of manipulation WrenVM's objects directly, you have a custom struct having 2 fields (the VM and your data).

Nelarius:
The tricky part is accessing that struct in the C-style global functions such as `WrenBindForeignMethodFn` or `WrenReallocateFn`. One way of doing it storing data in a global data structure, but that becomes tricky if your application uses many VM instances. In that situation you want to store a pointer to your struct in the VM itself 🙂

munificent:
Awesome, thanks! I don't think it's worth adding a finalizer since the host is responsible for calling `wrenFreeVM()` anyway. The host can always do any finalizing it wants right before it calls that.

PR #412, opened 2017-02-10
New System.puts

SteveVnzk:
I think we need System.print alternative without newline at end

### DIFF
--- a/src/vm/wren_core.wren
+++ b/src/vm/wren_core.wren
@@ -271,6 +271,14 @@ class System {
     for (object in sequence) writeObject_(object)
     writeString_("\n")
   }
+  
+  static puts() {
+    writeString_("")
+  }
+  
+  static puts(obj) {
+    writeObject_(obj)
+  }
 
   static write(obj) {
     writeObject_(obj)
--- a/test/core/system/puts.wren
+++ b/test/core/system/puts.wren
@@ -1,0 +1,8 @@
+class Foo {
+  construct new() {}
+
+  toString { "Foo.toString" }
+}
+
+// Calls toString on argument.
+System.puts(Foo.new()) // expect: Foo.toString


novabyte:
It would break backwards compatibility but it might be more consistent to have "print" and "println" rather than "puts" and "print". Depends on how @munificent feels about stdlib changes.

bjorn:
Having `print` and `puts` like this would be especially confusing for people used to Ruby, where`puts` adds a newlines and `print` does not.

SteveVnzk:
@bjorn you right, have idea for other name?

SteveVnzk:
There is 2 ways
1) Create `println` and `print` (it would break backwards compatibility)
2) Choose right name, maybe from this list: `put`, `p`, `printc`, `format` (can by like `printf` in C)

Noctisdark:
I guess it would be better to implement `System.printf`. Not only it would solve this problem, but it'll be a cool feature.

ayosec:
`puts` is part of the [C standard library](https://en.wikibooks.org/wiki/C_Programming/C_Reference/stdio.h/puts). It works adding a `\n` at the end of the string. If wren adds a function like `puts`, it should do the same.

`println` and `print` (proposed by `SteveVnzk`) can be a better alternative.

If backward compatibility is critical, another option is `print` and `write`.

SteveVnzk:
Ok, how say @Noctisdark, `printf` is solve this problem and cool feature, start work on it.

munificent:
Something along these lines would be nice. I thought about `println()`, but it feels weird to give the longer name to what is typically the more common operation. `puts()` makes sense if you know the C standard lib, but it's probably not helpful if you don't.

Lua calls them `print()` (newline) and `write()` (no newline). What do you think of those?

SteveVnzk:
`print()` and `write()` sounds good

PR #413, opened 2017-02-10
Added Sequence.take and Sequence.skip

bjorn:
These lazy iterator producing methods are useful when working with arbitrary sequences and you need to skip or take some number of elements at the start.


munificent:
Love it!

I made one change: I turned skipping or taking a negative count into a runtime error. My hunch is that is more often than not a user error, so I figured it was better to bring that to their attention. In particular, I don't want them to think that calling `take()` with a negative index takes from the end, which is doesn't.

bjorn:
> I made one change: I turned skipping or taking a negative count into a runtime error. My hunch is that is more often than not a user error, so I figured it was better to bring that to their attention. In particular, I don't want them to think that calling take() with a negative index takes from the end, which is doesn't.

Sounds like a good call, though it could also lead to otherwise unnecessary conditions in the code. Do you know what other frameworks do in this case?

Though it's been a while, my reasoning was probably that these `skip` and `take` methods replace the same functionality implemented with a `for` loop, which also silently does what you'd expect with negative numbers.

munificent:
> Though it's been a while, my reasoning was probably that these skip and take methods replace the same functionality implemented with a for loop, which also silently does what you'd expect with negative numbers.

That's a fair point. I looked a couple of other languages to see what they do:

* [Dart requires a non-negative count](https://api.dartlang.org/stable/1.22.1/dart-collection/IterableMixin/take.html).
* [Ruby throws if the count is negative](https://docs.ruby-lang.org/en/2.0.0/Array.html#method-i-take).
* [C# allows it](https://msdn.microsoft.com/en-us/library/bb503062(v=vs.110).aspx).
* [Java throws](https://docs.oracle.com/javase/8/docs/api/java/util/stream/Stream.html#skip-long-).

So I think it's OK for Wren to make it an error. We could always loosen it later if it ends up being annoying for people.

bjorn:
Alright, thanks for looking up the other languages!

PR #418, opened 2017-03-08
Split and replace in wren.

JonesAndrew:
Rewrote https://github.com/munificent/wren/pull/409 in wren.


bjorn:
Did you do a performance comparison between your C and Wren implementations? Would be interesting I think.

JonesAndrew:
replace.wren
```dart
var start = System.clock

for (i in 0...1000000) {
    "Lorem ipsum dolor sit amet, consectetur adipiscing elit.".replace(" ", "-")
}

System.print(System.clock - start)
```
split.wren
```dart
var start = System.clock

for (i in 0...1000000) {
    "Lorem ipsum dolor sit amet, consectetur adipiscing elit.".split(" ")
}

System.print(System.clock - start)
```

I ran both 5 times for both the c version and the wren version and took the best case.

The c version took 3.323829 seconds for replace and 2.357955 seconds for split.
The wren version took 6.305593 seconds for replace and took 4.425018 seconds for split.

munificent:
Awesome! Nicely done. :)

PR #427, opened 2017-03-22
List.sort

JonesAndrew:
First attempt at https://github.com/munificent/wren/issues/426. Didn't write all the docs and test yet, because wasn't sure if this was the approach people wanted to take.


test/core/list/sort.wren
```diff
@@ -0,0 +1,19 @@
...
+    <(other) { _value < other.value }
+
+    >=(other) { _value >= other.value }
+}
+
+System.print([3, 1, 5, 9, 7].sort()) // expect: [1, 3, 5, 7, 9]
+
+System.print([A.new(3), A.new(1), A.new(5), A.new(9), A.new(7)].sort()) // expect: [1, 3, 5, 7, 9]
+
+System.print([A.new(3), A.new(1), A.new(5), A.new(9), A.new(7)].sort{|n| n.value }) // expect: [1, 3, 5, 7, 9]
```
iwillspeak: Maybe try sorting by something different here to verify that the list isn't being sorted by the `<(other)` and `>(other)` impementations. Could try `-n.value` maybe?

src/vm/wren_core.wren
```diff
@@ -302,6 +302,22 @@ class List is Sequence {
     }
     return result
   }
+
+  sort() { sortStep_(this) }
+
+  sortStep_(l) {
+    if (l.isEmpty) return l
+    return sortStep_(l[1..-1].where {|x| x < l[0] }.toList) + [l[0]] + 
+           sortStep_(l[1..-1].where {|x| x >= l[0] }.toList)
```
iwillspeak: We're going to trigger allocations for each step in the sort this way. Would it be possible to do an in-place sort?

src/vm/wren_core.wren
```diff
@@ -302,6 +302,22 @@ class List is Sequence {
...
+  }
+
+  sort(key) { sortStep_(this, key) }
+
+  sortStep_(l, key) {
+    if (l.isEmpty) return l
+    return sortStep_(l[1..-1].where {|x| key.call(x) < key.call(l[0]) }.toList, key) + [l[0]] + 
+           sortStep_(l[1..-1].where {|x| key.call(x) >= key.call(l[0]) }.toList, key)
```
iwillspeak: This code duplicates the case without the key. Maybe have it delegate to this one with a key implementation that just returns itself (`|x| x`).

bjorn:
I think it is a funny and creative approach to sorting! It's also incredibly inefficient though, since it has quadratic complexity and indeed it is allocation-heavy on top of it. So [bubble sort](https://en.wikipedia.org/wiki/Bubble_sort) would be better at this point, which has the same complexity while avoiding any allocations (by shifting in-place, as @iwillspeak suggested).

However, I think a generic sort method should implement a much more efficient algorithm. One possibility would be an in-place [merge sort](https://en.wikipedia.org/wiki/Merge_sort).

iwillspeak:
👍 for quicksort. If not then maybe an in-place bubble sort for now so we get the API and tests done and the underlying sort could be improved in another pull request.

JonesAndrew:
This is actually an implementation of quicksort, it chooses the first element as the pivot every time, but i agree that it's currently very slow and does a lot of allocations. I think i'll try to write an iterative approach today and see how that compares.

bjorn:
> This is actually an implementation of quicksort

Whoops, shows how much I remember about its implementation... so in terms of complexity it should be fine, but it gets completely ruined by allocating new lists instead of swapping elements in-place.

JonesAndrew:
```
var l = [8400, 6538, 3462, 7401, 2935, 2474, 8912, 250, 6304, 1611, 7686, 2376, 1712, 4315, 2209, 9502, 7785, 2326, 255, 745, 4456, 498, 9487, 6762, 7113, 7126, 3084, 4656, 5770, 9878, 4202, 2475, 7294, 7258, 7194, 8448, 8531, 3546, 7384, 1790, 3961, 8852, 4511, 8305, 9728, 3560, 6415, 3106, 7858, 4220, 6728, 9184, 8373, 8137, 4954, 276, 8147, 6652, 2361, 5528, 7762, 5800, 1334, 4857, 7036, 5945, 9684, 2812, 5318, 7203, 3013, 4439, 6243, 2904, 7338, 5258, 3633, 8719, 9693, 2159, 6403, 6722, 5116, 979, 1047, 3547, 6036, 9019, 9667, 1674, 9727, 3502, 8731, 768, 4431, 6103, 5153, 5803, 9975, 2897]

var start = System.clock
for (i in 0...10000) {
    l.sort()
}
System.print(System.clock - start)
```

Ran the benchmark on the old and new five times and took the best time.

old: 5.871245 secs
new: 2.251747 secs

And just to prove I wasn't completely crazy.

bubble sort: 8.919916 secs

Noctisdark:
This is a variant of quicksort I've never seen before and somehow looks like mergesort.
I think the original quicksort will do better, random pivot will increase performance and if possible, a sorting callback will very useful.
Thanks.

src/vm/wren_core.wren
```diff
@@ -302,6 +302,52 @@ class List is Sequence {
     }
     return result
   }
+
+  sort() { sort_(this.map {|n| [n, n] }.toList) }
+
+  sort(key) { sort_(this.map {|n| [n, key.call(n)] }.toList) }
```
bjorn: Why did you do this? Did your performance measurements show that just the function call for each comparison is likely to have more overhead than the duplication of the entire list into pairs and later duplicating it back to single elements?

JonesAndrew: No I just wasn't smart enough to think of doing it that way :(.

JonesAndrew:
@Noctisdark Ya this new version is a merge sort. When you say a sort callback, what do you have in mind?

iwillspeak:
@JonesAndrew I think sort callback means allowing the caller to specify a custom comparison function. It would be more flexible than just allowing them to specify which key to sort on.

src/vm/wren_core.wren
iwillspeak: I think another thing we want to consider here is do we want `sort` to change the existing object or return a new one. There's arguments for both.

Maybe have two functions: `sort` and `sorted`, where `sort` sorts in place and `sorted` returns a sorted copy.

iwillspeak: Maybe emulate Ruby's `sort`, `sort!` and `sort_by` API: https://ruby-doc.org/core-2.2.0/Array.html#method-i-sort. The current `sort(key)` would map to `sort_by`. Ruby's `sort!` woud map to `sort` and Ruby's `sort` would be `sorted`.

JonesAndrew: Calling the key function on comparison instead of duplicating and then converting back in to single elements took the time down to 1.459349 secs.  Having a sort in place would also be a good idea.

test/core/list/sort.wren
```diff
@@ -0,0 +1,17 @@
+class A {
+    value { _value }
+
+    toString { _value.toString }
+
+    construct new(value) {
+        _value = value
+    }
+
+    <=(other) { _value <= other.value }
```
bjorn: Just a small observation, but the comparison I've seen used consistently in other languages by sorting algorithms is the "less than" operator and not the "less than or equals" operator. Was there any particular reason why you picked `<=`? I think it may be a little confusing.

JonesAndrew: Ya not really sure why i picked that one, I can change it in the next commit.

JonesAndrew:
Went back to an implementation of quick sort. It was performing slightly better at 1.042511 secs. Added in sorting in place with `sorted()`, however did not write tests for it yet. `sorted()` and `sort()` might need to be swapped.

iwillspeak:
Nice! Looks like you're just left with the difficult bit now then: working out what to name them. To my intuition `sort` should modify the object, and `sorted` would be the one which returns a copy.

doc/site/modules/core/list.markdown
```diff
@@ -110,3 +110,40 @@ backwards from the end of the list where `-1` is the last element.
     System.print(list) //> [a, new, c]
 
 It is a runtime error if the index is not an integer or is out of bounds.
+
+### **sort**()
+
+Sorts the list in place.
```
skybrian: drive-by documentation nit: "... using the < operator."

In the core library, it looks like only the Num type has this operator so far.

(Also, since you're using quicksort, maybe say that it's not a stable sort?)

PR #435, opened 2017-03-28
Modernize the Visual Studio solution

Nelarius:
I tinkered a bit with the visual studio build after some build issues with libuv 😈 
* the libuv download & build is done as a pre-build step if you have python2 installed and available in the path
* I added an x64 build target
* updated the solution to the 2017 version. If that's a tad too new, I can do a solution for the 2015 edition as well.


munificent:
This looks great! (Or, at least, I guess it does, since I don't run Windows...) Two minor things:

1. Any chance you get switch it to Python 3? The other Python scripts we use are all Python 3.

2. Is it possible to have it build the object files into something like `build/obj` instead of a new top level directory? I try to minimize the number of top level directories in the repo, and also minimize the variance there between platforms.

Thanks!

Nelarius:
Regarding Python 3, the Visual Studio solution doesn't actually care (or even know) about your python version. It just executes the commands `python util/libuv.py download`, etc. However, I noticed that when running it using Python 3, I was getting errors in `libuv/build/gyp/pylib/__init__.py`, due stuff like `print "..."` (which is Python 2 specific as far as I know?).

Regarding the top-level directories, all of VS's intermediate directories are contained within the `vs2017` directory, including the `deps` directory for libuv. Hence the changes to `.gitignore` 🙂

munificent:
Ah, dig it. Thanks!

PR #436, opened 2017-03-29
Set Object

Noctisdark:
Impelmented a Set type for wren, it uses a splay tree internally to represent the data.
It's *theoretically* blazing fast mostly due to the fact that its time complexity is **Amortized O(log n)**.

I tried to make the interface very simple and intuitive and provided a documentation with some primitive examples and a bit of details about how it work and what you should expect.

I'm actually excited about this pull request, and have some questions.
- Should I remove the comments and reformat the code ?
- What is the tree missing to be fully wren`ish? the interface is currenly very simple and only provides basic operations and lacks a lots of good ones.
- How can I improve it ?

Thanks ! :heart:

### DIFF
--- a/doc/site/modules/core/set.markdown
+++ b/doc/site/modules/core/set.markdown
@@ -1,0 +1,205 @@
+^title Set Class
+
+An orderered self-adjusting binary search tree variant called [Splay tree](https://en.wikipedia.org/wiki/Splay_tree) representing a [Set](https://en.wikipedia.org/wiki/Set_(mathematics)).
+
+Allows fast Dynamic Set operations such as find, delete, min, max, successor and predecessor.
+
+**Time complexity**: **Amortized O(log n)** in the worst case.
+
+**Space complexity**: **O(n)** linear on the number of elements in the set.
+
+## Constructors
+
+### Set.**new()**
+
+Creates a new initially empty set.
+
+##### Complexity
+**O(1)** time and space.
+
+## Static Methods
+
+### **Set.init()**
+
+Called on the initialization of the Set class to define the `SetNode` class.
+
+##### Note
+There is not point of calling this function again.
+
+## Methods
+
+### **insert(item)**
+
+Inserts a new item into the set.
+
+##### Note
+If the item can be equated via the `==` operator the the insertion is discarded. The newly inserted item is moved to the root of the tree
+allowing for constant time re-access.
+
+##### Return value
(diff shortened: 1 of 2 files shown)


JonesAndrew:
My understanding is that unique collections are usually implemented with dictionaries. Considering there is already a map type in wren it might be simpler to just take that route. I might take the time complexity stuff out of the docs, but other than that they seem very clean.

Noctisdark:
@JonesAndrew thank you very much for the answer.
It's actually the other way around.
Maps are usuallly implemented as either hash-tables or self-balancing binary tree, C++ STL uses a red black tree for example. In wren it is map implemented as a `Sequence` which will not scale well, the difference between `O(n)` and `O(log n)` is very huge. for that even, maps should be reimplemented as hash table or trees.
The point of implementing `Set` for wren is to get better performance for dealing with unique comparable objects, if the time complexity is in the way, i'll glady remove It.

bjorn:
> The point of implementing Set for wren is to get better performance for dealing with unique comparable objects, if the time complexity is in the way, i'll glady remove It.

Could be interesting to do a performance comparison between your Set implementation and a Set implemented as a thin wrapper around a Wren map.

Noctisdark:
@bjorn great idea !
How should I hash integers, floats or arbitrary wren objects ?
The current implementation only uses string keys, right ?

bjorn:
> How should I hash integers, floats or arbitrary wren objects ?
> The current implementation only uses string keys, right ?

I believe the Wren map takes any Wren value as key.

bjorn:
See https://github.com/munificent/wren/blob/master/src/vm/wren_value.c#L416-L442

Noctisdark:
@bjorn, deal!
 I'll run benchmarks soon on wren browser to decided the algorithmic performance.

Noctisdark:
@bjorn 
I was just doing a speed test for `Set.insert` vs `Map[]=` and `Set.contains` vs `Map[]`.

Map and Set performed nearly as well for small inputs, less than 1000.
For larger inputs, map a lot faster larger inputs, 12 time faster for 1000000 insertion.

I'm guessing the penalty is coming from the fact that `Map` is mostly implemented in C which is a bit faster than Wren and that becomes significant when the number of operations blows up.

Thanks for the insights and sorry for taking your time.

Noctisdark:
@bjorn, will surely do ! Thanks.

PR #440, opened 2017-03-30
Fix typos in documentation

nathanhourt:
The documentation for the bitwise or operator described the behavior of the bitwise and operator.


nathanhourt:
The prose in wren/doc/site/embedding/calling-c-from-wren.markdown at line 115 is a bit mangled, but I'm not sure what it's supposed to say so I can't submit a fix for that.

munificent:
Oops, thanks!

Issue #430, opened 2017-03-22
Import statements can leak memory


munificent:
From u/sponkemonke.

Not sure where the leak is.

minirop:
after checking with it myself, the loaded code of a module doesn't seems to be freed.

In the CLI, there is the `readModule()` function that return the malloc'd source code inside `readFile()` then pass it to the VM, but this *string* is never deleted.

I added `free((void*)source)` in [https://github.com/munificent/wren/blob/master/src/vm/wren_vm.c#L490](loadModule) line 491 (between `wrenCompile` and the `if`), now valgrind is happy. I didn't do a PR since I'm not entirely sure (but still pretty sure), and maybe you'll see something I didn't (if it's all OK for you, I'll do one)

The cast is mandatory else GCC warns that, since `free` ask for `void*` it discard the *const* modifier.

JonesAndrew:
The problem with freeing memory there is that if you pass a string literal to `wrenInterpret` than you can't free it, or if reading a module returns a string literal like Meta and Random. Not sure what the best way to fix this because it's not really possible for the read function to free the data. Maybe add a flag that the read function can set if the source needs to be freed or not.

folkertk1:
I may have solved this. It depends whether the munificent agrees with the way I handled it.
https://github.com/folkertk1/wren/blob/iss430/src/vm/wren_vm.c#L505 is the link to the main part of the change. I added another function called scalloc() which is exposed to the end user, however, I had to
add another variable to the vm struct to make it work

jvidziunas:
I'm going to give a dissenting opinion here and state that the real design issue is that the import function has the 'wrong signature'/goes about the problem backwards. Instead of adding difficult to check state, why not promote the core pieces of loadModule() into the public API and have the import function return the fiber/some sort of result obtained from calling loadModule()?

PR #446, opened 2017-04-08
Iss430

folkertk1:
This pull request tries to solve the issue #430. It adds a function called scalloc and an integer called isSourceAllocated to the vm struct. scalloc uses the reallocate function described by either the user or the standard reallocate function. It also sets isSourceAllocated to 1. During initialization isSourceAllocated is set to zero. After freeing the memory isSoureAllocated is set to zero again.

### DIFF
--- a/src/vm/wren_vm.c
+++ b/src/vm/wren_vm.c
@@ -67,6 +67,7 @@ WrenVM* wrenNewVM(WrenConfiguration* config)
     wrenInitConfiguration(&vm->config);
   }
 
+  vm->isSourceAllocated = 0;
   // TODO: Should we allocate and free this during a GC?
   vm->grayCount = 0;
   // TODO: Tune this.
@@ -506,6 +507,7 @@ static ObjFiber* loadModule(WrenVM* vm, Value name, const char* source)
      void * check = vm->config.reallocateFn((void *) source, 0);
      if(check)
       return NULL;
+    vm->isSourceAllocated = 0;
   }
   if (fn == NULL)
   {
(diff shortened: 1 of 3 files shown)


src/vm/wren_vm.c
```diff
@@ -488,18 +502,25 @@ static ObjFiber* loadModule(WrenVM* vm, Value name, const char* source)
   }
 
   ObjFn* fn = wrenCompile(vm, module, source, false, true);
+  if(vm->isSourceAllocated)
+  {
+     void * check = vm->config.reallocateFn((void *) source, 0);
+     if(check)
+      return NULL;
+    vm->isSourceAllocated = 0;
```
paulsapps: You've just called reallocateFn.. so why is isSourceAllocated 0?

paulsapps:
you seen to have added lots of white space changes, quite a noisy diff

src/vm/wren_vm.c
folkertk1: If you look at https://github.com/munificent/wren/blob/02bcefcbe44cdb73efeebc9b2a9702c41a9674a9/src/include/wren.h#L50

The comments describe the way to free an object using wren's allocation functions is to pass the allocated data to the reallocateFn function pointer and give it a size 0. In the defaultReallocate function when the size is 0 it is freed. The reason isSourceAlocated is set to zero afterwards is because it is freed

folkertk1:
Yeah, sorry about that. I removed two misakes from my earlier commit. If the diff is too noisy it can make a new pull request without so much noise

minirop:
no need for another PR, you can squash the commits and force push.

folkertk1:
It seems the issues of the newline added and removed in the commits is also present on munificent's master branch. If you know anyway to avoid these annoying newlines can you please tell me?

munificent:
Thanks for taking a stab at this! I did a little digging and I think I figured out a simpler solution instead. See: 889cae5ff1d870dd6c8443d81fabcb94cbde7e87

Issue #437, opened 2017-03-29
gyp: Daily bandwidth rate limit exceeded


minirop:
the CI can fail because of the daily limit of gyp's repository. [Like it one](https://travis-ci.org/munificent/wren/jobs/216072524).

Is there a mirror without the limit or is copying the repo (or the necessary files of the repo) a thing to do? (so the "[good-known commit](https://github.com/munificent/wren/blob/master/util/libuv.py#L75)" would be "known")

paulsapps:
Why not just add it as a git sub repo? Then its optional for people to clone down with --recursive

minirop:
@paulsapps it would still have the issue since it's Travis' ip that gets "banned" from checking the repo (since the makefile just git clone the repo).

bjorn:
> Oh, dang. It might make sense to vendor libuv into the main Wren repo.

Why put whole of libuv in it if only gyp fails to clone? Maybe just gyp would be small enough to include in the repo, if there really isn't some mirror or package that could be installed instead?

munificent:
> Why put whole of libuv in it if only gyp fails to clone? 

Oh, good point. We could vendor gyp in. I wonder if libuv has or is planning to move off gyp anyway? My understanding is that it's basically dead. I don't think Google uses it for anything anymore. (Chrome now uses GN and Ninja.)

kirbyfan64:
@munificent Doesn't look like it: https://github.com/libuv/libuv/pull/952

munificent:
OK, I took a stab at solving this: https://github.com/munificent/wren/pull/447

It works on my Mac and on Travis, but I'd like to see what people think and hopefully get some Windows and Linux testing before I merge it in. So... thoughts?

munificent:
Alright, libuv and gyp are vendored into the repo now. :)

PR #447, opened 2017-04-09
Vendor GYP and libuv into the repo.

munificent:
Instead of dynamically downloading these as needed during a build, this
checks in those two dependencies directly into the Wren repo. That's a
little lame because users of Wren who aren't building the CLI don't
actually need them, but they aren't too big, so it's not a huge deal.

It makes builds (particularly on Travis) more reliable, because they
don't have to pull down additional content over the network.


mhermier:
Built for me correctly on linux-x86_64.

While at it, speaking of libuv, I recently find and issue in wren code where libuv was using already free'd data. See the first patch of my change set:
https://github.com/mhermier/wren/commit/b5f74dcf186b291184664248260da7a5934c79c3 (libuv fixup.)
It seems libuv is taking care at freeing stdinStream, and will do it later, when hitting the event loop. So the extra free makes valgrind cries.

Nelarius:
The Visual Studio solution will have to be updated, since it downloaded libuv into the project folder. I can submit a pr for that a bit later.

Nelarius:
Argh, it seems like `vcbuild.bat` wants to build everything, tests included, even though we don't run them. There's only one `msbuild` incantation in there, and it's for the entire solution. 😕

Maybe there's a work-around, it would be a bit of a bummer to have to include **all** of libuv.

paulsapps:
Hmm.. wouldn't a sub module make more sense instead of adding it to the wren repo itself? It will be counted in the LOC etc.

kirbyfan64:
@paulsapps Because then it's still being cloned, therefore it still counts towards the bandwith limit (#437).

bjorn:
> @paulsapps Because then it's still being cloned, therefore it still counts towards the bandwith limit (#437).

But only Gyp suffered from a bandwidth limited repository, not libuv.

I would suggest to only include Gyp in the repository.

paulsapps:
A submodule on github will not have that problem @kirbyfan64  if it did then cloning the wren repo itself will too. By submodule I mean git submodule on github - a mirror.

Issue #448, opened 2017-04-09
System Call [Proposal & PR]


FreddieRidell:
Add support for creating new processes with shell commands.

I propose to add this as a `call` method to the `Process` object in `os` with the following signature:

+ `call(command)`: run the command, and return the exit code
+ `call(command, async)`: run the command, and return the `stdout` of the command
  + if `async` is false, blocks until the command is done and returns its output as one string
  + if `async` is true, does some cool fibrous stuff, I'd like to discuss exactly what, as I'm not sure what the best course of action is.

My initial thought for the async method would be to mirror the API for `Stdin.readline`, though I'd like a little guidance on how best to achieve this.

I've already got a PR of the first method ready to go, tested on linux, I'm afraid I have no access to a windows PC.

CodogoFreddie:
On reflection, it seems that it would make most sense to superclass `Stdin` from `io` into a `Stream` class, that is inherited by `Stdin`, `Subprocess.Stdin`, and possibly `File`.  I'm really happy to re-write my code and try to do this properly over the next few months, but would really appreciate some outside input before I start coding :)

PR #462, opened 2017-06-20
Subprocess spawning

CodogoFreddie:
implements #448 

FEEDBACK REQUESTED
I tried to produce what I considered the most logical API, basing it somewhat on the `File` API.
There has been some talk of a stream super-class, and now might be a good time to think about working on and standardizing this.
This PR is incomplete, but is at a minimum viable stage to start discussion.

I believe this is an important feature for wren, as it would allow us to start prototyping more features, using other programs to mock native functionality. We could focus on producing a strong wren API for things such as HTTP, crypto, etc. by relying on 3rd party programs, before having to implement them in c code.

### DIFF
--- a/util/wren.mk
+++ b/util/wren.mk
@@ -43,7 +43,7 @@ ifeq ($(VERBOSE),1)
 endif
 
 C_OPTIONS := $(WREN_CFLAGS)
-C_WARNINGS := 
+C_WARNINGS := -Wall -Wextra -Werror -Wno-unused-parameter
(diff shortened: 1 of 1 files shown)


CodogoFreddie:
I stopped working on this because I realised we probably wanted to nail down a relatively good implementation for streams and/or asynchronous code before we started this; as those are such important features for good sub-process spawning.

I currently have a RFC for async that I'm working on in my spare time, and I'm still very open to picking this back up once some more supporting infrastructure has been defined :)

PR #464, opened 2017-07-21
Add "as" keyword

ruby0x1:
Some tinkering!

- I prefer `as`, it's clearer than `in`, although you could keep `in` and not add a token.
- I like the syntax of `import "io" for File as F, Process as P` it reads nicely
- It's optional, so `import "io" for File as F, Process` is valid

Thoughts welcome!

### DIFF
--- a/src/vm/wren_compiler.c
+++ b/src/vm/wren_compiler.c
@@ -95,6 +95,7 @@ typedef enum
   TOKEN_FOREIGN,
   TOKEN_IF,
   TOKEN_IMPORT,
+  TOKEN_AS,
   TOKEN_IN,
   TOKEN_IS,
   TOKEN_NULL,
@@ -571,6 +572,7 @@ static Keyword keywords[] =
   {"foreign",   7, TOKEN_FOREIGN},
   {"if",        2, TOKEN_IF},
   {"import",    6, TOKEN_IMPORT},
+  {"as",        2, TOKEN_AS},
   {"in",        2, TOKEN_IN},
   {"is",        2, TOKEN_IS},
   {"null",      4, TOKEN_NULL},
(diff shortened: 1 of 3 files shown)


munificent:
I love tinkering!

I did something similar with "as" way back when I was working on [some other language](http://magpie-lang.org/modules.html). I agree it's clearer than `in`. I did always feel weird that it reads in the opposite direction as assignment. Assignment is `dest = source` and `as` is `source as dest`.

My general tendency towards minimalism makes me hesitate to pull this in, but I'm curious to see what others think. It does look nice to my eye, the implementation is simple and clean (always a good sign), and other languages have similar functionality.

underscorediscovery:
Yea that makes sense. To add more reasoning, I have definitely found as projects get larger that the overlap of variable names can become a frustrating problem, not just because of the encouragement to decorate variable names unnecessarily (Button vs UIButton), but specifically because it creates an impossible situation that is dumped on the users lap, with no tools to tackle it.

I find for example, not being able to name a class anything that matches the name of an import, the only option is forcing my code into less intuitive naming patterns.

Another example is two unrelated modules, maybe there's a `Button` in `moduleA` and a `Button` in `moduleB` - both from third parties. My code **cannot** use both in the same place, it is impossible. The user is forced to rearrange their code, edit the third party module code, or just not use them. None of that sits well with me.

While I highly appreciate the minimalism, the intent is to offer a tool to mitigate impossible or frustrating situations that the user didn't create, but the language allows :)

IngwiePhoenix:
I would love having at least the `import` option there - being able to "rename" a variable during imports may come in really handy! Also, I agree that `in` and `as` read totally the opposite way... Well, in the `for` loop, it does make sense for `in` to be used.

```
for(line in text) ...
```

Whereas this would feel odd again:

```
for(line as text) ...
```

So I totally support @underscorediscovery 's 2nd idea, but am unsure about the first. Though, third is absolutely important.

PR #471, opened 2017-09-18
Add a little example demonstrating Functions

oz123:
* We show how to use recursion and for loops

### DIFF
--- a/example/fibster.wren
+++ b/example/fibster.wren
@@ -1,0 +1,16 @@
+// calculate fibonacy numbers
+
+var fibs = [0, 1, 2, 3, 4, 5]
+
+var Fib = Fn.new {|num|
+    if (num <= 1) {
+        return 1
+    } else {
+        return Fib.call(num - 2) + Fib.call(num - 1)
+    }
+}
+
+
+for (number in fibs) {
+    System.print(Fib.call(number))
+}


bjorn:
There is already [a Fib implementation](https://github.com/munificent/wren/blob/master/test/benchmark/fib.wren) in the benchmarks though, and it was a little faster when not implemented as a function but as a static method (d5377faf113f4417abe8be40e2a061b16ccebca7). :-)

oz123:
Great example!
It demonstrates much more than recursion:
 - Class definition
 - lookup of methods in a class
 - recursion

For teaching the language though, I think it's a bit overwhelming.
Right now, the public attracted to wren might be more experienced programmers, which won't be confused by it. However, I wanted a classical demo with explicit usage of `Fn.call` - without classes. I think it can live side by side with the bench mark examples ...

munificent:
Having more examples is good. I agree that letting users start simple is important. One of the things that's been an open design issue with Wren in my mind is that it doesn't actually have good support for straight procedural code. You can use function objects stored in variables like you do here, but it goes against the grain of the language, and you do still have to understand OOP since you are constructing instances of class "Fn" and invoking a `.call()` method on them.

I've tried a bunch of different ways to reconcile that and support "top level functions" in Wren but so far nothing has worked out. I haven't entirely given up yet. Until then, though I'm not sure if I'd consider this example idiomatic and the kind of code we want to encourage people to write. At least right now, Wren really is an object-oriented language and it wants you to put behavior inside classes like you would in Java or C#. That feels a little strange in a scripting language where the top level of a file is imperative code, but... *shrug*.

If it's OK, I think I'll leave this example out, though it definitely is worth thinking about how to teach users the language incrementally and how to deal with procedural style code.

PR #474, opened 2017-10-19
missing language specifier and variable declaration

minirop:
the last C code in the embed doc is missing syntax highlighting and the declaration of the variable used by `closeFile`.


munificent:
That was fast! 😁

Issue #454, opened 2017-05-02
define a wrenStringToNum() for foreign methods?


iweindesmedt:
Maybe it's a good idea to put the logic from `DEF_PRIMITIVE(num_fromString)` into a new `wren...` function, so this functionality is available (like `wrenNumToString`) to foreign methods

mhermier:
Can you show a use case? I don't see a real one where it is not better to *force* the user to use Num::fromString() from wren language and deal with exceptions/errors from there.

iweindesmedt:
A few reasons:
- There is `wrenNumToString`, so it seems intuitive to have `wrenStringToNum` too.
- `Num::toString()` uses `wrenNumToString` anyway, so why not have `Num::fromString()` use `wrenStringToNum` (or to use the same naming, maybe `wrenNumFromString`, but thats another discussion)
- Converting strings to numbers is a very common use case. For example getting input from stdin as a number, parsing a serial number like 'ABC-123' and split it in a string and number, and so on...

This way, this functionality would be 'globally' available through a `wren...` function instead of having to use `strtod` or any other string-to-num function.

mhermier:
While I agree with all these arguments, they only are profs of form and not presence.
There are quite a number of use-case of strtod in the code base, in particular in the compiler, that would tend to validate its existence.
In the other hand there is only one caller of wrenNumToString so maybe it should be inlined.

munificent:
> I don't see a real one where it is not better to force the user to use Num::fromString() from wren language and deal with exceptions/errors from there.

I think I generally agree with @mhermier here. In most cases, it should be cleaner to do the conversion in the Wren side of the code. If it's a pain, though, I could see adding support for this to the API.

I should point out that `wrenNumToString()` is *not* currently part of the public API for Wren. If it was, it would be declared in "wren.h". Its name starts with "wren" because it's used by other source files in the VM. That means it needs to be exported and visible to the linker and I want to ensure it doesn't collide with other functions in the host application. The only functions in Wren that aren't prefixed are static functions that are local to a single compilation unit.

PR #476, opened 2017-10-27
Number conversions

mhermier:
My attempt at fixing issue #454. I try to inline functions with one call site, and try rationalize string to Num conversion. By doing so there is only one utility function used to convert numbers to strings, that is now shared in Num::fromString and the compiler. Note that it accepts scientific and hexadecimal values.

In extra, but commented, there are 2 extra *features*:
- Num::fromString would/should thow an error on invalid conversion, instead of returning null.
- wrenNewNumFromStringLength can be hardened to detect precision loss in hex format.

### DIFF
--- a/src/vm/wren_compiler.c
+++ b/src/vm/wren_compiler.c
@@ -706,25 +706,26 @@ static int readHexDigit(Parser* parser)
 }
 
 // Parses the numeric value of the current token.
-static void makeNumber(Parser* parser, bool isHex)
+static void makeNumber(Parser* parser)
 {
-  errno = 0;
+  const size_t length = parser->currentChar - parser->tokenStart;
 
-  if (isHex)
+  parser->current.value = NUM_VAL(0);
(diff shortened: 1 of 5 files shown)


munificent:
This looks pretty nice! A couple of questions:

* Is there a reason you moved `wrenNewStringFromDouble()` into the header and made it inline? It seems a little large to benefit from that.

* This means that `Num.fromString()` now accepts hex values too, right?

mhermier:
Le 1 nov. 2017 00:22, "Bob Nystrom" <notifications@github.com> a écrit :

This looks pretty nice! A couple of questions:

   -

   Is there a reason you moved wrenNewStringFromDouble() into the header
   and made it inline? It seems a little large to benefit from that.

There is only one caller of it, no need to have it non static.


   -

   This means that Num.fromString() now accepts hex values too, right?

Yes, there is also a commented check to ensure hex values fits in double as
unsigned values, contrary to master.

PR #479, opened 2017-10-31
Proposal for smarter imports.

munificent:
I'd like to make imports work better for reusing code. The change is big enough that I thought it made sense to write it up like an actual "RFC" proposal as a separate doc. This PR tracks the proposal text itself. Feel free to comment on it here or on the mailing list.

[Rendered proposal](https://github.com/munificent/wren/blob/9f79dfcdac466c1198cecedd12fc857d6e62bc36/doc/rfc/0001-smarter-imports.md)

### DIFF
--- a/doc/rfc/0001-smarter-imports.md
+++ b/doc/rfc/0001-smarter-imports.md
@@ -1,0 +1,475 @@
+# Smarter Imports
+
+Here's a proposal for improving how imported modules are identified and found
+to hopefully help us start growing an ecosystem of reusable Wren code. Please
+do [let me know][list] what you think!
+
+[list]: https://groups.google.com/forum/#!forum/wren-lang
+
+## Motivation
+
+As [others][210] [have][325] [noted][346], the way imports work in Wren,
+particularly how the CLI resolves them, makes it much too hard to reuse code.
+This proposal aims to improve that. It doesn't intend to fix *everything* about
+imports and the module system, but should leave the door open for later
+improvements.
+
+[210]: https://github.com/munificent/wren/issues/210
+[325]: https://github.com/munificent/wren/issues/325
+[346]: https://github.com/munificent/wren/issues/346
+
+### Relative imports
+
+Today, it's hard to reuse your own code unless you literally dump everything in
+a single directory. Say you have:
+
+```text
+script_a.wren
+useful_stuff/
+  script_b.wren
+  thing_1.wren
+  thing_2.wren
+```
+
+`script_a.wren` and `script_b.wren` are both scripts you can run directly from
+the CLI. They would both like to use `thing_1.wren`, which in turn imports
+`thing_2.wren`. What does `thing_1.wren` look like? If you do:
+
+```scala
+// thing_1.wren
+import "thing_2"
+```
+
+Then it works fine if you run `script_b.wren` from the `useful_stuff/`
+directory. But if you try to run `script_a.wren` from the top level directory,
+then it looks for `thing_2.wren` *there* and fails to find it. If you change the
+import to:
+
+```scala
+// thing_1.wren
+import "useful_stuff/thing_2"
+```
+
+Then `script_a.wren` works, but now `script_b.wren` is broken. The problem is
+that all imports are treated as relative to the directory containing the
+*initial script* you run. That means you can't reuse modules from scripts that
+live in different directories.
+
+In this example, if feels like imports should be treated as relative to the
+file that contains the import statement. Often you want to specify, "Here is
+*where* this other module is, relative to where *I* am."
+
+### Logical imports
+
+If we make imports relative, is that enough? Should *all* imports be relative? I
+don't think so. First of all, some modules are not even on the file system.
+There is no relative path that will take you to "random" — it's built into the
+VM itself. Likewise, "io" is baked into the CLI.
+
+Today, when you write:
+
+```scala
+import "io"
+```
+
+You aren't saying *where* that module should be found, you're saying *what*
+module you want. Assuming we get a package manager at some point, these kinds of
+"logical" imports will be common. So I want these too.
+
+If you look at other langauges' package managers, you'll find many times a
+single package offers a number of separate libraries you can use. So I also
+want to support logical imports that contain a path too — the import would say
+both *what* package to look in and *where* in that package to look.
+
+### Only logical imports?
+
+Given some kind of package-y import syntax, could we get rid of relative imports
+and use those for everything? You'd treat your own program like it was itself
+some kind of package and anything you wanted to import in it you'd import
+relative to your app's root directory.
+
+The problem is that the "root directory" for your program's "package" isn't
+well-defined. We could say it's always the same directory as the script you're
+running, but that's probably too limiting. You may want to run scripts that live
+in subdirectories.
+
+We could walk up the parent directories looking for some kind of "manifest" file
+that declares "the root of the package is here", but that seems like a lot of
+hassle if you just want to create a couple of text files and start getting some
+code running. So, for your own programs, I think it's nice to still support
+"pure" relative imports.
+
+### Ambiguity?
+
+OK, so we want both relative imports and logical imports. Can we use the same
+syntax for both? We could allow, say:
+
+```scala
+import "a/b"
+```
+
+And the semantics would be:
+
+1.  Look for a module "a/b.wren" relative to the file containing the import. If
+    found, use it.
+
+2.  Otherwise, look inside some "package" directory for a package named "a" and
+    a module named "b.wren" inside it. If found use that.
+
+3.  Otherwise, look for a built in module named "a".
+
+This is pretty much how things work now, but I don't think it's a good idea.
+Relative imports will tend to be short — often single words like "utils".
+Assuming we get a healthy package ecosystem at some point, the chances of one of
+those colliding with a logical import name are high.
+
+Also, when reading code, I think it's important to be able to easily tell "this
+import is from my own program" without having to know the names of all of the
+files and directories in the program.
+
+## Proposal
+
+OK, so here's my goals:
+
+1.  A way to import a module relative to the one containing the import.
+2.  A way to import a module from some named logical package, possibly at a
+    specific path within that package.
+3.  Distinct syntaxes for each of these.
+
+I tried a few different ideas, and my favorite is:
+
+### Relative imports
+
+Relative imports use the existing syntax:
+
+```scala
+// Relative path.
+import "ast/expr"
+```
+
+This looks for the file `ast/expr.wren` relative to the directory containing the
+module that has this import statement in it.
+
+You can also walk out of directories if you need to import a module in a parent
+folder:
+
+```scala
+import "../../other/stuff"
+```
+
+### Logical imports
+
+If you want to import a module from some named logical entity, you use an
+*unquoted* identifier:
+
+```scala
+import random
+```
+
+Being unquoted means the names must be valid Wren identifiers and can't be
+reserved words. I think that's OK. It would confuse the hell out of people if
+you had a library named "if". I think the above *looks* nice, and the fact that
+it's not quoted sends a signal (to me at least) that the name is a "what" more
+than a "where".
+
+If you want to import a specific module within a logical entity, you can have a
+series of slash-separate identifiers after the name:
+
+```scala
+import wrenalyzer/ast/expr
+```
+
+This imports module "ast/expr" from "wrenalyzer".
+
+## Implementation
+
+That's the proposed syntax and basic semantics. The way we actually implement it
+is tricky because Wren is both a standalone interpreter you can run on the
+command line and an embedded scripting language. We have to figure out what goes
+into the VM and what lives in the CLI, and the interface between the two.
+
+### VM
+
+As usual, I want to keep the VM minimal and free of policy. We do need to add
+support for the new unquoted syntax. The more significant change is to the API
+the VM uses to talk to the host app when a module is imported. The VM doesn't
+know how to actually load modules. When it executes an import statement, it
+calls:
+
+```c
+char* loadModuleFn(WrenVM* vm, const char* name);
+```
+
+The VM tells the host app the import string and the host app returns the code.
+In order to distinguish relative imports (quoted) from an identical unquoted
+name and path, we need to pass in an extra to bit to tell the host whether there
+were quotes or not.
+
+The more challenging change (and the reason I didn't support them when I first
+added imports to Wren) is relative imports. There are two tricky parts:
+
+First, the host app doesn't have enough context to resolve a relative import.
+Right now, the VM only passes in the import string. It doesn't tell which module
+*contains* that import string, so the host has no way of knowing what that
+import should be relative *to*.
+
+That's easy to fix. We have the VM pass in the name of the module that contains
+the import.
+
+The harder problem is **canonicalization**. When you import the same module
+twice, the VM ensures it is only executed once and both places use the same
+module data. This is important to ensure you don't get confusing things like
+duplicate static state or other weird side effects.
+
+To do that, the VM needs to be able to tell when two imports refer to the "same"
+module. Right now, it uses the import string itself. If two imports use the same
+string, they are the same module.
+
+With relative imports, that is no longer valid. Consider:
+
+```text
+script_a.wren
+useful_stuff/
+  thing_1.wren
+  thing_2.wren
+```
+
+Now imagine those files contain:
+
+```scala
+// script_a.wren
+import "useful_stuff/thing_1"
+import "useful_stuff/thing_2"
+
+// useful_stuff/thing_1.wren
+import "thing_2"
+
+// useful_stuff/thing_2.wren
+// Stuff...
+```
+
+Both `script_a.wren` and `thing_1` import `thing_2`, but the import *strings*
+are different. The VM needs to be able to figure out that those two imports
+refer to the same module. I don't want path manipulation logic in the VM, so it
+will delegate to the host app for that as well.
+
+Given the import string and the name of the module containing it, the host app
+produces a "fully-qualified" or "canonical" name for the imported module. It is
+*that* resulting string that the VM uses to tell if two imports resolve to the
+same module. (It's also the string it uses in things like stack traces.)
+
+This means importing becomes a three stage process:
+
+1.  First the VM asks the host to resolve an import. It gives it the (previously
+    resolved) name of the module containing the import, the imports string, and
+    whether or not it was quoted. The host app returns a canonical string for
+    that import.
+
+2.  The VM checks to see if a module with that canonical name has already been
+    imported. If so, it reuses that and its done.
+
+3.  Otherwise, it circles back and asks the host for the source of the module
+    with that given canonical name. It compiles and executes that and goes from
+    there.
+
+So we add a new callback to the embedding API. Something like:
+
+```c
+char* resolveModuleFn(WrenVM* vm,
+    // Canonical name of the module containing the import.
+    const char* importer,
+
+    // The import string.
+    const char* path,
+
+    // Whether the path name was quoted.
+    bool isQuoted);
+```
+
+The VM invokes this for step one above. The other two steps are the existing
+loading logic but now using the canonicalized string.
+
+### CLI
+
+All of the policy lives over in the CLI (or in your app if you are embedding the
+VM). You are free to use whatever canonicalization policy makes sense for you.
+For the CLI, and for the policy described up in motivation, it's something like
+this:
+
+*   Imports are slash-separated paths. Resolving a relative path is normal path
+    joining relative to the directory containing the import. So if you're
+    importing "a/b" from "c/d" (which is a file named "d.wren" in a directory
+    "c"), then the canonical name is "c/a/b" and the file is "c/a/b.wren".
+
+    ".." and "." are allowed and are normalized. So these imports all resolve
+    to the same module:
+
+    ```scala
+    import "a/b/c"
+    import "a/./b/./c"
+    import "a/d/../b/c"
+    ```
+
+*   If an import is quoted, the path is considered relative to the importing
+    module's path, and is in the same package as the importing module.
+
+    So, if the current file is "a/b/c.wren" in package "foo" then these are
+    equivalent:
+
+    ```scala
+    import "d/e"
+    import foo/a/b/d/e
+    ```
+
+*   If an import is unquoted, the first identifier is the logical "package"
+    containing the module, and the remaining components are the path within that
+    package. The canonicalized string is the logical name, a colon, then the
+    resolved full path to the import (without the ".wren" file extension).
+    So if you import:
+
+    ```scala
+    import wrenalyzer/ast/expr
+    ```
+
+    The canonical name is "wrenalyzer:ast/expr".
+
+*   If an import is a single unquoted name, the CLI implicitly uses the name as
+    the module to look for within that package. These are equivalent:
+
+    ```scala
+    import foo
+    import foo/foo
+    ```
+
+    We could use some default name like "module" instead of the package name,
+    similar to Python, but I think this is actually a little more usable in
+    practice. If you're hacking on a bunch of packages at the same time, it's
+    annoying if every tab in your text editor just says "module.wren".
+
+*   The canonicalized string for the main script or a module imported using a
+    relative path from the main script is just the normalized file path,
+    probably relative to the working directory.
+
+*   Since colon is used to separate the name from path, path components with
+    colons are not allowed.
+
+### Finding logical imports
+
+The last remaining piece is how the CLI physically locates logical imports. If
+you write:
+
+```scala
+import foo
+```
+
+Where does it look for "foo"? Of course, if "foo" is built into the VM like
+"random", then that's easy. Likewise, if it's built into the CLI like "io",
+that's easy too.
+
+Otherwise, it will try to find it on the file system. We don't have a package
+manager yet, so we need some kind of simple policy so you can "hand-author" the
+layout a package manager would produce. Borrowing from Node, the basic idea is
+pretty simple.
+
+To find a logical import, the CLI starts in the directory that contains the main
+script (not the directory containing the module doing the import), and looks for
+a directory named "wren_modules". If not found there, it starts walking up
+parent directories until it finds one. If it does, it looks for the logical
+import inside there. So, if you import "foo", it will try to find
+"wren_modules/foo/foo.wren".
+
+Once it finds a "wren_modules" directory, it uses that one directory for all
+logical imports. You can't scatter stuff across multiple "wren_modules" folders
+at different levels of the hierarchy. If it can't find a "wren_modules"
+directory, or it can't find the requested module inside the directory, the
+import fails.
+
+This means that to reuse someone else's Wren "package" (or your own for that
+matter), you can just stick a "wren_modules" directory next to the main script
+for your app or in some parent directory. Inside that "wren_modules" directory,
+copy in the package you want to reuse. If that package in turn uses other
+packages, copy those into the *same* "wren_modules" directory. In other words,
+the transitive dependencies get flattened. This is important to handle shared
+dependencies between packages without duplication.
+
+You only need to worry about all of this if you actually have logical imports.
+If you just have a couple of files that import each other, you can use straight
+relative imports and everything just works.
+
+## Migration
+
+OK, that's the plan. How do we get there? I've start hacking on the
+implementation a little and, so far, it seems straightforward. Honestly, it will
+probably take less time than I spent writing this up.
+
+The tricky part is that this is a breaking change. All of your existing quoted
+import strings will mean something different. We definitely *can* and will make
+breaking changes in Wren, so that's OK, but I'd like to minimize the pain. Right
+now, Wren is currently at version 0.1.0. I'll probably consider the commit right
+before I start landing this to be the "official" 0.1.0 release and then the
+import changes will land in "0.2.0". I'll work in a branch off master until
+everything looks solid and then merge it in.
+
+If you have existing Wren code that you run on the CLI and that contains
+imports, you'll probably need to tweak them.
+
+If you are hosting Wren in your own app, the imports are fine since your app
+has control over how they resolve. But you will have to fix your app a little
+since the import embedding API is going to change to deal with canonicalization.
+I think I can make it so that if you don't provide a canonicalization callback,
+then the original import string is treated as the canonical string and you
+fall back to the current behavior.
+
+## Alternatives
+
+Having both quoted and unquoted import strings is a little funny, but it's the
+best I could come up with. For what it's worth, I [borrowed it from
+Racket][racket].
+
+[racket]: https://docs.racket-lang.org/guide/module-basics.html
+
+I considered a couple of other ideas which are potentially on the table if
+most of you don't dig the main proposal:
+
+### Node-style
+
+In Node, [all imports are quoted][node]. To distinguish between relative and
+logical imports, relative imports always start with "./". In Wren, it would be:
+
+[node]: https://nodejs.org/api/modules.html
+
+```scala
+import "./something/relative"
+import "logical/thing"
+```
+
+This is simpler than the main proposal since there are no syntax changes and we
+don't need to push the "was quoted?" bit through the embedding API. But I find
+the "./" pretty unintuitive especially if you're not steeped in the UNIX
+tradition. Even if you are, it's weird that you *need* to use "./" when it means
+nothing to the filesystem.
+
+### Unquoted identifiers
+
+The other idea I had was to allow both an unquoted identifier and a quoted
+path, like:
+
+```scala
+import wrenalyzer "ast/expr"
+```
+
+The unquoted name is the logical part — the package name. The quoted part is
+the path within that logical package. If you omit the unquoted name, it's a
+straight relative import. If you have a name but no path, it's desugars to use
+the name as the path.
+
+This is a little more complex because we have to pass around the name and path
+separately between the VM and the host app during canonicalization. If we want
+the canonicalized form to keep those separate as well, then the way we keep
+track of previously-loaded modules needs to get more complex too. Likewise the
+way we show stack traces, etc.
+
+The main proposal gloms everything into a single string using ":" to separate
+the logical name part from the path. That's a little arbitrary, but it keeps
+the VM a good bit simpler and means the idea of there being a "package name" is
+pure host app policy.


underscorediscovery:
I definitely have more thoughts than this, but wanted to note my gut reaction to no string quote syntax is "oh no". 

I _really_ like that the import syntax is a string because it let's you do stuff that the parser would choke on without modifications, for example `import "package: module"` like dart becomes an error/impossible. If the intent is not to stifle the embedding import flow then that aspect specifically doesn't fit (and in fact would break my whole package system). The parser could just read until `for` or `eol` as an import, and treat _that_ as the string though, which amounts to allowing both? 

I'm just not sure the syntax difference there is as clear as hoped... consider for example the difference between:

`import "non relative"
`require "relative"`

Here the difference is stark and obvious, where quotes/no quotes is a bit more opaque. I don't hate the idea, I just don't want to lose the flexibility gained from having a plain old string based import path. That doesn't even consider things like `import "http://"` or anything that is not a simple file on disk, so it's quite confining :)

I also liked that the quotes kept things a little more obvious/clearer, like the space in this case makes the package seem like a keyword of sorts, but not sure if that's a big deal.
`import package: path/things for Stuff`

There's some considerations on the paths too related to IDE's and canonicalization when reporting errors and things too that might need noting.

munificent:
> I definitely have more thoughts than this, but wanted to note my gut reaction to no string quote syntax is "oh no".

OK, good to know. I kind of like how it looks unquoted, but I agree it's also kind of fishy. It's one of the main things I wanted a gut check on.

> ```
> import "non relative"
> require "relative"
> ```

Ah, yes. I should have mentioned that in the alternatives. Another idea I considered was some kind of modifier in the import to distinguish relative and non-relative. I didn't consider an entire different keyword like you have here. I was thinking more like:

```
import "package/thing"
import from "something/relative"
```

But I couldn't come up with a modifier I really liked. "from" was the best idea I had but even then I couldn't decide if that sounded more like the thing you'd use for a relative import or a logical one.

An entirely different keyword like `import`/`require` would work, though I'd like to keep the number of reserved words low if possible. Also, the distinction between those two words is pretty arbitrary. Coming up with a pair of words that send the right signal is probably hard.

underscorediscovery:
> Another idea I considered was some kind of modifier in the import to distinguish relative and non-relative. 

It's worth stating the obvious one for the sake of it, that there is a "relative modifier" for this already in `./relative` vs `notrelative`. I'd consider it being required as acceptable, since it's already a signifier of a relative-to-this-module import.

I like the simplicity of `import "module" for Stuff (as Things)` :P but adding more words feels wordy... and syntax is less ideal too probably like `import @"stuff/thing"`.

munificent:
> It's worth stating the obvious one for the sake of it, that there is a "relative modifier" for this already in `./relative` vs `notrelative`. 

Yeah, I mention that option in the alternatives section. I'm not super crazy about it (the "./" feels nonobvious and magical to me), but if lots of others really like it, I could be persuaded.

underscorediscovery:
> Yeah, I mention that option in the alternatives section

My brain glossed that bit >.> I don't think it's the _best_ option for that either.

mhermier:
What do you think about an alternative synatax since we have the 'foreign' keyword available.
So instead of introducing 'from' we would have:
import "package/thing"
import foreign "path/like"

Otherwise personally, I would prefer the shell way of resolving path, with the twist of resolving relative to the current script. So everything is treated as a local path first (unless an absolute path is given). And the resolution should follow a path list. If no source file is found, then it tries to load an internal module.
This is for me the more interesting way of doing, so the user can override an internal module and be more *future* proof.

munificent:
> What do you think about an alternative syntax since we have the 'foreign' keyword available.

That's not a bad idea, but I think it would be good to reserve `foreign` only for things related to the embedding API so that it sends a clearer signal to users. (Think about how confusing terms like "static" and "virtual" are in C/C++ because they are used for such unrelated things.)

I do like the idea of using some kind of modifier, though. Maybe "relative"?

```
import "package/thing"
import relative "path/like"
```

That's pretty long, but it is explicit.

> This is for me the more interesting way of doing, so the user can override an internal module and be more future proof.

Yes, that's a good point. A logical import of a local package should take precedence over a built-in one. Otherwise, upgrading to a new version of a CLI could break your app if it adds a new built-in module that shadows one they are already using. I'll update the proposal to mention that.

bjorn:
I just thought I'd throw in a reference to another system which used quoted vs. non-quoted imports: QML. Example from [the documentation](http://doc.qt.io/qt-5/qtqml-syntax-imports.html):
```qml
import QtQuick 2.0 as CoreItems
import "../textwidgets" as MyModule
```
(the "as X" part is optional for both lines, when left out any items in that module or directory are available without prefix)

To me this always seamed quite natural. The unquoted version clearly indicates some resolving magic is going on (it will be using modules defined in C++ and search the import paths). Whereas with the quotes it is clearly some kind of file reference  (it can even be a URL, as long as there is a `qmldir` file that defines the items available). There is no absolute vs. relative distinction, just a difference in how the module is looked up.

Also, dots are used instead of slashes in module identifiers, for example:
```qml
import org.example.Functions 1.0
import Qt.labs.controls 1.0
```
I think this works nice in further distinguishing them from directory imports (though in terms of lookup, [they would require](http://doc.qt.io/qt-5/qtqml-modules-identifiedmodules.html) `org/example/Functions/` to be in the [import path](http://doc.qt.io/qt-5/qtqml-syntax-imports.html#qml-import-path)).

In QML the version number is actually not optional for module imports, and can be used to maintain a certain level of compatibility. However I could imagine Wren doesn't need this.

munificent:
> Also, dots are used instead of slashes in module identifiers

I considered that too, but went with slashes for two reasons:

1. It makes converting from one form to another a little easier. This might not be a big consideration. (In fact it might be better to deliberately make the two forms more visually distinct.)

2. Using slashes leaves open the possibility for a dotted identifier as the first component, as in:

    ```
    import some_framework.testing/mocks
    ```

    That could be used to support some hierarchical notion of package names independent of the path *within* the package. I'm not generally a fan of hierarchical namespaces like this (for long complex reasons I won't get into here), but it doesn't hurt to leave the door open.

underscorediscovery:
To add data points, the dot syntax matches many package based languages like Java, Actionscript, Haxe and more. I prefer paths because they're paths inside a module and not classes/members/properties and the ambiguity in languages that have package.Class and then Class.stuff can be really confusing to people learning to program.

mhermier:
if we really want to add a new keyword why not do something like:

```scala
builtin "package/thing"
import "path/like"
```

And personally, I really don't like the addition of *extended identifiers* it just only bloat the language/parser. While I see the benefits of having package name as identifiers, I don't really see how it could fit the language nicely.
So for me it should be all encapsulated in a string representing an URL/URI to the module.

bjorn:
> Yeah, I definitely don't want version numbers in the import strings since that doesn't play nice with version resolution on shared dependencies. I'm a big fan of Bundler's model where versions are constraints which are solved and stored in a separate lockfile.

But with Bundler, the dependencies aren't actually shared, right? Each project adds the dependencies to its own local folder. The version numbers in QML are needed exactly because the dependencies are shared between projects, and help resolve ambiguities that would be caused when a newer version of a module adds members that conflict with other modules. In QML the module can expose a different API based on the version requested, rather than the version determining which module is loaded.

underscorediscovery:
Version numbers in the import line do not scale past a single file or two. A simple version bump would mean changing 100's of files, in my usage a package contains several wren modules and almost every file uses the core library package in some way or other. That level of micro management is a nightmare to me.

Does the bundler one look like dart and rust and others? It has a single list of dependencies for a project that get a dependency tree, then resolve for the highest satisfied version for each dependency, and then store it in the lockfile?

munificent:
> I prefer paths because they're paths inside a module and not classes/members/properties and the ambiguity in languages that have package.Class and then Class.stuff can be really confusing to people learning to program

Yeah, I agree overloading `.` is confusing.

> And personally, I really don't like the addition of extended identifiers it just only bloat the language/parser. 

Yeah, I'm not a fan of contextual keywords that act like a reserved word in some context but not others, like "yield" in C#. They make sense in later versions of a language where you want to introduce new syntax without breaking existing uses, but for a <1.0 language, I don't like them. It makes tools like syntax highlighters more difficult to do well.

> But with Bundler, the dependencies aren't actually shared, right? Each project adds the dependencies to its own local folder.

They are shared within the transitive dependency graph of a single project. Let's say my_app uses a and b which both use c. With Bundler, you'll get a single flat set of packages for my_app: a, b, and. a and b will share the same version of c.

> Does the bundler one look like dart and rust and others?

Yes, exactly. Natalie and I designed Dart's package manager after Bundler and Rust's cargo was written by Yehuda Katz, the creator of Bundler.

underscorediscovery:
> Yes, exactly. Natalie and I designed Dart's package manager after Bundler and Rust's cargo was written by Yehuda Katz, the creator of Bundler.

That's the approach I opted for in my engine package system too fwiw.

bjorn:
> Version numbers in the import line do not scale past a single file or two. A simple version bump would mean changing 100's of files, in my usage a package contains several wren modules and almost every file uses the core library package in some way or other. That level of micro management is a nightmare to me.

I think it scales fine because it's just a matter of search & replace to update that version. But it the end, it's needed to avoid ambiguities like if you did:
```qml
import foo
import bar

FooBar {   // resolves to bar.FooBar
}
```
But now some new version of `foo` also adds a `FooBar` item, now your app would not run anymore because it does not know which version you want. So by doing:
```qml
import foo 1.0
import bar 1.0
```
You're explicitly saying what version of the API you expect, and even `foo` 2.0 is installed with a `FooBar` item, it will not expose that in this file and no ambiguity problems appear.

Now, this is all on the assumption that `foo` and `bar` might be modules provided by the system, and not modules that you ship or install specifically for your app like you do with Bundler. So which version your app runs against is outside of your control, and you specify it only to get a certain amount of backwards compatibility. As such there is also no need to actually increment these version numbers unless you're using types or properties that were introduced after it (and thereby you also increase the minimum version your app requires).

Of course, all of this doesn't really apply to Wren because it is meant to be embedded rather than being a language to write applications in that depend on high-level functionality provided by the system. In the case of Wren, the developer would have full control over the versions of the dependencies, and then having an additional level of versioning on the imports is of course pointless.

> They are shared within the transitive dependency graph of a single project. Let's say my_app uses a and b which both use c. With Bundler, you'll get a single flat set of packages for my_app: a, b, and. a and b will share the same version of c.

And if a and b depend on different versions of c, I guess it will raise an error? In any case this is indeed not the kind of sharing that would need versioned imports.

underscorediscovery:
Oh yea, if they're system wide and fed directly into the code with no version on the project level that would be pretty awkward to me! What you say makes sense though, I will still hesitate to consider it inconsequential to litter every file with a version and make the user micro manage it however, but it bears no consequence here :)

Regarding "the different versions of c", the descriptions and approaches to version lock and conflicts of versioning are well laid out in this dart documentation and explain the approach:

https://www.dartlang.org/tools/pub/versioning

munificent:
> And if a and b depend on different versions of c, I guess it will raise an error? In any case this is indeed not the kind of sharing that would need versioned imports.

The doc @underscorediscovery links too has the full story, but the short answer is that a and b don't (ideally) depend on specific versions of c. Instead, each specifies a range of versions they are compatible with and the package manager picks a version of c that satisfies both of those constraints.

jlauman:
Thank you Bob, and community, for this spectacular embedded scripting language. 

I just spent two days working on an embedding and haven't developed a historical knowledge of this project (from the message board), so please excuse and ignore any naive comments.

There seems to be a love triangle here between Wren, the host application and a future package manager. In a complete (future) ecosystem an end-user is writing Wren scripts/modules and using a package manager to pull down hosted modules from a repository, but may not be intimate with the the host application. In my case, I know that the CLI uses libuv, but I don't need to know the details of the embedding to use the CLI and package manager tools.

I will attempt to synthesize a complete Wren module story here from what has already been discussed with one or two new pieces. 

First, the `resolveModuleFn` in the configuration is a must-have. As stated in RFC-0001, the return string of the resolve function is the canonical string that the Wren VM uses establish module identity. It is also necessary to provide coherent diagnostic information to the end-user. Wren doesn't need to know the context of the canonical string, it just needs to report it when a compile error occurs.

```c
char* resolveModuleFn(WrenVM* vm, const char* importer, const char* module);
```

The `resolveModuleFn` signature I've shown here doesn't have the `isQuoted` argument because I believe that is information that can kept internal to Wren and is explained later.

The next piece is establishing the relationship between the package manager and the host application. I propose that this is specific to the host language and environment, but that the Wren community can establish conventions for the most common cases. For example, on mainstream operating systems an environment variable may be set by the package manager in the end-user's account, or read in from a file like `~/.wren/module_path.txt`. 

```scala
WREN_MODULE_PATH=/usr/local/share/wren/0.2.0/modules/%(module).wren;%(HOME)/.wren/0.2.0/modules/%(module).wren;%(importer_dirname)/wren_modules/%(module).wren;./%(module).wren
```

The `WREN_MODULE_PATH` may look familiar to Lua users since I lifted `LUA_PATH` from the Lua reference manual and Wrenified it. It does not include the scenario that is identified by Node.js as LOAD_AS_DIRECTORY (in the link below).

Alternately, the host application may exec the package manager binary to return the module search path when the host application starts. 

The next piece is the rules for module resolution that the host application must implement. The simpler the rules the better. Node.js triggered an explosive non-browser JavaScript ecosystem, and has a well document set of rules for module resolution.

https://nodejs.org/dist/latest-v8.x/docs/api/modules.html#modules_all_together

For Wren, we could develop a reference implementation of module resolution in Python as an example for other developers to port to their own host environment.

**All of this, of course, does not force a developer to adopt the community conventions when embedding Wren in their application.**

Given the prior pieces as host application and Wren configuration infrastructure, the next section proposes how modules work inside of Wren.

One additional first-class type will be needed to support the following examples -- the `Module`. I know that the cost of adding complexity has to be weighed against the value provided, so I will forge ahead and allow you to make that determination.

A simple example that is similar to a Wren 0.1.0 import shows `OldOnes` coming into scope as a `Module`. Once a module is in scope and assigned to a variable its contents may be accessed. 

```scala
// lovecraft.wren
import from "lovecraft-mythos/great-old-ones" as OldOnes
var Cthulhu = OldOnes:Cthulhu
```

This adds some verbosity and an additional piece of syntax to pull out a class or variable from a module. I agree that overloading `.` could confuse an end-user about what is being assigned. This syntax for `import` focuses the purpose of the statement -- it brings a `Module` type into scope (and doesn't try to do anything with the module internals).

The next examples extend the basic use of `import` and use `Random` from the current Wren documentation as a module name.

The following import brings `Random` module into scope and then `Random` class is assigned to a variable. There could some kind of sugar provided to make single class in module less verbose if it is a common scenario.

```scala
import Random
var Random = Random:Random;
```

The next example shows bringing the module into scope as `OtherRandom`. The goal for this is to prevent namespace collisions, and give the script writer control of naming within the script. This use of `import` signals that `Random` is provided from inside of Wren, or that it already exists in the current scope. (This is where `isQuoted` may not need to be an argument of the `resolveModuleFn` function.)

```scala
import Random as OriginalRandom
```

Again, in this next example assumes that `Random` is built into Wren as a `Module`. This example shows bringing two random modules into scope, one is the builtin and the other is resolved by the host application.

```scala
import Random as OriginalRandom
import Random from "experimental/random" as OtherRandom
var Random1 = OriginalRandom:Random
var Random2 = OtherRandom:Random
```

The interesting part here is the second import -- it specifies `Random` after the import where it could have done a vanilla module import like `import from "experimental/random" as OtherRandom`. In this case the builtin `Random` module is used to establish a module signature (using classes, methods and arities) to compare against the module loaded from `"experimental/random"` and ensure that `OtherRandom` at least includes the signature items from `Random`.

The other use of this mechanism could be to establish module templates (like Java interfaces or Swift protocols) that are builtin to Wren, but do not provide implementations. If not builtin to Wren, then they could be loaded from outside.

```scala
import from "wren/interface/random-0.2.0" as Random
```

Once imported `Random` can now be used as the interface for loading a module that implements the `Random` signature.

The starting Wren scripter doesn't have to be aware of these module mechanisms, and can leverage a batteries-include distribution if one is provided. As Wren evolves and larger applications are written in Wren establishing predictable module behavior is important. I believe that a package manager (and repository system) could behave like Elm (and probably others I'm unaware of) and ensure modules maintain backward compatible signatures as they evolve.

One final note. I'd like to see an `export` keyword to control what is visible in a module rather than rely on an uppercase identifier. (Just expressing a preference here.)

```scala
// great-old-ones.wren
export class Cthulhu {
  construct new() {}
  message { "Ph'nglui mglw'nafh Cthulhu R'lyeh wgah'nagl fhtagn" }
}
```

Thank you for your time. I'm looking forward to contributing to the Wren project.

munificent:
>  I propose that this is specific to the host language and environment, but that the Wren community can establish conventions for the most common cases.

❤️  Yes, I totally agree. The VM should be mostly policy free. The CLI should implement a policy that makes sense for it. And the entire community should have a wider set of conventions to help ensure other host applications and tools play nice together.

In particular, I hope that eventually there will be good editor/IDE support for Wren. That means editors that read Wren files, traverse their imports, and find the imported files. For that to work, we'll need good conventions so that each editor doesn't need to have separate hard-coded support for a variety of host applications' disparate policies.

> All of this, of course, does not force a developer to adopt the community conventions when embedding Wren in their application.

+1!

> One additional first-class type will be needed to support the following examples -- the Module.

First-class modules are definitely on my mind. I have put a *lot* of effort into trying to design and implement something along these lines. So far, nothing has gelled, but I haven't given up.

However, that's a little orthogonal to this proposal and I'd like to keep the two concepts separate in the interests of being able to make incremental progress. This proposal is basically about how you go from an import identifier/string to a blob of source code that the VM can compile and run. The semantics of what that blob represents — a first-class module object, script, etc. — can be separated out.

Roughly, think of it like an import statement containing two pieces. The first half is what you're importing. It might be specified as a string literal with some policy, identifier, dotted identifier, etc. There's a resolution process and some policy. Then the latter half of the import statement is `as ...` or `for ...` or some combination. It's what you do with the resulting imported stuff and how you make it accessible to the importing module.

This proposal is just about the first half. If we can get this into a good state and land it, I'm definitely interested in looking at improving the latter half too.

jlauman:
Got it -- focus on the *what* of the import statement as you said. Incremental change is good.

So, given a `resolveModuleFn` function in configuration (without `isQuoted`)...

```c
char* resolveModuleFn(WrenVM* vm, const char* importer, const char* module);
```

Wren VM could implement the following steps for module resolution:

1. Wren VM internal module is imported when identified.
2. If module is internal, Wren VM still calls `resolveModuleFn` function for canonical string (which could be NULL).
3. If `resolveModuleFn` function returns a canonical string then a warning is issued
4. If module is not internal then canonical string from `resolveModuleFn` is used to load external module.

The warning callback would follow the pattern of the error callback.

```c
void (*WrenWarnFn)(WrenVM* vm, WrenWarnType type, const char* module, int line, const char* message);
```

And, the warning would be something like, "internal module 'random' shadows external module '.../my-modules/random.wren'" (the canonical path). If I want my specific "random" module I can then change the import to be more specific, "my-modules/random", and the warning goes away.

Current `import` syntax stays the same, and road is clear to work on module resolution conventions and tools.

mhermier:
For the default resolution of path in the CLI, instead of a dedicated file
with a random syntax (INI, wren, what ever) I would go shell PATH
environment way, with additionnal command line path. It is a KISS and
well-known solution, which would allow to have and test multiple build of
wren at once (and maybe simplify test suite, if we can automate autoloading
of native module that way).
In extra, I don't think wren CLI would reach a popularity and
expressiveness to replace a shell, so I think a custom user file is
overkill for now.

jclc:
This might not be very popular, but what about using `<angle_brackets>` instead of `"quotation_marks"` for absolute imports, akin to C/C++?

skybrian:
I don't see any problems.

One corner case: is it possible to run a script *within* the wren_modules directory? If so that would change the search algorithm for *finding* the wren_modules directory. Not sure it's needed though.

It seems like it would be pretty easy and clean to skip relative imports altogether? Just put all your modules in a wren_modules directory. There are a couple nice things about it:

   *  source code always refers to a module using its canonical name, making it easy to search for all imports of a module even with search tools that aren't Wren-specific.
   * It's easier to to see the difference between scripts and modules. Anything outside the wren_modules directory is a script.

But since the VM shouldn't prescribe policy, it probably should have a way to hook in a canonicalization function.

For what it's worth (and some kind of user survey would be more helpful here), I think starting relative imports with "./" or "../" is easy and natural. I like this a bit better because then everything in the quoted string is opaque to the VM. It doesn't even know if the environment has different kinds of imports. The quotes also work a bit better with search-and-replace.

On the other hand, just knowing that there two syntaxes for imports (quoted and unquoted) and leaving their interpretation to the environment doesn't seem so bad either. An embedder doesn't have to use both syntaxes.

mhermier:
For me string quoting is a must have feature. It would protect the import
from potentially problematic characters, it doesn't require any change to
lexing/parsing. And is more future proof. To add to the wheel, I don't see
why one should not be able to do:

import MyExpression.call() for Foo

CodogoFreddie:
For my money, I strongly dislike having imports be both quoted and unqutoted. You're right that `./` for local imports doesn’t actually make a huge amount of sense, but I think the solution is to have two different keywords/a modifier to select between local and global imports

munificent:
I made [a bunch of progress on the implementation](https://github.com/munificent/wren/tree/smarter-imports). In particular, the machinery for letting the host resolve a relative import in the context of the importing library is there. I'm at the point now where I need to pin down the syntax. The comments here have persuaded me that unquoted-slash-separated isn't a winner, so let's try to figure out what is.

I tried out a bunch of different ideas and [wrote up my thoughts on them](https://gist.github.com/munificent/b4373df431b7a07a6c7a74bff857cedd). My two favorites right now are:

```
// Use
use "relative/path"
import "logical/path"

// Node-style
import "./relative/path"
import "logical/path"
```

If you folks are OK with `use`, that's my preference. But otherwise, the Node style will definitely work too. I'm open to other ideas as well, including a few in the linked doc, but I'd like to not bikeshed this forever. Thoughts?

CodogoFreddie:
I personally prefer `use`, I was about to champion `require` until I read your points about it :)

If you'd like to get a quick resolution on this, you could make two separate posts on this PR, and use reactions to vote for the most preferred? We could still have comments, but it would direct the discussion towards a conclusion.

underscorediscovery:
Thinking about it a bunch, I like `use`, and think it's a fine option. 

I'd also be ok with `./` specifically because it's how paths work in general and isn't surprising, and isn't unique to wren (vs something like `use`). It doesn't originate in node, so that point doesn't matter as much to me.

Having already implemented a whole package system and dealt with this distinction, _having_ the distinction exist is definitely helpful. Especially, having access to "from where did this relative originate", as we know. 

**shortest path**    
In that regard, one way that seemed possible to _just_ expose the origin of the import (and do nothing else), on all imports, like:

```c++
typedef const char* (*WrenLoadModuleFn)(WrenVM* vm, const char* module, const char* origin_module);
```

From here, the host has enough information to distill a relative path if they so chose (using whatever scheme the host wanted) and can do the resolution. Consider for example a relative include from a logical import.

- in `game.wren` - `import "package: file"` - `loadModuleFn("package: file", "game.wren")`
- in `package: file` - `import "other_file"` - `loadModuleFn("other_file", "package: file")`

From here the host can trivially do:
```c++
root = path_directory( resolve_module_to_path(origin_module) );
path = path_join(root, module);
```

**rigidity of use and ./ ../**   
The above frees the vm of "which is relative??" policy, and keeps the code nearly the same. It moves the distinction onto the host, which is where it has actual meaning. For example, does `/file.wren` mean "from the project root" or is it an absolute path? is that `logical` or is that `relative` so do I `import` or `use` for that path? 

I think there's a lot of different schemes and ways to handle imports, and it gets more nuanced when doing packages (because relative becomes relative to a logical module). I don't know if having `use` around will only make that harder. 

I also don't know if the vm doing `startsWith("./") || startsWith("../")` is actually doing anything of value except making the import logic more specific and rigid. I would let the host decide what relative means, and just give them the tools needed to resolve it.

**import string to file on disk resolution**    
Since this is semi related, worth noting that for the debugger we added a notion of `WrenModulePathFn` as a placeholder, which `// Returns the path to the physical file representing the module [name], or null if the file is not physical.`, this doesn't relate to imports specifically but figured I'd mention there's another "ask the vm to ask the host" thing lurking.

jclc:
`use` sounds great, I hadn't even thought about that option.

munificent:
> In that regard, one way that seemed possible to just expose the origin of the import (and do nothing else), on all imports, like:
> 
> ```
> typedef const char* (*WrenLoadModuleFn)(WrenVM* vm, const char* module, const char* origin_module);
> ```
>
> From here, the host has enough information to distill a relative path if they so chose (using whatever scheme the host wanted) and can do the resolution. Consider for example a relative include from a logical import with a uri schemeish syntax.

It's a little more complex than that because we need to ensure that import canonicalization works — if you import the same module twice, it should only be loaded and executed once. That means the VM needs to be able to reliably tell what "same module" means, even though it doesn't own the policy for how the parent module's path is combined with the import path.

To address that, the in-progress branch I have [splits importing into two separate operations, resolution and loading](https://github.com/munificent/wren/commit/8a71735e0f18b2f8f6022f5c8c8fe1044bbc741d). Resolution is where the host combines the parent module's path and the import path into a single canonical string. Then, if a module with that resolved string has not already been loaded, then the VM calls the load function to load it.

(If your host app doesn't care about relative imports and you want to use the existing behavior, you can omit the resolution callback. When you do that, the default resolution behavior is to use the import string as-is.)

I have that implemented, and I'm pretty happy with how it's working. It gives the host app the context it needs to support relative imports, keeps the VM policy-free, addresses redundant imports, and doesn't add too much complexity to the VM.

underscorediscovery:
@munificent right yea, the edit that I added immediately after writing it had the same realization, and I added the 

>The mechanics so far seem to make sense and help the host and vm well.

for that reason too. 

Only thing I was last wondering was in your thinking `/path` is a logical path (not a relative one) correct? so it would be `use "./path"` and `import "/Users/blah/path/"`, which might not matter, but worth noting

skybrian:
I think either ./ or something like "import local" would be more like
pseudocode, in the sense that someone who doesn't know Wren could guess
what it means. By contrast, "use" and "import" are near synonyms, so it's
easy to get it backwards.

mhermier:
Personnaly I don't really care about the keyword, I'm more interested in
the functionality. As such the "./" Form looks more interesting. Since it
would allow local and built-in module loading.
It would only one keyword, and it let the other keyword available for
built-in only loading. Security wise I think it is an interesting feature
to have built-in module integrated in the default include, but have a
dedicated form for them.

skybrian:
On Thu, Mar 29, 2018, 10:59 AM Bob Nystrom <notifications@github.com> wrote:

> I'm not sure if that's true for "./", though I could be wrong. My
> impression is that it sort of makes sense to people with a UNIX background,
> but may mean absolutely nothing to someone outside of that tradition.
>
It's certainly a Unixism, but I think the influence of Unix is pretty
widespread, at least among more experienced programmers? After all, people
are pretty used to forward slashes these days. :)

(There is even a rare case where you need to use ./ in a relative URL. This
happens when the same HTML for a directory listing gets served both with
and without a trailing slash.)

But without a user study, this is all hand-waving.

>

munificent:
I started hacking in support for a `use` keyword and then plumbing it through to the resolver API here. You can see it here: https://github.com/munificent/wren/commit/0485c8de21bf4fda1efc6330d56317a108907880.

Now that I look at it, I'm not a fan. Seems like it adds quite a bit of complexity to push a single bit all the way through the compiler and runtime, especially when that bit doesn't actually mean anything to Wren itself or the VM. The end result, I fear, is having to explain to users over and over again why there are two kinds of imports and what they mean.

How about we just do Node-style for now? That's the simplest implementation, the simplest semantics, and fits with Wren's minimal nature.

underscorediscovery:
sounds good to me. 
The resolver bit makes it most useful anyhow!

mhermier:
All this bloat comes from the fact that import is an opcode. With old style
it would be cleaner, since it would not change the VM.

marcobambini:
I am implementing an import system for my Gravity programming language and I am following this discussion with a lot of interest. I'd like to add my two cents.

As a general rule of thumb I don't like to have different keywords for the same operation, the main action is an **import** and we just need a way to distinguish between a path (relative or full) and a logical import. So here you go my proposal:

```
import "./relative_path" (equivalent to import "relative_path")
import "/full_path"
import "@logical_path"
```

jclc:
"path:///" before every logical import? Are you actually being serious?

marcobambini:
@mhermier banning the @ symbol? It is just the first character that should be interpreted, everything else should be used as is.

mhermier:
@ is just a marked that forbid the usage of "@" beeing the first symbol of your file/directory.
Since it is a valid character on windows https://msdn.microsoft.com/en-us/library/windows/desktop/aa365247(v=vs.85).aspx I don't think it is a good idea to restrict file naming even more.
It would have been better to choose at least one of the forbidden windows character (I don't speak about linux/unix since the VFS axcept almost everything but '\0').

Don't get me wrong, I see the benefits of using a single letter as a marker. And of all possible systems I would prefer to have mount points or hurd translators. But considering implementation difficulties, its versatility, the fact that url was made wisely and its wide spread usage, it become a solution to consider.

mhermier:
And for the reference Qt use something similar for its resources (http://doc.qt.io/qt-5/resources.html) by choosing to do ":/path".

marcobambini:
Not sure to understand... if you need to reference a file that starts with the @ symbol you should use:
`import "/@myfile"`
or
`import "./@myfile"`

mhermier:
Does this all means that we have to always prefix our path ?
This is how I see the things:
1. `import "path"` local else logical
2. `import "./path"` local
3. `import "/path"` global
4. `import ":/path"` logical

mhermier:
I forgot about builtin, so something like `import "|/path"` or url style `import "builtin:///path"`.

heyitsmeuralex:
+1 to node-style `./` imports. Seems simplest, and easiest to understand.

The `import "@logical"` idea is... ehh. I'd prefer a secondary keyword if we were going to go that route, eg. `import pkg "my-package"` for logicals and `import "path/to/file"` for locals.

munificent:
I think if we start needing a lot of sigils and magic syntax inside import strings then we've gone off into the weeds and don't have a good solution. The policy I'm currently imagining is:

* If the import string starts with "./" or "../", it is a relative import. (We may also extend this to include absolute paths that start with "/" or "C:/", etc.) Join it with the resolved path of the containing module and look for a ".wren" file at that path.

* Otherwise, it is logical.

    * First, we look for a "wren_modules" directory in any of the directories the entrypoint script being run. If we find one:

        * If the import path has no "/", implicitly replace it with "<path>/<path>". So `import "foo"` becomes `import "foo/foo"`. This rule is so that every entry immediately inside "wren_modules" is a directory, and not a mixture of directories and bare files. At the same time, if your package only exposes a single module, I don't want users to have to explicitly write `import "foo/foo"` (which you do have to do in Dart, and it sucks).

        * Take the first component of the import path. So if you have `import "foo/bar/baz"`, take "foo". Look for a directory inside "wren_modules" with that name.

        * If found, the import is treated like an import of the corresponding module in that directory.

    * If we get here, we failed to find a directory with name of the first component in the import path inside a "wren_modules" directory. In that case, assume it must be a built-in module and try to import that.

        We look in "wren_modules" for a logical import first to allow them to shadow built-in modules. I don't think that will be done often in practice, but it's important to support so that Wren releases that include new built-in modules do not spontaneously break existing user programs that may have a module with the same name.

This policy is a little complex, but I think it covers all of the cases I care about. The short summary of the policy rules is:

* Paths starting with "./" or "../" are relative to the current module.
* Paths with a single component are treated as "<path>/<path>".
* Logical modules look for a directory with the first path component's name in a nearby "wren_modules" directory.
* Otherwise, they are assumed to be built-in.

So, given this file structure:

```
my_app/
  wren_modules/
    random/
      random.wren
    parser/
      ast.wren
  io.wren
  main.wren
  utils/
    collections.wren

Here is what "main.wren" could contain and what its imports would resolve to:

```
import "io" // Built-in io
import "./io" // my_app/io.wren
import "random" // wren_modules/random/random.wren (shadows built-in)
import "parser/ast" // wren_modules/parser/ast.wren
import "./utils/collections" // my_app/utils/collections.wren
```

Here is what "my_app/utils/collections.wren" could contain and what its imports would resolve to:

```
import "io" // Built-in io
import "../io" // my_app/io.wren
import "random" // wren_modules/random/random.wren (shadows built-in)
import "parser/ast" // wren_modules/parser/ast.wren
```

I realize this sounds pretty complicated, but my belief is that it mostly does what you want, even if you don't realize it. Other languages have surprisingly complex import resolution policies too. Also, all of this policy is in the Wren CLI, not the VM or language itself.

underscorediscovery:
I 100% think the `./` or `../` "tagged as relative" is a perfectly fine solution, it doesn't impose policy on the import string, and doesn't create ambiguity. 

on the `/` thing: I would **_not_** try to handle absolute paths or `c:\` and friends because 
1) it's a huge rabbit hole, infinitely wide. network drives, cross platform, it's a real mess the vm can't and should not have to answer. 
2) it's making the host have to dance around vm decisions, which is the opposite of the vm intent. If for example the rule was `<scheme>://<path>` then `custom://` form comes in as relative and not logical, which is 'wrong', so the host code has to figure out their own intent in multiple places based on vm rules. 

It's much more preferable that "relative import" is a finite, and well defined concept that the vm is facilitating. `./` and `../` mean something explicit and give you a hint. anything else is host decided, and that's how I think it should be. No magic, no being clever :)

munificent:
> it's a huge rabbit hole, infinitely wide. network drives, cross platform, it's a real mess the vm can't and should not have to answer.

For what it's worth, I have personally implemented a very robust, cross-platform `isAbsolute()` function. It's actually not *that* bad. You have `/` for (UNIX, OS X, and root-relative Windows paths), drive letters, and UNC paths.

But, agreed, it's a pain that we probably don't need to worry about. Imports of absolute paths are *weird*.

> it's making the host have to dance around vm decisions, which is the opposite of the vm intent.

I should stress that everything in my previous comment is policy that lives *entirely* in the host. The VM doesn't know *anything* about import resolution, not even "./" and "../". From its perspective, everything is an opaque string. The only thing it knows is that a host may need to resolve an import in the context of the importing module's import path (which, again, is an opaque string).

underscorediscovery:
ah ok, right right, so you meant that the absolute handling is CLI, not VM. Gotcha

mhermier:
That sounds a little be complicated, trying to solve module packaging at
first glance. In particular having to deal with local to app modules.

skybrian:
Regarding resolving logical paths, also keep in mind that you could automatically add the extension: `import "io"` and `import "io.wren"` could do different things, where the latter is more obviously a filesystem path, and the former is a bit more magical, where you need to know the rules.

This might even make a decent alternative to the "./" prefix.

underscorediscovery:
Oh yea, I didn't mention but that's what I use atm as well, logical has no extension, relative does. The only downside is you can have relative files without the wren extension for several reasons, and this becomes ambiguous again @skybrian  :)

bjorn:
I'm not sure if I'm adding anything of value here, but I'm just sad to see the dotted syntax being discarded, especially with the reason given (looking similar to getter calls). The dotted syntax is so common (Cobra, D, Java, C#, Kotlin) that you'd hardly expect anyone to dislike it for that reason.

skybrian:
A slight variation on the above: suppose we say that paths ending with ".wren" look like a file, while paths with no extension look like a directory?

So, for paths that look like a directory (such as "foo/bar"), assume the directory exists in wren_modules, double the last segment and add ".wren" to get "$wren_modules/foo/bar/bar.wren".

For paths that look for like a file: would it suffice to assume they're _all_ relative? Otherwise, the "./" prefix is still useful to resolve the ambiguity.

munificent:
>  also keep in mind that you could automatically add the extension: import "io" and import "io.wren" could do different things, where the latter is more obviously a filesystem path, and the former is a bit more magical, where you need to know the rules.

Oh, I should have mentioned that, but I implicitly add the ".wren" extension when looking up all imports. It seems perverse to me to require the user to write ".wren" in every import. Of *course* you're importing a Wren file. What else could it be?

> Oh yea, I didn't mention but that's what I use atm as well, logical has no extension, relative does.

Interesting. We could use that as the trigger to distinguish relative and logical imports, but it feels kind of strange to me. Once we have packages, users will hop between relative and logical imports pretty frequently, I think. I believe it will trip them up to have to remember which ones need an extension.

The "./" makes more sense to me there because you *have* to think about *where* you're importing the module from. Whereas the ".wren" feels disconnected from that to me. After all, inside the "wren_modules" folder, those files all have ".wren" extensions too.

> I'm not sure if I'm adding anything of value here, but I'm just sad to see the dotted syntax being discarded, especially with the reason given (looking similar to getter calls). The dotted syntax is so common (Cobra, D, Java, C#, Kotlin) that you'd hardly expect anyone to dislike it for that reason.

I'm not discarding it forever, but it does add some real complexity in parsing and plumbing stuff through the compiler and VM. I can still see us adding a dedicated syntax for logical imports at some point, but I think it's a good starting point to do Node-style, which is the simplest solution, and see how that feels.

underscorediscovery:
> We could use that as the trigger to distinguish relative and logical imports

I used it as that temporarily, originally, yea. 
Once I had packages I was using `package: file` as a placeholder, the extension was redundant because `:` was easy to distinguish. 

> working at all 

I feel we've covered a ton of corners and have a way forward!

mhermier:
I rethought about the situation here, and I think we have small issue in the design.
Basically we have: the vm, the language infrastructure, and the CLI.
The CLI should interact with the language infrastructure by providing IO. But currently the signification of include path has to be defined in CLI, instead of language infrastructure.
This means every libwren user has to respect the include syntax in its code. I'm not sure it is a good idea to delegate that.

munificent:
> The CLI should interact with the language infrastructure by providing IO. But currently the signification of include path has to be defined in CLI, instead of language infrastructure.
This means every libwren user has to respect the include syntax in its code. I'm not sure it is a good idea to delegate that.

Yes, that's been on my mind for a long time. Basically, if you want to share Wren code across multiple host applications (CLI, your own personal app, etc.) then the hosts will all need to roughly agree on the same import string policy.

In practice, I think we can solve this by having a good policy in the CLI and other popular hosts and then rely on conventions in the community to rally around that.

PR #480, opened 2017-10-31
Fix embedding path in getting started

aliou:
Hey! This is a PR to fix the embedding path in the getting started page. You might need to then re-publish the docs on the gh-pages branch.

Thanks!


munificent:
Thanks! I'll regen the docs when I get a chance.

PR #481, opened 2017-11-08
Unify error result values.

mhermier:
An attempt at unifiy error values. Instead of defining multiple Error value types, lets define a global one.

### DIFF
--- a/src/cli/vm.c
+++ b/src/cli/vm.c
@@ -153,11 +153,14 @@ static void write(WrenVM* vm, const char* text)
   printf("%s", text);
 }
 
-static void reportError(WrenVM* vm, WrenErrorType type,
+static void reportError(WrenVM* vm, WrenError type,
                         const char* module, int line, const char* message)
 {
   switch (type)
   {
+    case WREN_ERROR_NOERROR:
+      break;
+
     case WREN_ERROR_COMPILE:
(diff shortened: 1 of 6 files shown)


KyleCharters:
WREN_ERROR_NOERROR seems a bit weird to me, Maybe we could base it off of WrenResult instead? Then we would have:

- WREN_RESULT_SUCCESS
- WREN_RESULT_ERR_COMPILE
- WREN_RESULT_ERR_RUNTIME
- WREN_RESULT_ERR_TRACE

mhermier:
NOERR was chosen to try to mirror errno somehow, but it is something that
can be changed/not critical to me.

underscorediscovery:
if using all error prefix,  it also could be `WREN_ERROR_NONE` or `WREN_ERROR_OK`

munificent:
Interesting. I like the simplification, but I worry a little that some amount of expressiveness is lost. With this, there's no clear way to tell that `errorFn()` will never be called with `WREN_ERROR_NOERROR`, that `wrenInterpret()` will never return `WREN_ERROR_STACK_TRACE`, etc.

Given that many C compilers also report an error if a switch on an enum type isn't exhaustive, that ends up pushing users to add cases that will never be matched, which doesn't feel super great to me. Since this only saves a couple of lines, I think I'll keep things the way they are if that's OK.

PR #491, opened 2017-12-08
Reentrancy

mhermier:
The smallest patch set I was able to make (from scratch again).
Some patches might be edited, but this is a good start to discuss.


underscorediscovery:
I'll test this in my codebase in a while!

One of the things I was working on for the patch, was making the usage of the stack consistent as we discussed. I still think that it needs to be that way, i.e slot 0 should always be receiver, 1 = arg 1, so keeping an index into the apistack such that it's always relative to the call frame. This can also be aided by the concepts like negative indices like lua or similar stack slots based ones work.

But I haven't checked the difference here, and will let you know what I see in testing,

mhermier:
Stack alignment is identical to original wren. The twist is that the every
wrenCall is not forced to be at index 0. So that variables can survive
between frame and invocations.
So if you setup stack to be:
foo | bar | 42
And call bla(_) that evaluate to 54 on bar, the resulting stack will be
foo | 54
And this is how you can perform consequent calls. Nothing fancy or
fabulous. Just basic stack manipulation. Another twist is that you can
setup your first frame to hold some objects so they can survive between
invocations.

The only major difference is that you have to perform proper stack
management through wrenSetSlotCount, since wrenCall frame is now determined
by stack top. And therefore wrenEnsureSlot was removed for wrenSetSlotCount
to express that change of constraints.

Thinking at it again there is a window where the garbage collector could
crash the VM, which exist in the current VM that needs to be fixed. When
changing the stack size, new values are not reseted. Which can make the GC
trying to access some deleted object references.

Unfortunately I'm off for the weekend without computer till, so I can't do
any the changes till then. Cheers.

underscorediscovery:
No worries, and yes this description of your stack is what I think is a "problem", I don't think it is ideal, I don't think manipulating the stack outside of your scope is a good idea.

I will let you know how the test works but since the ensureSlots function is  gone, it is no longer trivial, I will have to stub it in to test instead since I have over a hundred foreign endpoints.

mhermier:
Well if you use the current convention without abuse, you should be able to
replace wrenEnsureSlot with wrenSetSlotCount in a no brainier way for
migration.
The only issue will happens if you reserved more slodt than you use in your
wrenCall. And this should be pretty rare.

Stack manipulation should only be necessary when starting to do reentrancy.

So it should be pretty straight forward to do the migration.

jclc:
So, will this PR allow you to call Wren in the middle of calling a foreign function?

mhermier:
In short yes, but there might be some not covered senarios.

mhermier:
Updated patch against master.
Still not perfect 2 tests are failling, but I have an interesting combo of gdb/valgrind failing and distro maintenance. So I postpone the final corrections for now.

mhermier:
Found that the qtcreator was updated/lying to me.
Everything is fixed and working as it was before.

mhermier:
Went down through the rabit hole. The change set is not pushable as this, some parts need rewording and reworking.

But one thing become evident when I tried to make api_call benchmark reentrant: we need another way to control the life time of WrenVM and WrenFiber objects. The fact that WrenVM owns *one current* WrenFiber makes no sense anymore.

So I'll try to add Obj some refcounting unless somebody as a better idea. I considered wrapping WrenVM and WrenFiber in WrenHandle, but while it seems one interesting way to go, it brings one indirection on each API call, and doesn't lower the number of API call needs to operate.

mtneug:
Hi @mhermier. I'm testing out your reentrancy implementation and so far it works quite well; that is, as long as the code execution stays on the "happy path".

I'm still new to Wren and have not a thorough understanding of the inner workings, but I think some of the assumptions no longer hold when using reentrance. For example, consider [this gist](https://gist.github.com/mtneug/26246057216b443cda8ea65ec3113191). Here a foreign functions tries to reenter by executing the non-existing function `doesNotExist()`. Of course, this will lead to a `RuntimeError`, but after leaving the foreign function the program segfaults, which should not happen. This is because the VM sets it's current fiber to `NULL` as part of the error handling. Further, the `try()` call is ignored, as the fiber status is set to `FIBER_ROOT` when `runInterpreter` is executed a second time during the reentrance.

I'm not sure, what the best way to fix this problem would be. Should a `RunntimeError` during a reentrance even cause the fiber to stop? This essentially dictates that the foreign function always returns after an error an no further code should be executed.

mhermier:
It is probably the case. And there quite some political discussion to be
made before that change is accepted, and is an experiment it is not
polished.
That said, the consideration you have about propagating errors is normal.
The way the interpretor works implies that it should be able to transfer to
another fiber. But it has 2 major problems for reantrancy:

What to do with the host/C stack?
How to handle errors?

Basically because of the way the API is made, it hides a basic state
machine that as only on state the fiber. But as soon as you want to
re-enter, you have to be garanteed that when you return from a wrenCall you
end in the same state. And this is not the case because of the fiber
switching implied by the interpreter. I tried to mitigate the issue, but it
seems that it failed.

Now what to do with the C stack... It depends on what you consider wrenCall
should do. Because of the way it is done, and for performance reasons,
error handling is *fake* (by that I mean a fiber die on error and not
recover) so each time you 'try' there is a context switch. For that reason,
I considered that a wrenCall should be transparent and not be transferring
to another fiber. It does imply that when the fiber has an error you have
to return.
From my point of view it is the only sane way to do because there is no
guarantee that you'll end up with the same fiber when returning from a
wrenCall, because of the hidden state.

The other way, creating a new fiber on wrenCall, has other problems. It
implies creating a a fiber, which is not cheap, and also don't garantee
that you return to the same state.

This is also one of the reasons I forked the project. Having VM has the
main object is an *error*. Dealing with fibers/threads that can fork others
is more sane, at the cost of a refcounting them or alike. And I don't
reached that state yet in my fork.

PR #500, opened 2018-01-11
[critical] Fix getNumArguments and wrenBindMethodCode opcodes data.

mhermier:
Some instructions data in these 2 functions are out of syncs, and can try to
patch the code in bad manner.

I took the slower way in wrenBindMethodCode, to ease future maintenance. It is not a critical path, and it decouples patching from instruction walking.

### DIFF
--- a/src/vm/wren_compiler.c
+++ b/src/vm/wren_compiler.c
@@ -2730,7 +2730,6 @@ static int getNumArguments(const uint8_t* bytecode, const Value* constants,
     case CODE_LOAD_FIELD:
     case CODE_STORE_FIELD:
     case CODE_CLASS:
-    case CODE_IMPORT_MODULE:
       return 1;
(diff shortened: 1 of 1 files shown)


Akaricchi:
Could you add a test case?

mhermier:
It is afaik not really possible to test case this. The only way would be to
compare disassembly, or inject specially crafted assembly to trigger the
issue in a controlled manner.
I wrote a private patch that test patching padding against padding made by
the internal disassembler. But quality wise and line of code change
required, it doesn't sounds the way to go.

Akaricchi:
I'm interested in the private patch regardless of the quality -- for an educational experience, if nothing else. I've discovered Wren recently and I've been hacking on it myself the past few days.

munificent:
Thank you!

PR #502, opened 2018-01-18
Fix variable in assert

### LINKED ISSUE #493: Wrong variable for assert?
Shouldn't this ASSERT be ```name``` not ```module```?

https://github.com/munificent/wren/blob/c4d08bcb81045f471ff182d48f0a3de353fc0f46/src/vm/wren_vm.c#L1748

maxdeviant:
Fixes #493.


maxdeviant:
Travis seems to be falling over :cry:

munificent:
Nice catch, thanks!

PR #509, opened 2018-03-13
Fix a documentation error in "Configuring the VM" page

msakuta:
The signature of errorFn function pointer (WrenErrorFn) actually has a WrenVM as the first parameter

I realized that this is duplicate of #484 after posting a pull request.


munificent:
No worries. I went ahead and merged that one in. Thank you for caring about docs!

PR #513, opened 2018-03-19
Fix REPL being broken by latest commit

### LINKED ISSUE #512: System.print("~") breaks
`
> System.print("Hello world~")
Fn does not implement 'try()'.
[repl line 185] in executeInput()
[repl line 74] in handleChar(_)
[repl line 306] in handleChar(_)
[repl line 32] in run()
[repl line 932] in (script)
[main line 1] in (script)
`

This seems to get mis-interpreted?

snocl:
The REPL was broken by commit 2c88e19, as it tried to call `try()` on the result of `Meta.compile`.

I also took the liberty of removing the `_allowAnsi` field, as it seemed like it was never actually used—hopefully I’m not just missing something here!

This should fix #512.


munificent:
Oh, nice catch, thank you!

I really need tests for the REPL...

PR #515, opened 2018-03-21
fix bindForeignMethod() example 

kjk:
add() method is bound as static but bindForeignMethod responds to non-static


munificent:
👍

PR #517, opened 2018-03-29
Unify string

mhermier:
Hi,
This patch set tries to have only one string representation (kill SymbolTable String).
That way we don't need special code to create and use SymbolTable values.
In bonus, it makes metaGetModuleVariables much more simpler, reducing (number of symbol + 1) allocation to 1 allocation.
The first patch is a utility pach that move definitions of all Obj* to wren_common.h to help to reduce the interdependecies of include files. I got doomed a lots of times having to change the order of includes without that, when hacking the code.


munificent:
I love it. I made a few tweaks here and there but otherwise this looks great. :)

Issue #520, opened 2018-04-03
Ignore UTF-8 byte order mark


jclc:
Apparently the Wren compiler doesn't like when the UTF-8 byte order mark is present.
```
$ wren test.wren 
[main line 1] Error: Invalid byte 0xef.
[main line 1] Error: Invalid byte 0xbb.
[main line 1] Error: Invalid byte 0xbf.
```
The Unicode standard mandates that a UTF-8 encoded file may have this byte order mark. Although discouraged by pretty much everyone, Microsoft insists on using it on all UTF-8 encoded text files. If the byte order mark is missing, Notepad, for instance, will try to open it using another encoding.

The Wren compiler should just skip these bytes if they are the first 3 bytes in a file.

bjorn:
Maybe this helps (I had to deal with this problem in another project when loading Lua scripts, cause Lua tripped over this as well):

```c
static char *skipPotentialBom(char *text)
{
    // Based on the C version of bomstrip
    const char * const utf8Bom = "\xef\xbb\xbf";
    const int bomLength = strlen(utf8Bom);
    return (strncmp(text, utf8Bom, bomLength) == 0) ? text + bomLength : text;
}
```

jclc:
I fixed this in wren_compiler.c by adding an offset, but the tests fail since I forked the master branch which is also failing. Not sure if I should make a pull request...

munificent:
Fixed by #525. Thanks, @jclc!

PR #525, opened 2018-04-06
Skip UTF-8 byte order mark

jclc:
#520 Skips the Unicode UTF-8 BOM by adding a 3-byte offset when compiling. Tested by running a few scripts with and without the BOM.

### DIFF
--- a/src/vm/wren_compiler.c
+++ b/src/vm/wren_compiler.c
@@ -3446,12 +3446,11 @@ ObjFn* wrenCompile(WrenVM* vm, ObjModule* module, const char* source,
   parser.printErrors = printErrors;
   parser.hasError = false;
 
-
   // Read the first token.
   nextToken(&parser);
 
   int numExistingVariables = module->variables.count;
-  
+
   Compiler compiler;
   initCompiler(&compiler, &parser, NULL, true);
   ignoreNewlines(&compiler);
--- a/src/vm/wren_compiler.c
+++ b/src/vm/wren_compiler.c
@@ -3416,20 +3416,27 @@ void definition(Compiler* compiler)
 ObjFn* wrenCompile(WrenVM* vm, ObjModule* module, const char* source,
                    bool isExpression, bool printErrors)
 {
+  // Skip potential UTF-8 BOM
+  size_t sourceOffset = 0;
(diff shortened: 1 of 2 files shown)


src/vm/wren_compiler.c
```diff
@@ -3416,20 +3416,27 @@ void definition(Compiler* compiler)
 ObjFn* wrenCompile(WrenVM* vm, ObjModule* module, const char* source,
                    bool isExpression, bool printErrors)
 {
+  // Skip potential UTF-8 BOM
+  size_t sourceOffset = 0;
+  if (source[0] == (char) 0xef && source[1] == (char) 0xbb && source[2] == (char) 0xbf)
```
jvidziunas: This can be crunched down even further to `if (strncmp(source, "\xEF\xBB\xBF", 3) == 0) source += 3;` with the added bonus of not needing to do all the offsets below.

Akaricchi: @jclc that's not a const pointers, that's a pointer to a const. A const pointer would be `char *const source`.

jvidziunas: Bonus aside you can also have turbo-`const` pointers taking the form `const <type>* const` that have both properties.

jclc:
Made the mistake of pushing this to master, now I can't make another PR from another branch without including these commits... Hope this gets merged soon

bjorn:
> Made the mistake of pushing this to master, now I can't make another PR from another branch without including these commits... Hope this gets merged soon

You just need to create the branch for the other PR based on the master branch in this repository rather than your master branch that is ahead.

Suppose you have this repo added as a remote called `upstream`:

    git checkout -b my-new-pr-branch upstream/master

jclc:
I don't think I have write permissions

bjorn:
> I don't think I have write permissions

You don't need write permissions.

jclc:
```
$ git push -u upstream python3-utils 
ERROR: Permission to munificent/wren.git denied to jclc.
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
```
Am I missing something? I'm not very good when it comes to using git.

jclc:
Ah, my bad! I misread and thought @bjorn meant making a new branch in the upstream repository.

munificent:
Sorry for the delay! This is in now. :)

PR #527, opened 2018-04-08
Add optional typing to the language.

mhermier:
A WIP change to add optional typing to wren language.

Rationale: It might be interesting to be able to tag method arguments, their return value and variable definition with type definitions.

Short term: Be able to do weak type definition so user can document in code their types expectations, instead of relying on documentation or variable name tagging. It also helps reasoning when one expect a particular type or interface to be provided by the passed object, instead of needing to dig the code.

Long term: The compiler might evolve or an external compiler might emerge and use this to reason about the code generation.

### DIFF
--- a/src/vm/wren_compiler.c
+++ b/src/vm/wren_compiler.c
@@ -1293,12 +1293,18 @@ static int declareVariable(Compiler* compiler, Token* token)
   return addLocal(compiler, token->start, token->length);
 }
 
+static void optionalTypeDefinition(Compiler* compiler);
+
 // Parses a name token and declares a variable in the current scope with that
 // name. Returns its slot.
 static int declareNamedVariable(Compiler* compiler)
 {
   consume(compiler, TOKEN_NAME, "Expect variable name.");
-  return declareVariable(compiler, NULL);
+  int symbol = declareVariable(compiler, NULL);
+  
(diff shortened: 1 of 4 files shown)


src/vm/wren_compiler.c
```diff
@@ -3364,6 +3381,20 @@ static void import(Compiler* compiler)
...
+static void optionalTypeDefinition(Compiler* compiler)
+{
+  if (match(compiler, TOKEN_COLON))
+  {
+    ignoreNewlines(compiler);
+    parsePrecedence(compiler, PREC_IS);
+    
+    // Discard the type definition.
+    // This compiler is too simple to handle optional type definition.
+    emitOp(compiler, CODE_POP);
```
underscorediscovery: @mhermier if the type definition is entirely ignored, why does it affect the bytecode with a pop here? (I'm not reading too closely!)

mhermier:
Because the compiler automatically push that value on the stack, like in
the case it would care. If we should make the compiler really not care, we
should hack it deeper, so it doesn't generate code on expression by passing
some arguments. I decided that it was not worth the effort since it was a
*one time* evaluation at the creation and also switching back to the case
we're we care would remove the need for the extra changes.

PR #530, opened 2018-04-18
Python 2-3 compatible utility scripts

jclc:
The shebang `#!/bin/env python` is meant to point to python2 but some systems have it point to the newer and preferable python3, making these scripts fail when run without explicitly invoking Python. Since util/generate_docs.py already requires python3, it shouldn't be an issue to migrate all the scripts to the newer version. Removed unused imports and \_\_future\_\_ imports as well.

Building libuv still requires python2 due to its build system, but there is some momentum to migrate it to python3.


jclc:
On second thought, I'll consider making them python2-3 compatible. Although python3 is the future (should've been the present for years now) it's probably better to only require python2 for now rather than both versions.

jclc:
Ironically, the bit where the code actually checked which version of Python was running was causing compatibility issues.

test/language/ignore_carriage_returns.wren
```diff
@@ -1,6 +1,6 @@
...
.
print(
"one"
)
+System.
print(
"two")
+
+// Generate a compile error so we can test that the line number is correct.
++ * // expect error
```
Akaricchi: I think this file has been commited by mistake.

jclc: I believe it used to have carriage returns right before newlines, which confused Python into thinking they were newlines in utils/test.py. Since then I made it open the files as binary to circumvent it, so I suppose it doesn't really make much of a difference.

underscorediscovery:
This seems to have been superseded by #541, closing.

PR #541, opened 2018-05-12
Repaired amalgamation script

kext:
I am not sure if anybody uses the amalgamation, but I love the idea of single file libraries. It makes it so easy to use the code in any project.

The amalgamation script now searches for files in the different directories of src.

Also the full paths in the amalgamation were replaced by the basename to produce the same content on all machines.

The makefile now creates the build directory for the amalgamation if it does not exist. I have added the ‘optional’ folder’s contents to the prerequisites, but the rule is phony anyway, so maybe they should be dropped completely.

I have not tried to run the test suite on the amalgamation, but at least it works with a simple test program. If you want, I can try to add support for amalgamation testing.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -101,7 +101,8 @@ gh-pages: docs
 	$(V) cp -r build/docs/. build/gh-pages
 
 # Build amalgamation of all Wren library files.
-amalgamation: src/include/wren.h src/vm/*.h src/vm/*.c
+amalgamation: src/include/wren.h src/vm/*.h src/vm/*.c src/optional/*.h src/vm/*.c
+	mkdir -p build
 	./util/generate_amalgamation.py > build/wren.c
 
 .PHONY: all amalgamation builtin clean debug docs gh-pages release test vm watchdocs ci ci_32 ci_64
--- a/util/generate_amalgamation.py
+++ b/util/generate_amalgamation.py
@@ -1,7 +1,7 @@
 #!/usr/bin/env python
 
 import sys
-from os.path import basename, dirname, join, realpath
+from os.path import basename, dirname, join, realpath, isfile
 from glob import iglob
 import re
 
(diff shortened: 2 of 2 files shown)


jclc:
Is this Python 2 or Python 3? The shebang `#!/bin/env python` makes it ambiguous as it is system dependent which one it points to. I have another PR #530 for making the basic build and test scripts compatible with either one.

generate_docs.py is still Python 3 only (as it was originally) so if you don't care about 2-3 compatibility it's probably better to just go for Python 3 for non-essential scripts. It's newer and there is absolutely no ambiguity with `#!/bin/env python3`.

kext:
I had originally used this with Python 2. With `python` pointing to Python 3 I could not build wren. But I just replaced the shebang with `python3` and it seems the script works just as well. So it probably runs with both 2 and 3. Can you confirm this? If so, should i replace the shebang with `python3`?

jclc:
Seems to work great with Python 3, so just changing the shebang should suffice :+1:

underscorediscovery:
I tried this out locally on windows and it seemed to work well (along with https://github.com/wren-lang/wren/commit/cdeb0db5c34691c1b715dd3e48e77596dd94fc58). 

Since [python 2 is sunsetting](https://www.python.org/doc/sunset-python-2/) I'm fine with python 3, if there's a discussion to be had about that it can be done in a separate issue.

Thanks for the PR!

mingodad:
Also when trying to compile in Android within termux that comes with python3 it doesn't work.

bjorn:
@mingodad Since this was already merged a while ago, do you mean that the amalgamation still doesn't work on Android with Termux? And if so, what problem is it running into? Just trying to clarify.

mingodad:
Yes that's right !
I'm trying to compile it on an Android 8 device with termux and it doesn't compile due to python scripts missing "()" in print statements (at least that's the error message now).

bjorn:
@mingodad Alright, but that's unrelated to this pull request. Fixing those problems will require updating Wren's copy of Gyp and possibly libuv as well. Probably it's something @underscorediscovery will get around to while he's [separating the command-line interface from the VM for Wren 0.3.0](http://wren.io/blog/0.2.0-and-beyond.html#0.3.0).

underscorediscovery:
Yep that's correct, I am updating the build process to fix the outdated libuv stuff.

PR #555, opened 2018-06-04
Update file.markdown

GHvW:
minor fix to close() language


munificent:
Man that was totally wrong. Thanks!

Issue #553, opened 2018-06-04
How is the "correct" way to debug the Wren when developing it?


macabeus:
When I was studying compilers I found the Wren, and I want to developer it, but I think that is very hard programming in C without a debugger, like LLDB.

I trying many ways to debug the Wren, but I can't find the best way to debug the Wren.

* On lldb, the console isn't type any command:
```
➜  wren git:(master) ✗ lldb wren
(lldb) target create "wren"
Current executable set to 'wren' (x86_64).
(lldb) process launch
Process 41345 launched: '/Users/macabeus/Desktop/wren/wren' (x86_64)
\\/"-
 \_/   wren v0.1.0
> var x = 1Unhandled byte: 10
> var x = 1Unhandled byte: 10
> var x = 1Unhandled byte: 10
```

* I tried many times to use Clion, but I can't attach the executable on debug

* I tried many times to use lldb.nvim plugin, [but I can't work too](https://stackoverflow.com/questions/50671977/error-when-try-to-import-lldb-in-a-python-script)

* On Xcode we have breakpoint, but the console has bug (colors and refresh isn't work).

I "fix" the console on Xcode erasing the function refreshLine and colors, but it is a workaround (#554).

Well... How to debug the Wren? C isn't my main language, sorry if I said something wrong.

mhermier:
I don't think it is a C error, but most probably Xcode failing to emulate an ANSI console, or our test being to much optimistic.
Anyway, I think you should have a working terminal if you replace in https://github.com/munificent/wren/blob/master/src/module/repl.wren#L928:
if (Platform.isPosix && Stdin.isTerminal) {
by
if (false) {
So you get a SimpleRepl.

macabeus:
@mhermier Thank for your reply.
I tried it, but it still didn't works on `refreshLine`. When I try to type `var x = 1` I get on console:

![image](https://user-images.githubusercontent.com/9501115/40951494-a74ca94a-684d-11e8-8b2b-55f0dc8b9c2c.png)

mhermier:
Ok I think I see what is happening. Despite the coloring is disabled, I
think it still try to issue the colored *input help* but since ANSI term is
disabled, it fails somehow at cleaning the line.

mhermier:
Looked at the code and indeed the problem is in refreshLine.
On Mac `\r` is used for newline, while its `\n` on linux and `\r\n` on Windows.
[...]

PR #557, opened 2018-06-08
Fix Xcode console

### LINKED ISSUE #554: Console on Xcode can't work
Colors and refresh can't work very well on Xcode. If you try type something like "var x = 1", you get:

```
\\/"-
 \_/   wren v0.1.0
[2K[30;1m
> [0m
[2Cvar[2K[30;1m
> [0m[0mv[0m
[3C[2K[30;1m
> [0m[0mva[0m
[4C[2K[30;1m
> [0m[36mvar[0m
[5C x[2K[30;1m
> [0m[36mvar[0m[0m [0m
[6C[2K[30;1m
...
```

I "fix" it erasing the function refreshLine and colors, but it is a workaround. What is the correct way to fix it?

macabeus:
Fix Xcode console. Also it adds a new static method on `Process` class: `getEnv()`. I used it to know if the RELP was opened on Xcode or not.

Closes #554 (and a topic in #553)


Akaricchi:
I realize this PR is not really about this, but I think it's still worth pointing out:

>Gets the value of an environment variable. Returns a blank string if the environment variable does not exists.

That's not the right way to do this. An environment variable may exist and be set to an empty string. With this API it's impossible to distinguish it from a non-existent variable. I propose a dict-like object API instead, much like Python's `os.environ`.

mhermier:
Considering how argument passing is done, getEnv should be part of Process and not platform. But if we go a little bit deeper in the rabbit hole, Process should have a method returning an object to the process env. That object would implement a minimal map behavior, and restrict key value pairs to strings only. That way we could have access to the real process env, and if the notion of process is extended, it can return a non system object representing the env.

For the other REPL fix, I'm not sure it is the right approach. Technically I would say that what you wrote is the SimpleRepl.
Not sure how correct it is, but I think refresh should be disabled when `!stdin.isRaw` while the ANSI code should be protected by a variable representing the level of supported ANSI terminal features.
Since I'm pretty sure Xcode implementation seems too dumb to have grabbed a terminal handle, or at least fake it correctly enough, we could disable the isRaw setting by env at construction, in a single consolidated code. Easier said than done, but terminal interaction always implies dirty hands and too many conditionals....

macabeus:
@mhermier In addition of disable the refresh, we still need print the `> ` to have a difference between line typed by user and line of a result.
And, the history still doesn't work, and I really haven't any idea how to fix it (or maybe is better simply disable it).

Since the Xcode console has this differences, I think that is better it has a different class.

macabeus:
@Akaricchi @mhermier I did what you suggested:

- [x] `getEnv` now returns a map with the format `[key: value]`, like Python's `os.environ`
- [x] Moved the `getEnv` static method to Process class
- [x] Disabled history on Xcode console

This method really should called `getEnv`? Or it should be rename it to a better name?
I didn't find any way to returns a map by C API, then I created a Wren function to map the array with environment values to a map.

mhermier:
Considering what the API looks around, and prior art `environ` or `environment` sounds a better name then `getEnv` to me.
You should to do it manually the way compiler does. Creating the map add calling something like `_addCore` (which should really be renamed to `append`)

Akaricchi:
I agree with @mhermier about calling it either `environ` or `environment` (and no parentheses in the signature). Ideally, the returned object shouldn't be a real map with an actual copy of the entire environment. Instead, I'd make it a map-like object with the same API, so that `Process.environ["foo"]` calls `getenv("foo")`, `Process.environ["foo"] = "bar"` calls `setenv("foo", "bar")`, etc. But that warrants a separate pull request at this point.

mhermier:
Agreed with Alexey.

One minor subject that may need a separated discussion, how to handle
software/fake process in general and it's environment in particular, to
stay in the subject of this thread. I would say it should use a plain map,
to be able to pass strings/lists/maps/images/any kind of resources. This
would remove one of the most annoying barrier I have with any shell/program
intercommunication.

macabeus:
@Akaricchi @mhermier I did what you suggested:

- [x] Renamed `getEnv` to `environ`
- [x] No parentheses in the signature

And is better that additional discussions in another pull request.

PR #558, opened 2018-06-10
Add syntax sugar for maps

### LINKED ISSUE #137: Set syntax sugar for maps
The typical way to emulate sets in a language with only maps is by making a map with dummy values. The keys become the set. We can add a little bit of syntax sugar to Wren to make that more natural. In a map literal, you can leave off the `: <value>` and it will implicitly be `true`. This lets you create a "set" like so:

``` dart
var set = {1, 2, 3}
```

By making the dummy value `true` (and since `null` is falsey), it lets you use the subscript operator as a membership test. You can do `containsKey()` too, of course, but this is a bit more terse:

``` dart
if (set[2]) ...
```

macabeus:
Closes #137

This code

```
var set1 = {1, 2, 3}
var set2 = {"foo", "bar": "baz"}
```

now is the same of:

```
var set1 = {1: true, 2: true, 3: true}
var set2 = {"foo": true, "bar": "baz"}
```

### DIFF
--- a/doc/site/maps.markdown
+++ b/doc/site/maps.markdown
@@ -134,3 +134,17 @@ that every entry will appear exactly once.
 
 <a class="right" href="method-calls.html">Method Calls &rarr;</a>
 <a href="lists.html">&larr; Lists</a>
+
+## Set
+
+You can create a set using a syntax sugar of map:
+
+    ::wren
+    var set1 = {1, 2, 3}
+    var set2 = {"foo", "bar": "baz"}
+
+is the same of:
+
+    ::wren
+    var set1 = {1: true, 2: true, 3: true}
+    var set2 = {"foo": true, "bar": "baz"}
--- a/src/vm/wren_compiler.c
+++ b/src/vm/wren_compiler.c
@@ -2023,12 +2023,19 @@ static void map(Compiler* compiler, bool canAssign)
 
     // The key.
     parsePrecedence(compiler, PREC_UNARY);
-    consume(compiler, TOKEN_COLON, "Expect ':' after map key.");
-    ignoreNewlines(compiler);
 
     // The value.
-    expression(compiler);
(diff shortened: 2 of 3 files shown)


mhermier:
I suggest that `var a = { 1, 2, 3}` is equivalent to `var a = { 1: 1, 2: 2,
3: 3}`. Since most values are evalued to true, it should be equivalent to
your suggestion.


Le 10 juin 2018 8:16 PM, "Bruno Macabeus" <notifications@github.com> a
écrit :

@mhermier <https://github.com/mhermier> What do you suggest? I think that
it's a simple and useful implementation to give a set-like feature in Wren.

true is a good choice in order to do something like:

> var x = {1, 2, 3}
> if (x[1]) { System.print("foo") } //> "foo"
> if (x[5]) { System.print("foo") } //> nil

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/munificent/wren/pull/558#issuecomment-396069790>, or mute
the thread

Akaricchi:
But then you can't put `false` in a "set" like this, which is a bit counter-intuitive. What exactly is wrong with `true`?

mhermier:
There is no problems with true, there are advantage of having the key back
has values. When the keys match, there is no guarantee that it is the same
object, so you have to traverse the list of keys to get the key back and
use that instance.

Akaricchi:
https://github.com/munificent/wren/blob/master/src/vm/wren_value.c#L389

Every object that can be hashed is immutable. Why would user code ever need to care whether two immutable objects of equal value are actually the same object or not?

mhermier:
Because this immutability is an implementation detail of the map. It should
not restrict language design.

underscorediscovery:
Thanks for the PR! 
Right now, I don't feel like this fits _just right_ with Wren as is, so I'm gonna close it.

PR #562, opened 2018-06-12
Check that there is space in local variable array

### LINKED ISSUE #561: Array overflow in the locals array by hidden variables of an for loop
There is no overflow guard in addLocal function[1] nor in forStatement function[2]. This causes stack overflow.

[1] https://github.com/munificent/wren/blob/master/src/vm/wren_compiler.c#L1229
[2] https://github.com/munificent/wren/blob/master/src/vm/wren_compiler.c#L2909

PS this seams to be quite serious bug as the Compiler struct is allocated in stack so this could lead to stack smashing attack.

Salmela:
Otherwise it can overflow due to for loop's hidden variables.

I was not sure should I try to patch addLocal function instead. Also I think the error message is too long, but I wanted to make it unique in case that somebody is trying to find where that error message came from. Also I couldn't figure out better way to do the regression test expectation.

Fix #561

### DIFF
--- a/src/vm/wren_compiler.c
+++ b/src/vm/wren_compiler.c
@@ -2906,6 +2906,16 @@ static void forStatement(Compiler* compiler)
   // The space in the variable name ensures it won't collide with a user-defined
   // variable.
   expression(compiler);
+
+  // Verify that there is space to hidden local variables.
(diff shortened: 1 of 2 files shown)


underscorediscovery:
Looks good, thanks!

PR #565, opened 2018-06-17
Add multiple assignment

macabeus:
Add support for write codes like

```
var a, b = [1, 2] // a is 1 and b is 2
```

It's more useful when you have a function with multiples returns, for examples

```
var string = "foo=1"
var key, value = string.split("=") // key is "foo" and value is "1"
```

I influenced myself with [Lua's assignments](https://www.lua.org/pil/4.1.html), then if you want unpack more values than the list length, the variables receives `null`:

```
var a, b, c = [1, 2] // a is 1, b is 2 and c is null
```

You also can unpack less values that the list length:

```
var a, b = [1, 2, 3] // a is 1, b is 2 and 3 is ignored
```

This new feature does not break compatibility, because the behaviour when you want assignment only one variable is the same:

```
var a = [1, 2, 3] // a is [1, 2, 3]
```

You only can do a multiple assignment with lists:

```
var a, b = 1 // raise an error
var a, b = "foo" // raise an error
```

Closes https://github.com/munificent/wren/issues/559

### DIFF
--- a/src/vm/wren_compiler.c
+++ b/src/vm/wren_compiler.c
@@ -2723,6 +2723,7 @@ static int getNumArguments(const uint8_t* bytecode, const Value* constants,
     case CODE_CONSTRUCT:
     case CODE_FOREIGN_CONSTRUCT:
     case CODE_FOREIGN_CLASS:
+    case CODE_UNPACK_LIST:
       return 0;
 
     case CODE_LOAD_LOCAL:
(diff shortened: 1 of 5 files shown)


ruby0x1:
Thanks for the PR! 

While there may be interest in multiple returns at the language level, I don't think this particular syntax or approach would be the right fit for Wren. Specifically, one could consider this a stepping stone toward multiple returns, but the implementation details (like list allocations) wouldn't be ideal in that case. 

That leaves this as just a sugar for unpacking a list into variables, and I don't find the added complexity to the language or compiler to be a good fit at this time. 

Thanks for exploring the option!

PR #566, opened 2018-06-18
add korean translation

sheee7:
Hello, 
I added part of Korean version of language guide, and I followed your guide! 

Have a good day :)


munificent:
Hi. This is really cool!

I would love to have the documentation translated into other languages. 😁 At the same time, I don't want to be the one holding it back, so I think it probably makes the most sense to do this in a separate repo and host it separately. That way you can commit changes as frequently as you need, collaborate with others, etc. I wouldn't really add any value to that since I don't know Korean.

Cheers!

Issue #531, opened 2018-04-21
Failure to build new checkout on gcc 8.0.1, implied buffer overflow


PH0lder:
On Fedora 28 Beta, get a fresh copy and do a make, and get warning errors that prevent building.  It seems to imply some sort of possible buffer overflow.

```
make
make[1]: Entering directory '/home/user/Dev/Wren/wren'
       str src/vm/wren_core.wren          
       str src/optional/wren_opt_meta.wren 
       str src/optional/wren_opt_random.wren 
        cc src/optional/wren_opt_meta.c    -O3 -std=c99 -fPIC
        cc src/optional/wren_opt_random.c  -O3 -std=c99 -fPIC
        cc src/vm/wren_utils.c             -O3 -std=c99 -fPIC
        cc src/vm/wren_compiler.c          -O3 -std=c99 -fPIC
src/vm/wren_compiler.c: In function ‘error’:
src/vm/wren_compiler.c:402:36: error: ‘: ’ directive writing 2 bytes into a region of size between 1 and 139 [-Werror=format-overflow=]
   int length = sprintf(message, "%s: ", label);
                                    ^~
src/vm/wren_compiler.c:402:16: note: ‘sprintf’ output between 3 and 141 bytes into a destination of size 139
   int length = sprintf(message, "%s: ", label);
                ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
cc1: all warnings being treated as errors
```

PH0lder:
I edited line 402 (as noted above) to add +2 to the constant so that I could continue building.  The next error was:

      run util/build_libuv.py            
        cc src/cli/modules.c               -O3 -std=c99 -fPIC
src/cli/modules.c: In function ‘readBuiltInModule’:
src/cli/modules.c:244:3: error: ‘strncpy’ specified bound depends on the length of the source argument [-Werror=stringop-overflow=]
   strncpy(copy, *module->source, length + 1);
   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
src/cli/modules.c:242:19: note: length computed here
   size_t length = strlen(*module->source);
                   ^~~~~~~~~~~~~~~~~~~~~~~
cc1: all warnings being treated as errors

As this is in a codebase upon which wren depends, I shall leave it to someone else to report this issue if relevant.  (It sure seems relevant, as, judging from the description, this could be a potential security issue/buffer overflow.)

munificent:
[...]

PR #568, opened 2018-06-22
Fix compilation on gcc8 and improve memory safety

megous:


bjorn:
The readme says that Wren compiles cleanly as C++98, and `snprintf` was only standardized in C99 and C++11. Maybe Wren should simply only support C99 and C++11, not sure what the reason is behind sticking to C++98.

minirop:
if was already discussed in #531 (and is still C++98-compliant), but I've still yet to find why gcc <= 7 doesn't complain (this warning wasn't implemented before?)

megous:
@minirop There are several new warning categories in gcc 8. Often times it is false positive, because it doesn't catch all precautions that the programmer can make around various "unsafe" functions that makes the call safe.

avivbeeri:
Is there a resolution for this issue? These warnings halt the build for me.

minirop:
> Is there a resolution for this issue? These warnings halt the build for me.

one easy *fix*(?), is to add "+2" for the "message" size: `char message[ERROR_MESSAGE_SIZE + 2];`

avivbeeri:
I've actually been doing this manually, but I was hoping for an upstream
fix :)

PR #579, opened 2018-07-10
Adding List.remove so you can remove elements from a list in an iterator friendly way

avivbeeri:
First time contributing, hope it's okay.

I found myself wanting for a way to remove specific values from a list, during a loop, but there wasn't a simple or syntactically clean way to do so.

I've added a List.remove(value) method so you can remove elements during for loops and such.
It's not amazingly efficient, but I felt it was missing.

Also added some documentation and tests.

(Sorry, I think my editor's auto-whitespace removal kicked in as well)

### DIFF
--- a/src/vm/wren_compiler.c
+++ b/src/vm/wren_compiler.c
@@ -3458,6 +3458,7 @@ ObjFn* wrenCompile(WrenVM* vm, ObjModule* module, const char* source,
   if (isExpression)
   {
     expression(&compiler);
+    consume(&compiler, TOKEN_EOF, "Expect end of expression.");
   }
   else
   {
@@ -3465,7 +3466,7 @@ ObjFn* wrenCompile(WrenVM* vm, ObjModule* module, const char* source,
     {
       definition(&compiler);
       
-      // If there is no newline, it must be the end of the block on the same line.
+      // If there is no newline, it must be the end of file on the same line.
       if (!matchLine(&compiler))
       {
         consume(&compiler, TOKEN_EOF, "Expect end of file.");


avivbeeri:
Oh wait, it doesn't play perfectly with iterators yet. Back to the drawing board.

bjorn:
Apart from the issue with the iterator, I could see the following issues:

* Does it remove only one element or all elements matching that value? In C++, `std::remove` actually removes all elements matching a value, but this implementation removes only one. In Qt, the `QVector` has `removeOne` and `removeAll` methods to avoid the ambiguity.

* When removing only one element, the implementation could re-use the implementation for removing with index.

avivbeeri:
I was aiming for a `removeOne` implementation. I started from the "remove with index" and tried to keep it to a semi-efficient single pass iteration.

You could do `removeAll` using `Sequence.where` I suppose?

bjorn:
> I was aiming for a removeOne implementation. I started from the "remove with index" and tried to keep it to a semi-efficient single pass iteration.

Personally I think the [Sequence API](https://github.com/munificent/wren/blob/40c927f4402bb6ff74fe8aa257bf8042eeff6544/src/vm/wren_core.wren#L7) is missing an `indexOf` function, in which case you'd do this using:

```dart
var index = list.indexOf(value)
if (index) list.removeAt(index)
```

> You could do removeAll using Sequence.where I suppose?

Sure, though that would be a lazy computation and it would not modify the original list, except if you did `seq = seq.where { |e| e != value }.toList`, but now a convenience function would be useful that avoids the reallocation of the entire list.

> My intention was to replicate the functionality of "del" which is used in the pico8 dialect of Lua, which allows for iterator-safe deletion from a table.

I'm not familiar with the pico8 dialect of Lua. When iterating a table to remove elements with `table.remove` in Lua, I generally just iterate it backwards to keep the index valid.

avivbeeri:
I agree with you about the Sequence API being a little lacking, however an indexOf. Is "removeAt" iterator-safe? Otherwise we are in the same position.

Deleting an item in place is certainly better than allocating a whole new list but it works for now. I may try to implement a `Sequence.removeAll` in a bit, if it makes sense as an addition.

bjorn:
> I agree with you about the Sequence API being a little lacking, however an indexOf. Is "removeAt" iterator-safe? Otherwise we are in the same position.

The iterator for a list is just the index, and the implementation of `iteratorValue` just returns the value at that index, as we can see here:

https://github.com/munificent/wren/blob/40c927f4402bb6ff74fe8aa257bf8042eeff6544/src/vm/wren_core.c#L311-L339

So as far as I can see, there is no way to make any insertion or removal iterator-safe, except if you can ensure that it only ever happens _beyond_ the iterator index (or, in case you're changing a location that affects the index, you need to adjust it manually - if you're comfortable with relying on the assumption that the iterator is just the index, or you just avoid the fancy for-loop).

bjorn:
> My intention was to replicate the functionality of "del" which is used in the pico8 dialect of Lua, which allows for iterator-safe deletion from a table.

Btw, I found the documentation for this [`del` function](http://pico-8.wikia.com/wiki/Del) and see nothing about it that would make it iterator-safe. If you iterate a list and use `del` to delete the current index, you will end up skipping a value if you iterate forwards, just like you encountered with your implementation here.

avivbeeri:
Pico8's del function is officially documented here: 
https://www.lexaloffle.com/pico-8.php?page=manual

That wiki is a community effort.

> del() can be called safely on a table's item while iterating over that table.
> 
> 			A={1,10,2,11,3,12}
> 			FOR ITEM IN ALL(A) DO
> 				IF (ITEM < 10) THEN DEL(A, ITEM) END
> 			END
> 			FOREACH(A, PRINT) -- 10,11,12
> 			PRINT(A[3])       -- 12

bjorn:
Hmm, that's interesting. I don't find the documentation very clear in how `ALL` and `DEL` work to enable this feature. I know Qt allows this when using `foreach`, but it is by making an copy-on-write of the container.

avivbeeri:
Indeed, unfortunately it's a closed source product so I can't comment on its implementation, just it's visible behaviour. I think it would be quite convenient though, especially since Wren gives us a foreach syntax for lists.

bjorn:
> I think it would be quite convenient though, especially since Wren gives us a foreach syntax for lists.

I think it's anyway not efficient to remove by value though, at least if you're already iterating the list you want to write the loop such that you can remove by index.

Btw, the example nicely avoids the problem of skipping. Did you actually test whether it works correctly when it would have been `A={1,2,11,3,12}`?

avivbeeri:
I have checked, and you do get the expected output in Pico8.

I concede that it isn't the most efficient operation. It just seemed like a gap in the language design. Iterating a list with a foreach syntax is pretty natural, but because you don't get access to the index, your interactions with the List are somewhat limited. In retrospect, this is probably by design for safety, but the alternatives just feel so cumbersome to write.

Thinking about most basic game loops, you iterate through a set of entities and update them, and sometimes you have to delete them, which involves removing them from the list.

How would you implement that with reasonable performance in Wren?

avivbeeri:
Looking into an in-place `removeIf`, I think Wren would have to support VM re-entrancy in order to manipulate list memory, evaluate a filter function and then act upon the result, unless I've misunderstood how the core works.

bjorn:
> Looking into an in-place removeIf, I think Wren would have to support VM re-entrancy in order to manipulate list memory, evaluate a filter function and then act upon the result, unless I've misunderstood how the core works.

Yeah, this re-entrancy issue has come up repeatedly already and I definitely hope it will be dealt with eventually.

> Thinking about most basic game loops, you iterate through a set of entities and update them, and sometimes you have to delete them, which involves removing them from the list.
> 
> How would you implement that with reasonable performance in Wren?

I guess you could achieve reasonable performance by using two lists, and copying the still valid entities  from one list to the other while going over them, and swapping the lists at the end. If course, that assumes it's alright for the no longer valid entities to stay in the list until all entities have been updated.

PR #584, opened 2018-07-17
Print REPL output without ANSI escapes

### LINKED ISSUE #542: REPL output is invisible on light-colored terminals
This confused me for a good while.

Expected behavior:

```
> var x = 1
> x
1
> System.print(x)
1

>
```

Actual behavior:

```
> var x = 1
> x
 
> System.print(x)
1

>
```

I stared at `repl.wren` for a good while until I noticed that it actually prints the result of evaluating an expression, but in bright white. If your background color is the same, you won't see anything, even if selecting the text (because that swaps foreground and background color which are the same, so no difference).

What's the way forward here? Simply dropping the color escapes for this case to make use of the default foreground color? Introducing a light/dark color scheme? Checking the value type and reusing the built-in associations for picking the right color? Making a function that gives you the printed representation of a value, lexing it and printing it the same way as the input?

wasamasa:
Previously, REPL output was invisible on a terminal emulator with a light background, now no escapes are used, so the default foreground color is applied.

Fixes #542


bjorn:
How hard would it be to make this conditional? It seems other command-line tools manage to do this.

Akaricchi:
You could simply just set the bold attribute instead of also changing the color to white. Then the terminal's foreground color will be used. Example in the shell:
```bash
echo "\033[1mHello\033[0m world"
```
More info here: https://stackoverflow.com/questions/4842424/list-of-ansi-color-escape-sequences

underscorediscovery:
It's also possible to detect terminal features and colour support (it is as always going to be pretty nuanced and require possible system/exec type calls). [A random result from search](https://github.com/piotrmurach/tty-color). I've used similar ones in node land as well.

wasamasa:
@bjorn @underscorediscovery I think you're conflating two things here. Wren's REPL already detects whether it's run in a terminal (as opposed to inside Emacs) and conditionally disables all ANSI colors based on that.  This PR fixes an ergonomics issue, bright white being impossible to see on a light-colored terminal background.

@Akaricchi That sounds like an interesting alternative.

underscorediscovery:
I was assuming based on context, instead of reading deeper. My bad!

wasamasa:
Hm, tests fail, but it seems due to an unrelated issue. I've also taken the liberty to add `Color.bold`, despite that clearly not being a color (then, `Color.none` isn't one either).

PR #585, opened 2018-07-19
wrenHasError

mhermier:
Vectorize wrenHasError code detection to a dedicated function.


IngwiePhoenix:
How is this not merged? :) It's nice having a result from `wrenInterpret(...)`, but there might be a situation where it is simply easier to call `wrenHasError(...)` later on.

mhermier:
I guess because of political reasons.

IngwiePhoenix:
@mhermier How so? I mean, how is this related to a project's politics...? Or rather, to this one in particular.

mhermier:
Well because of the API "wrenHasError" as this occupies a name that should be global, but as it takes a ObjFiber*  the name at least should be replaced. And because of the nature of the API the fact that we care about the evaluation result is a little bit meaning less.

If you look at the API design, you can run an evaluation from C, but you can only know if it does succeed. You can't work with the result of that evaluation, because WrenVM is not guarantee to have the same stack available after the wrenInterpret/wrenCall call. Meaning you can't ask the state to get you back some data. You have to push a callback and handle the results.

For me this invalidate partially the interest of the API. the only way to circumvent the issue, is to push a foreign method to hook the result before the evaluation ends.

Otherwise yes, the real idea behind this patch remains, we need to abstract the error presence test to function, because the formulation is not really trivial when reading it.

munificent:
No political reasons, I just completely dropped the ball on maintaining Wren. :(

PR #589, opened 2018-07-27
Renamed include guard in wren_compiler.h

FICTURE7:
Renamed the include guard in wren_compiler.h from `wren_parser_h` to `wren_compiler_h`.


munificent:
Thanks!

PR #591, opened 2018-08-01
New C API functions: class retrieval, type checking and basic map support

qtnc:
This pull request adds the new following functions in the C API.

I think that these function are of a general use and they surprisingly appeared to be missing. So I propose to add them.

They are especially useful for foreign method implementors.
In the C/C++ side, type checking is an absolute must; not only basic types (Bool/Num/String), but also foreign types, to make sure the scripter calls a foreign method with an object of type X and not Y.
The more the API given to scripters is complex, the more it is important to type check all inputs.

### Class retrieval and comparison

- wrenGetSlotTypeName: get the type name of the object in slot, e.g. Num, Bool, String; useful for debugging or very simple type-checking
- wrenGetSlotClass: Retriev the class of an object into a slot
- wrenGetSlotIsClass, wrenGetSlotIsClassHandle: check if an object is an instance of a given class

### Type checking

- wrenCheckSlotBool: check if a vlue is a Bool
- wrenCheckSlotDouble: check if a value is a Num/double
- wrenCheckSlotString: check if a value is a string
- wrenCheckSlotList: check if a value is a list
- wrenCheckSlotMap: check if a value is a map
- wrenCheckSlotRange: check if a value is a range
- wrenCheckSlotForeign, wrenCheckSlotForeignHandle: check if a value is a foreign object of a given type

### Basic map support

- wrenSetSlotNewMap: store a new empty map
- wrenGetMapValue: retriev the value associated with a key in a map
- wrenPutInMap: associate a value with a key in a map
- wrenRemoveMap: Remove a key/value association from a map
- wrenClearMap: clear a map

### Range support

- wrenSetSlotNewRange: store a new range object
- wrenGetSlotRange: retriev the bounds and inclusivity of a range object

### miscellaneous functions

- wrenIsAborted: check if the current fiber has been aborted because of a runtime error or an explicit wrenAbortFiber call

### DIFF
--- a/src/cli/path.c
+++ b/src/cli/path.c
@@ -75,7 +75,7 @@ static size_t absolutePrefixLength(const char* path)
   // Drive letter.
   if (isDriveLetter(path[0]) && path[1] == ':')
   {
-    if (isSeparator(path->chars[2]))
+    if (isSeparator(path[2]))
     {
       // Fully absolute path.
       return 3;
--- a/src/include/wren.h
+++ b/src/include/wren.h
@@ -263,7 +263,9 @@ typedef enum
   WREN_TYPE_NUM,
   WREN_TYPE_FOREIGN,
   WREN_TYPE_LIST,
+WREN_TYPE_MAP,
   WREN_TYPE_NULL,
+WREN_TYPE_RANGE,
   WREN_TYPE_STRING,
 
   // The object is of a type that isn't accessible by the C API.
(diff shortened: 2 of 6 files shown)


mhermier:
No go for me:
- Move the 5 different parts you sited to different commits. They should be physically separated for readability and ease of understanding.
- Check functions can be vectorised to a single function:
  bool wrenCheckSlotType(WrenVM* vm, int slot, WrenType type);
- Check functions does Validation by asserting. So they should be renamed. Also make a duplicate with wren_primitive.* internal files, maybe they should be merged/unified ?
- Map and Range addition to the public API need more discution, could be done with a wrapper API. Does it bring value to have that here and like that ?

There is probably more, but with that first round of changes, we should be able to see better.

qtnc:
> Move the 5 different parts you sited to different commits. They 
should be physically separated for readability and ease of understanding.

AS far as I know, once it is commited together, pushed and proposed, 
it's too late to go back and split that. Or can you telle me how ?

 > Check functions can be vectorised to a single function:
bool wrenCheckSlotType(WrenVM* vm, int slot, WrenType type);

That's true. I thought it was easier to have different functions, like 
we have different functions to get and set. So I made the same. I can of 
course change it.


 > Check functions does Validation by asserting. So they should be 
renamed. Also make a duplicate with wren_primitive.* internal files, 
maybe they should be merged/unified ?

I completely skipped that file !
Effectively I could perhaps ruse a few of the functions.

However I have still a problem: these functions take an argument name, 
so whether I have also to take a name, i.e.
wrenCheckSlotBool(WrenVM* vm, int slot, const char* argName)
or generate a string "parameter 1" that will generated even if is not 
used, what is a loss of performance 99.9% of the time.

Even if I agree that the error message would be more explicit to the 
final wren scripter,  I don't like taking an argument name because it 
may not always be known. If I have a lot and a complex API to give to 
wren scripters from my embedding C++ application, I won't write the 
wrapper of every method manually (or at least not the simplest ones).

ABout naming, you mean that it would be better wrenAssertSlotXXX rather 
than wrenCheckSlotXXX ?

Thank you.

mhermier:
Le 01/08/2018 à 18:00, qtnc a écrit :
> > Move the 5 different parts you sited to different commits. They
> should be physically separated for readability and ease of understanding.
>
> AS far as I know, once it is commited together, pushed and proposed,
> it's too late to go back and split that. Or can you telle me how ?
Edit your patch set:
`git rebase -i origin/master`
this will allow you to introduce new patch and rework your patch set.
in interactive mode you can reedit patch reorder patches etc ...
You might want to rerun that rebase interactive multiple times to polish 
your patch set.
then
`git push -f`
`-f` to force refreshing the head, github will dynamically update your 
push request to the new head.
>
> > Check functions can be vectorised to a single function:
> bool wrenCheckSlotType(WrenVM* vm, int slot, WrenType type);
>
> That's true. I thought it was easier to have different functions, like
> we have different functions to get and set. So I made the same. I can of
> course change it.
get and set need special arguments dependent of type.
That function is a simple wrapper around wrenGetType, that does some 
validation and has the same signature for all the types...
>
>
> > Check functions does Validation by asserting. So they should be
> renamed. Also make a duplicate with wren_primitive.* internal files,
> maybe they should be merged/unified ?
>
> I completely skipped that file !
> Effectively I could perhaps ruse a few of the functions.
Don't re-use them yet, I think it is a good start to talk again about 
value type validation , since now you expressed a need for user land.
>
> However I have still a problem: these functions take an argument name,
> so whether I have also to take a name, i.e.
> wrenCheckSlotBool(WrenVM* vm, int slot, const char* argName)
> or generate a string "parameter 1" that will generated even if is not
> used, what is a loss of performance 99.9% of the time.
This as to be discussed.
>
> Even if I agree that the error message would be more explicit to the
> final wren scripter, I don't like taking an argument name because it
> may not always be known. If I have a lot and a complex API to give to
> wren scripters from my embedding C++ application, I won't write the
> wrapper of every method manually (or at least not the simplest ones).
Name is the only thing we have to differenciate what is the broken 
argument since wren is untyped.
So how do you differenciate which parameter have the broken type then ?
>
> ABout naming, you mean that it would be better wrenAssertSlotXXX rather
> than wrenCheckSlotXXX ?
wrenValidateSlotXXX, that would follow wren_primitive convention. Though 
it is not strictly a validation, but more a type check. So if unified 
something like wrenValidateSlotType(WrenVm* vm, int slot, WrenType type).
Well for me 'check' is equivalant to 'is', I would expect it to return 
truth and not to assert, but maybe it is only me.

mhermier:
Any news on this? Do you need more help ?

mhermier:
This is nice, have fun. This is what I usually do: hack hack hack commit on
a private branch. When I think I have a mature functionality, I start a
fresh branch and recreate that functionality for a push request. That way
you know what is your aim, and what are the points to pay attention to.

bjorn:
> - C++ implementation of List.sort using std::sort (it allows custom
> comparator callback passing, C qsort doesn't)

The custom comparator is what you pass as in as `int (*comp)(const void *, const void *)` in C qsort.

I really hope you will spend some time to turn some of those improvements into clean pull requests in the future, since for many of them it seems like they would be useful for most of us.

qtnc:
Hello,

 > The custom comparator is what you pass as in as int (*comp)(const 
void *, const void *) in C qsort.

Yes, but I will try to better explain what I mean. I realise it wasn't 
very clear:

Usually (meaning in all modern programming languages), sort come in two 
versions: one taking no argument and based on < or a comparator 
returning -1/0/1, and a version taking such a custom comparator as argument.

The simple version would obviously be:

var list = [ 4, 1, 3, 2, 5 ]
list.sort()
System.print(list) // [1, 2, 3, 4, 5]

But say I want to sort in reverse order. I should be able to write 
something like this:

list.sort{|a,b| a>b }
System.print(list) // [5, 4, 3, 2, 1]

I think that implementing this second version is impossible to do 
cleanly with C qsort.


IN order to do it, in the C comparator function int()(const void*, const 
void*),  I have to keep a global state of what Wren comparator I have to 
call to compare two elements. If the embedding application use two 
distinct WrenVM in two different threads, bad things are going to happen.

C qsort should have been defined as `void qsort (void* elements, int 
count, int size, int(*comparator)(const void* item1, const void* item2, 
void* userData), void* userData)` to allow passing the wren comparator 
to use in some way (probably userData would have been a pointer to a 
struct containing the VM and a call handle).

Some C compilers support closures (as long as they don't escape from 
their enclosing scope), but as far as I know it isn't standard C99.

Obviously I don't want to rewrite myself a version of qsort. So I take 
another option I have: C++ std::sort.

Additionally, using the C++ version allows to base the default version 
on operator <. From Wren point of view it probably looks like much more 
natural.
Otherwise, there should be a method Object.compareTo(_) or a special 
operator <=>. This becomes a language design question.

Java has the method or takes a -1/0/1 comparator. JavaScript takes a 
-1/0/1 comparator, too. PHP 7 took the operator. Python takes a key 
function which must return a natively comparable object. I don't know 
enough ruby to remember what they do but it's probably nothing new.

My personal opinion on this is:
- PHP 7's <=> operator looks very strange and unusual
- The compareTo method implies in cascade that you must implement ==. To 
do things well from there, comparison operators should probably become 
forbidden to override, and this would mean trading some of the beautiful 
simplicity of the language with more special handling code
- Python's way is probably the simplest for final users, but also 
probably the harder to implement and less performant. I see three ways 
to implement it:
1 - The provided key function must always return a number, which is 
quite weird if we want to compare Strings
2 - We have to make three calls for each comparison performed (first 
twice the key function, and then operator <)
3 - We first create a element=>key mapping, we sort on it, and map back 
at the end, but it doesn't let us avoid calling operator < for each 
comparison
None of the three look ideal to me.

qtnc:
Hello,

 > This is the way to go, thought wren API could have help a little. The 
same problems are visible in the libuv binding.

Yes, if we want to stay C99. However it would require writing our own 
version of quick sort (or another sort algorithm).

bjorn:
> IN order to do it, in the C comparator function int()(const void*, const
> void*),  I have to keep a global state of what Wren comparator I have to
> call to compare two elements. If the embedding application use two
> distinct WrenVM in two different threads, bad things are going to happen.

That makes total sense, thanks for elaborating on this problem!

PR #612, opened 2018-10-06
Disable 32 bit macOS Build

iwillspeak:
With the latest XCode it looks like 32 bit builds are deprecated. This
causes build warnings which fail the libuv build. This commit removes
the 32 bit arch from the macOS libuv build, and stops building the 32
build on Travis.


iwillspeak:
Squashed and rebased.

PR #614, opened 2018-10-06
Fix format overflow error, use snprintf/vsnprintf

wasamasa:
I'm on a system with a rather recent GCC (8.2.1) which enables `-Werror=format-overflow` by default. In combination with `-Werror` this makes the build fail for me:

```
src/vm/wren_compiler.c: In function ‘error’:                                       
src/vm/wren_compiler.c:402:36: error: ‘: ’ directive writing 2 bytes into a region o
f size between 1 and 159 [-Werror=format-overflow=]                                
   int length = sprintf(message, "%s: ", label);                                   
                                    ^~                                             
src/vm/wren_compiler.c:402:16: note: ‘sprintf’ output between 3 and 161 bytes into a
 destination of size 159                                                           
   int length = sprintf(message, "%s: ", label);                                   
                ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~                                    
make[1]: *** [util/wren.mk:223: build/release/vm/wren_compiler.o] Error 1
make[1]: Leaving directory '/home/wasa/code/forks/wren'
make: *** [Makefile:16: release] Error 2
```

To fix this I've replaced the `sprintf` call with `snprintf` (and `vsprintf` with `vsnprintf`) and used the actual buffer length for the second argument, this ensures that no overflow can happen. While I was at it, I fixed another use of `sprintf`.

Please let me know if this isn't acceptable (like, because you won't have `snprintf` on Windows) and disabling the warning is the way to go.

### DIFF
--- a/src/vm/wren_compiler.c
+++ b/src/vm/wren_compiler.c
@@ -399,8 +399,8 @@ static void printError(Parser* parser, int line, const char* label,
   
   // Format the label and message.
   char message[ERROR_MESSAGE_SIZE];
-  int length = sprintf(message, "%s: ", label);
-  length += vsprintf(message + length, format, args);
+  int length = snprintf(message, ERROR_MESSAGE_SIZE, "%s: ", label);
+  length += vsnprintf(message + length, ERROR_MESSAGE_SIZE, format, args);
   ASSERT(length < ERROR_MESSAGE_SIZE, "Error should not exceed buffer.");
   
   parser->vm->config.errorFn(parser->vm, WREN_ERROR_COMPILE,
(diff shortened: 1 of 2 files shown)


emekoi:
the first error is because the size of the buffer doesn't account for the semicolon and space in the format string. to fix that `char message[ERROR_MESSAGE_SIZE];` should be changed to `char message[ERROR_MESSAGE_SIZE + 2];`, although, i think `snprintf/vsnprintf` are the way to go.

bjorn:
> The buffer overflow happens before the assertion triggers, at which point it's too late.

It does not matter whether the assert triggers before or after the problem, since it merely functions as a helper during development. It is not meant to actually prevent illegal memory access, only to make sure it does not go unnoticed.

I think this patch is going the wrong path. The rationale for a fixed buffer and not using `snprintf` is written here:

https://github.com/wren-lang/wren/blob/7753dc01ab88c7954f0e8c48923d03dc0edce94d/src/vm/wren_compiler.c#L46-L51

munificent:
I think this this is fixed now by https://github.com/wren-lang/wren/pull/620.

wasamasa:
Indeed, now the build fails me on libuv :D

PR #615, opened 2018-10-08
Fixup Handling of `\` in REPL to Prevent Crash

### LINKED ISSUE #603: Interactive mode crashes when a `\` is typed after a `"`.
When I run wren in interactive mode and start to type out a string literal with a backslash escape sequence, the program immediately crashes. The crash is not very graceful, printing the error without a newline in front, and leaving the terminal text in the colour of the syntax highlighted `"`. 

The error is:
```
Range end out of bounds.
[repl line 555] in text
[repl line 344] in refreshLine(_)
[repl line 28] in run()
[repl line 929] in (script)
[<repl> line 1] in (script)
```
 
My system is GalliumOS, which is based on Xubuntu. I compiled wren per the instructions on the Getting Started page.

EDIT: I get the same problem on Arch Linux, with wren installed from community repo. The line numbers have changed, but this is probably just because they're different versions.

iwillspeak:
When parsing the string token if a `\` was used it would
unconditionally consume the next character. This would led to a token
with an invalid lenght causing a crash when it was printed. This fixes
that by only consuming the token after the `\` if there is one to
consume.

Fixes #603

### DIFF
--- a/src/module/repl.wren
+++ b/src/module/repl.wren
@@ -828,7 +828,7 @@ class Lexer {
 
       if (c == Chars.backslash) {
         // TODO: Process specific escapes and validate them.
-        advance()
+        if (!isAtEnd) advance()
       } else if (c == Chars.percent) {
         // Consume the '('.
         if (!isAtEnd) advance()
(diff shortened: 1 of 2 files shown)


iwillspeak:
bump

underscorediscovery:
Thanks for the PR, looks simple and makes sense. 

_And thanks for the bump post-hand-over :)_

iwillspeak:
No worries. Thought I'd rebase my PRs and bump them to get the ball rolling again now Wren activity is increasing.

PR #618, opened 2018-10-28
Update the example in doc/site/embedding/index.markdown

PrimordialHelios:
I felt the example in the embedding documentation was a bit misleading, and this is an attempt to clarify things a little.


underscorediscovery:
Closing as the user has been deleted, and the changes don't seem to align with the code in master.

PR #619, opened 2018-10-28
Fix doc generation script

PrimordialHelios:
`make docs` was failing for me due to to line 129 of `util/generate_docs.py`:
```python
html = markdown.markdown(contents, ['def_list', 'codehilite', 'smarty'])
```
The only markdown modules I could find had `markdown.markdown` taking two arguments: `text`, and `**kwargs`. The required kwarg in this was `extensions=`.

### DIFF
--- a/util/generate_docs.py
+++ b/util/generate_docs.py
@@ -126,7 +126,7 @@ def format_file(path, skip_up_to_date):
           # visible.
           contents += line
 
-  html = markdown.markdown(contents, ['def_list', 'codehilite', 'smarty'])
+  html = markdown.markdown(contents, extensions=['def_list', 'codehilite', 'smarty'])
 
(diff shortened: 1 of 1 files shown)


iwillspeak:
This seems to be based on your other PR. Might be good to separate it out so that _just_ the documentation generation changes (`generate_docs`) are in this one.

PrimordialHelios:
Oops, you're completely correct, I wasn't even thinking! Should I just `git revert <commit-id>` and then squash the commits?

iwillspeak:
You could do. Or you could reset to master and then cherry-pick the commit.

PrimordialHelios:
Alright, I think I did that right.

Cherry picking is a nice trick, I'll have to remember that one!

PR #620, opened 2018-10-31
Fix overflow on line 401 of wren_compiler.c

PrimordialHelios:
Line 401 of wren_compiler.c allocates space for `ERROR_MESSAGE_SIZE`, but line 402 has `int length = sprintf(message, "%s: ", label);`, which overflows `message` by two characters.

#614 tried to fix this, but I looked through the commit and it doesn't look like it solved the problem, and nobody's touched the PR for almost a month now, so I thought I'd send this one.

### DIFF
--- a/src/vm/wren_compiler.c
+++ b/src/vm/wren_compiler.c
@@ -446,7 +446,7 @@ static void error(Compiler* compiler, const char* format, ...)
   else
   {
     // Make sure we don't exceed the buffer with a very long token.
-    char label[ERROR_MESSAGE_SIZE];
+    char label[10 + MAX_VARIABLE_NAME + 4 + 1];
     if (token->length <= MAX_VARIABLE_NAME)
     {
       sprintf(label, "Error at '%.*s'", token->length, token->start);


bjorn:
I believe the compiler is actually wrong about there being a possibility of buffer overflow. It thinks the buffer can overflow because it assumes `label` could fully use its allocated space, but it really can't. The compiler error can also be suppressed by reducing the size of `label` to the maximum possible size:
```c
char label[10 + MAX_VARIABLE_NAME + 4 + 1];
```
in:

https://github.com/wren-lang/wren/blob/930bc8d5d213b7d627f7444c5e8fab4b2fcb8217/src/vm/wren_compiler.c#L449-L458

PrimordialHelios:
I tried reverting the changes from the commit and changing line 449 like you said, but I get a new error:
```
make[1]: Entering directory '/home/tay/Programming/git/not_mine/wren'
        cc src/vm/wren_compiler.c          -O3 -std=c99 -fPIC
src/vm/wren_compiler.c: In function ‘error’:
src/vm/wren_compiler.c:402:34: error: ‘%s’ directive writing up to 173 bytes into a region of size 159 [-Werror=format-overflow=]
   int length = sprintf(message, "%s: ", label);
                                  ^~
src/vm/wren_compiler.c:458:47:
     printError(compiler->parser, token->line, label, format, args);
                                               ~~~~~
src/vm/wren_compiler.c:402:16: note: ‘sprintf’ output between 3 and 176 bytes into a destination of size 159
   int length = sprintf(message, "%s: ", label);
                ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
I'm certainly not well-versed in C, but it seems to me that increasing the size of `label` just increases the possibility of over-running the `message` buffer.

bjorn:
> I'm certainly not well-versed in C, but it seems to me that increasing the size of `label` just increases the possibility of over-running the `message` buffer.

It would if you increase the size, but actually my change _decreases_ the size of `label` by 80, so I think you did something else.

PrimordialHelios:
You are 100% correct, I misread the fix and wrote `label[10 + ERROR_MESSAGE_SIZE + 4 + 1]`. Reverted the previous fix with your suggestion.

underscorediscovery:
Looking at this fix it seems to be a reasonable fix @PrimordialHelios / @bjorn 

One minor thing though, maybe we can add `ERROR_TOKEN_SIZE` define, similar to the error message size one (we'd need two now) so that it is more consistent. A comment at the top can be added in the same way, to clarify the magic numbers and intent.

underscorediscovery:
I suspect you mean `MAX_VARIABLE_NAME ` instead of `ERROR_MESSAGE_SIZE` again?

PrimordialHelios:
I certainly do, wow, I'm absolutely horrible at copying things today! Sorry about that :P

PrimordialHelios:
*sigh* accidentally pushed a commit from a different branch, my bad.

munificent:
This looks like a clean fix to me. I'm going to go ahead and pull the trigger on it. I hope that's OK, @underscorediscovery. If not, let me know and I'll leave merging PRs up to you. Just trying to lighten your load a little on simple fixes. :)

As far as `-Werror` goes, I think it's generally a win to keep it on in the Makefile. Otherwise, it's too easy for contributors to introduce mistakes that aren't caught. Many users do include Wren by source and use `-Werror` in their own code, so it's important to keep Wren as clean as possible. It occasionally causes false positives like this, but I'd rather find these early and fix them than have them appear later in a user's program when they try to compile the VM into their app.

PrimordialHelios:
@munificent Cool, thanks a bunch! Let me know if any problems pop up that need to be cleaned up.

underscorediscovery:
you're good @munificent, thanks!

PR #621, opened 2018-10-31
Add support for home, end, delete, and ctrl+w

PrimordialHelios:
This PR makes the REPL properly handle the following keys: home (jump to the beginning of the line), end (jump to the end of the line), delete (delete the character to the right of the cursor), and ctrl+w/^w (delete the previous word and all trailing spaces, works just like in a bash shell).

Sorry for all the small PRs, I figured it would be best to separate them since they don't really have anything to do with each other, but I can merge them all into one PR if requested.


munificent:
This is great, thanks!

PrimordialHelios:
Awesome, thanks!

PR #636, opened 2018-11-26
chore(README): point broken links to wren.io

datatypevoid:


underscorediscovery:
Good catch! Thanks for the fixes.

PR #641, opened 2018-12-10
Fix a bunch of thread value and call frames stack errors.

mhermier:
The issues are related to the fact that when calling a primitive of a
foreign, the frame need to be stored before invoking such
foreign/primitives, and not be used directly after invoking them.

The rationale is pretty simple: despite all the efforts, there is no
garantee that any function might not corrupt some cached values.

In the case of a primitive, there the function is at fault here, since
we get the fallowing stack call:
 runInterpreter() -- Has a cache of current frame pointer
 ->prim_fn_call*()
   ->wrenCallFunction()
     ->wrenAppendCallFrame() -- Might grow the frame stack.
Therefore calling STORE_FRAME after the primitive call in
runInterpreter is bogus, since current call frame is possibly dandling
after wrenAppendCallFrame, even if it does correctly report a need to
reload the frame.

In the case of foreign, it is as simple:
  runInterpreter() -- Has a cache of various stack frame pointers
  ->any foreign()
    ->wrenEnsureSlots() -- Might grow the stack.
Since any foreign might call wrenEnsureSlots, there is a chance that the
cached values are also invalid after a wrenEnsureSlot.

It does result in an unbanlanced [STORE/LOAD]_FRAME for performance
reasons. While the LOAD_FRAME could have been moved after the switch,
the switch, generalising the case, it has a huge impact on primitive
performance. Therefore the primitive case should still be left without a
LOAD_FRAME in case it returns true.

### DIFF
--- a/src/vm/wren_vm.c
+++ b/src/vm/wren_vm.c
@@ -988,6 +988,7 @@ static WrenInterpretResult runInterpreter(WrenVM* vm, register ObjFiber* fiber)
         RUNTIME_ERROR();
       }
 
+      STORE_FRAME();
       switch (method->type)
       {
         case METHOD_PRIMITIVE:
@@ -998,7 +999,6 @@ static WrenInterpretResult runInterpreter(WrenVM* vm, register ObjFiber* fiber)
             fiber->stackTop -= numArgs - 1;
           } else {
             // An error, fiber switch, or call frame change occurred.
-            STORE_FRAME();
 
             // If we don't have a fiber to switch to, stop interpreting.
             fiber = vm->fiber;
@@ -1011,10 +1011,10 @@ static WrenInterpretResult runInterpreter(WrenVM* vm, register ObjFiber* fiber)
         case METHOD_FOREIGN:
           callForeign(vm, fiber, method->as.foreign, numArgs);
           if (!IS_NULL(fiber->error)) RUNTIME_ERROR();
+          LOAD_FRAME();
           break;
 
         case METHOD_BLOCK:
-          STORE_FRAME();
           wrenCallFunction(vm, fiber, (ObjClosure*)method->as.closure, numArgs);
           LOAD_FRAME();
           break;


underscorediscovery:
This looks interesting thanks! 
Did you happen to run the benchmarks before and after for interest sake?

mhermier:
Yes, but my base might be old, because of my mix with canary:

api_call - wren                .......... 0.10s 0.0013  98.99% relative to baseline
api_foreign_method - wren      .......... 0.42s 0.0068 104.97% relative to baseline
binary_trees - wren            .......... 0.30s 0.0098 103.23% relative to baseline
binary_trees_gc - wren         .......... 1.41s 0.0205  99.61% relative to baseline
delta_blue - wren              .......... 0.26s 0.0040 102.67% relative to baseline
fib - wren                     .......... 0.27s 0.0041 139.02% relative to baseline
fibers - wren                  .......... 0.05s 0.0014 107.91% relative to baseline
for - wren                     .......... 0.17s 0.0040  98.84% relative to baseline
method_call - wren             .......... 0.21s 0.0123 101.01% relative to baseline
map_numeric - wren             .......... 0.44s 0.0093 106.77% relative to baseline
map_string - wren              .......... 0.15s 0.0033 103.12% relative to baseline
string_equals - wren           .......... 0.27s 0.0039 106.78% relative to baseline

If you can get a recent baseline to compare it with ? But you should get equivalent results.
The STORE_FRAME, is pretty cheap therefore moving before the switch easily enter the cache line.
The only change that could have been costly is calling LOAD_FRAME after a foreign, but it seems to be relatively inexpensive for some reasons, I don't fully understand.

kawa-yoiko:
Hi! Is there any update on this? I've come up with tests (#682) that I hope can be of help.

Personally I don't feel the unpaired store/save semantic 'unfinished' and benchmarks show little difference, thus I would be glad to see this fixed. What do you think?

mhermier:
This change is not enough to test. There should be a class if utility
functions to query the stack size, to force reallocation what ever its
current state.

kawa-yoiko:
Indeed. Do you mean it's better to add C to the Wren-only test and expose such a utility class for it?

mhermier:
These functions should be added separately.
At minimum, there should be a function that returns the size between the
stack frame base and the top of the allocated frame memory.

kawa-yoiko:
Is capacity information intentionally made opaque? (An API change deserves a new issue anyway.)

Perhaps a viable solution now is to use `vm->fiber->stackCapacity - (vm->fiber->stackTop - vm->fiber->stack)` and mark it as a TODO or NOTE. Is this fair?

mhermier:
There is no real functionality besides tuning execution, therefore it is
not required for execution. As such I think Bob did not even though about
it.

Your count is wrong, you should use something like: &stack[stackCapacity] -
stackFrameBase. (Don't have the code around to use the correct variable
names)

mhermier:
That way you know in one function how much you can call
'wrenEnsureSlotCount' before it triggers a reallocation.

kawa-yoiko:
Thanks! Will do that soon.

kawa-yoiko:
Reallocation may result in the same address in the heap with a larger capacity. The following helper function way may be better:

```c
static void growSlotsUntilStackMoved(WrenVM *vm) {
  int slots = 1;
  Value *initialStack = vm->fiber->stack;
  do {
    wrenEnsureSlots(vm, slots++);
  } while (initialStack == vm->fiber->stack);
}
```

I don't think putting a hacky interface (either this one or the one you mentioned) inside `wren.h` fits in with Wren's design, however. I guess the test will have to access these as private APIs — is this a big concern?

mhermier:
No, it is worse because it is a visibility violation. You are using
internals directly to reproduce a user bug. For me this is an absolute no.

kawa-yoiko:
This is not reproducing user bug. The function just emulates normal user behaviour until a certain internal condition is triggered.

Indeed, if white-box tests aren't to be included there, the only valid way is to add the API as you said. But exposing an interface just because of one test seems like another no... :thinking:

mhermier:
Thought there is somehow a user usage, where one can enforce stack
allocation, based on user side (cmd line argument or preallocation based on
user profiling). So there is a real usage for it.

kawa-yoiko:
Pre-allocation is desirable (which `wrenEnsureSlots` already does), but people should not need to deliberately make the stack _address_ change — which sadly is what we actually want to ensure in the test.

Back to the problem, I think there is no need to avoid the white-box approach, since we're testing a hidden internal condition after all :stuck_out_tongue_closed_eyes: Of course the best way is to open up a test suite dedicated to white-box tests, but a NOTE in the comments will hopefully do for now.

What do you think?

mhermier:
Waiting for other people opinion, I think boxing is required.

PR #646, opened 2018-12-19
Fix broken link for error handling

nenofite:
The "runtime error" link on `http://wren.io/embedding/` was broken, so I updated it to point to `http://wren.io/error-handling.html`. I checked all the other links in the Embedding guide to make sure they work too.


underscorediscovery:
Thanks!

Issue #651, opened 2019-01-16
Error message buffer too small


paxos1977:
> cc src/vm/wren_compiler.c          -O3 -std=c99 -fPIC
> src/vm/wren_compiler.c: In function ‘error’:
> src/vm/wren_compiler.c:402:36: error: ‘: ’ directive writing 2 bytes into a region of size between 1 and 159 [-Werror=format-overflow=]
>    int length = sprintf(message, "%s: ", label);
>                                     ^~
> src/vm/wren_compiler.c:402:16: note: ‘sprintf’ output between 3 and 161 bytes into a destination of size 159
>    int length = sprintf(message, "%s: ", label);
>                 ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
> cc1: all warnings being treated as errors
> 

It would appear the message buffer is two bytes too small as it doesn't account for the colon and the space in the `sprintf()` call.

underscorediscovery:
Hi thanks for the effort! See #620 as there has been discussion around this particular bug in several threads :)

paxos1977:
Oh sorry, I did a brief search through the issues and I missed that one.

PR #652, opened 2019-01-16
Extend the error message buffer by 2 bytes

paxos1977:
Extend the error message buffer by 2 bytes to accommodate error message formatting (colon and space). [Issue #651]


bjorn:
See #620 for an alternative fix for this compiler warning.

PR #655, opened 2019-02-09
Add a limited form of re-entrant calls.

### LINKED ISSUE #510: wrenEnsureSlots corruption/crash
I'm running into a weird crash/memory corruption issue where calling I'm calling a class constructor from C++, which calls into Wren, which then calls into a native function. This native function is then calling wrenEnsureSlots with a large enough amount that it triggers a reallocation. After that, Bad Stuff seems to happen, and I either end up with a crash in a random unrelated place, or the VM thinks it has a random negative number amount of slots, and fails.

I don't have a standalone example here, but I can probably try and create one if needed. Here's roughly what I'm doing:

```
	// make sure we can find a new Game class
	wrenEnsureSlots(vm, 1);
	wrenGetVariable(vm, "main", "Game", 0);
	WrenHandle *gameClass = wrenGetSlotHandle(vm, 0);

	if (gameClass == nullptr) {
		trap->Error(ERR_FATAL, "couldn't find Game class");
		return nullptr;
	}

	// instantiate a new Game
	wrenEnsureSlots(vm, 2);
	wrenSetSlotHandle(vm, 0, gameClass);
	wrenSetSlotString(vm, 1, constructorStr);
// wrenGetSlotCount returns 2
	wrenCall(vm, newHnd);
// wrenGetSlotCount returns a random negative number
	wrenReleaseHandle(vm, newHnd);
	wrenReleaseHandle(vm, gameClass);
```

The Game.new constructor then calls into a native function that returns a large amount of data: `var objs = TileMap.objectsInLayer(objectLayer)`

The handler for `objectsInLayer` is a complex function that returns a potential large amount of data, however I've simplified it down to just a wrenEnsureSlots call that will still break things:

```
void wren_map_getobjectsinlayer(WrenVM *vm) {
	wrenEnsureSlots(vm, 20);
}
```

I followed this into wrenEnsureStack, and as long as it returns out at `  if (fiber->stackCapacity >= needed) return;` everything works OK, but as soon as I cross that threshold, I start running into problems.

If I change the initial call to `wrenEnsureSlots(vm, 2);` to a larger number, avoiding the reallocation once I'm inside wrenCall, everything seems to be OK.

munificent:
This doesn't let you arbitrarily call back into the VM from within
foreign methods. I'm still not sure if that's even a good idea since
God knows what that would mean if you switch fibers while doing that.

But this does allow the very important use case of being able to call
a foreign method from within a call to wrenCall(). In other words,
foreign methods need to always be leaf calls on the call stack, but the
root of that stack can now come from runInterpreter() or wrenCall().

Fix #510.

What do you think?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -76,8 +76,7 @@ clean:
 	$(V) rm -rf lib
 
 # Run the tests against the debug build of Wren.
-test: debug
-	$(V) $(MAKE) -f util/wren.mk MODE=debug api_test
+test: api_test debug
 	$(V) ./util/test.py $(suite)
 
 benchmark: release
@@ -91,6 +90,10 @@ benchmark_baseline: release
 unit_test:
 	$(V) $(MAKE) -f util/wren.mk MODE=debug unit_test
 	$(V) ./build/debug/test/unit_wrend
+
+# Build API tests.
+api_test:
+	$(V) $(MAKE) -f util/wren.mk MODE=debug api_test
 
 # Generate the Wren site.
 docs:
@@ -113,4 +116,4 @@ gh-pages: docs
 amalgamation: src/include/wren.h src/vm/*.h src/vm/*.c
 	./util/generate_amalgamation.py > build/wren.c
 
-.PHONY: all amalgamation benchmark builtin clean debug docs gh-pages release test vm watchdocs ci ci_32 ci_64
+.PHONY: all amalgamation api_test benchmark builtin clean debug docs gh-pages release test vm watchdocs ci ci_32 ci_64
--- a/src/vm/wren_vm.c
+++ b/src/vm/wren_vm.c
@@ -268,8 +268,8 @@ static ObjUpvalue* captureUpvalue(WrenVM* vm, ObjFiber* fiber, Value* local)
   return createdUpvalue;
 }
 
-// Closes any open upvates that have been created for stack slots at [last] and
-// above.
+// Closes any open upvalues that have been created for stack slots at [last]
+// and above.
 static void closeUpvalues(ObjFiber* fiber, Value* last)
 {
   while (fiber->openUpvalues != NULL &&
@@ -370,8 +370,7 @@ static void bindMethod(WrenVM* vm, int methodType, int symbol,
 static void callForeign(WrenVM* vm, ObjFiber* fiber,
                         WrenForeignMethodFn foreign, int numArgs)
 {
-  // Save the current state so we can restore it when done.
-  Value* apiStack = vm->apiStack;
+  ASSERT(vm->apiStack == NULL, "Cannot already be in foreign call.");
   vm->apiStack = fiber->stackTop - numArgs;
 
   foreign(vm);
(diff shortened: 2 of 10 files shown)


munificent:
cc @mhermier

underscorediscovery:
This looks interesting @munificent , I'll take a closer look!

mhermier:
I have no proof but I smell something odd, like it would fail at detecting
reentrant call with this patch.

For me, it sounds like a reasonable short lived solution.

But for longer plans, apiStack concerns should be separated and scheduling
moved out of wrenInterpret. Rationale being that:
- apiStack is an optional value of the stack frame base, being present if
in foreign call. These are 2 different concerns, the stack base part being
reusable at other places, and the status *in a foreign context* is simply a
bool.
- wrenInterpret mix execution and scheduling. I found it structuraly weak.
It does not address to this patch directly, but this mix of these 2 in the
same function makes management of stack frame base overly complex to
handle. It may help to deal handling a fiber schedule request from an
reentrant call.

mhermier:
>
> - wrenInterpret mix execution and scheduling. I found it structuraly weak.
> It does not address to this patch directly, but this mix of these 2 in the
> same function makes management of stack frame base overly complex to
> handle. It may help to deal handling a fiber schedule request from an
> reentrant call.
>
I used fiber schedule request here, but I may replace it with execution
interruption management. So that execution returns *interrupt* code, to be
handled by the caller.

>

munificent:
> I have no proof but I smell something odd, like it would fail at detecting reentrant call with this patch.

It should detect re-entrant calls between foreign methods, because `apiStack` will already be non-null when it gets into the second method. It should detect a call to `wrenCall()` from within a foreign method because `vm->fiber->numFrames` will be greater than zero.

There may be some other combinations I missed. Either way, I think this is an improvement over the current code because it does allow one important combination and the code is a little simpler than it was before. Caching the `apiStack` in local variables and restoring it always felt hacky to me.

> apiStack is an optional value of the stack frame base, being present if in foreign call.

Yeah, I spent a little time trying something along these lines a while back. I didn't get it to work, but I might just need to try harder.

> I used fiber schedule request here, but I may replace it with execution interruption management. So that execution returns *interrupt* code, to be handled by the caller.

Interesting. I am definitely concerned about the interaction between fibers and the C stack if we allow arbitrary re-entrant calls. I don't want to push too much of the work of managing fibers onto the host application, but something like this might help.

mhermier:
The biggest concern I have with this code is that at some points the stack
does not represent what it is running.
Debugging wise (at minimum) it means that if reentrancy becomes a real
thing, we will have to push the foreign method on the frame stack. It is
removed because of speed, but the lack of this foreign frame will become a
real issue with reentrancy.

I used fiber schedule request here, but I may replace it with execution
> interruption management. So that execution returns *interrupt* code, to
> be handled by the caller.
>
> Interesting. I am definitely concerned about the interaction between
> fibers and the C stack if we allow arbitrary re-entrant calls. I don't want
> to push too much of the work of managing fibers onto the host application,
> but something like this might help.
>
I have done it in canary. It is a bit tricky, because we have to make
foreign return an enun instead of a bool, and it does expose some abuse of
value validation returned values.

munificent:
Chatted with @underscorediscovery offline and I think he's OK with merging this in.

underscorediscovery:
I forgot to click the review thing here too, thanks @munificent !

PR #658, opened 2019-02-21
start point for embedded

gadget114514:
removing libuv dependency; add new directory to hold source without libuv dependency, stubs and example.
in order to use this, user must implement user event functions in userev.c on each framework.

to build, execute util/makemini.sh


minirop:
libuv is only a dependency of the CLI, not of wren's VM itself.

munificent:
Hi, I think there might be some confusion with how the Wren repository and tech are organized.

We have a single Git repo that contains several things:

* The Wren virtual machine implementation.
* The website for wren.io.
* A host application that lets you run Wren apps and run them from the CLI. In order to write useful Wren code, this needs to talk to the OS. It uses libuv for that.

At some point, it probably makes sense to split these into separate Git repos, but Wren was and is pretty small and it was convenient for me to have them all in one.

Either way, if you just want to use the VM, you are not dependent on libuv whatsoever. It's just used by the CLI wrapper. It might be nice to have both a minimal CLI without libuv and a "maximal" one with it, but I think the reality is that it's unlikely that the Wren maintainers will be able to maintain both of those.

I'm going to close this PR out, but I encourage you to put the CLI in this PR in a separate repo and share it with the world. It's a good fit for users who want a CLI wrapper around Wren without the overhead of libuv. I don't think we're the right people to be responsible for it, but that doesn't mean it isn't useful.

PR #660, opened 2019-02-27
Added String.fromByte with docs and unit test

walterschell:


munificent:
Wow, this is beautifully done. Thank you!

PR #668, opened 2019-03-28
Fix building CLI on Windows

Heaven31415:
I was unable to build Wren CLI on Windows using vs2017. I had to do some changes in order to fix it:

1. I updated libuv dependency in repository from v1.10.0 to v1.13.0, because current version didn't support vs2017.

2. I changed how visual studio is building libuv. Before it was downloading another copy of libuv from github, instead of using that from Wren repository. Now nothing is downloaded, it is directly compiled from files in `deps/libuv`.

3. I'm keeping directory with libuv tests, because when you are compiling it using `vcbuild.bat` you cannot choose which project you want to build and you have to build everything: library, tests and benchmarks. Without tests directory build on Windows won't work.

4. I changed how Windows is detected by preprocessor. It wasn't working for me, so I looked at some more mature cross-platform project and found the proper way to do this. I used this material as a reference: [link](http://nadeausoftware.com/articles/2012/01/c_c_tip_how_use_compiler_predefined_macros_detect_operating_system).

5. I had to change which runtime library we are using, because libuv was using different one and there is no command line option in `vcbuild.bat` which you can use to change that (as far as I can see). This avoids linking both runtime libraries at the same time (and a linker warning).

And some other smaller things which I won't discuss unless someone asks about them.

With those changes you can build Wren CLI on Windows using vs2017. It works both in debug and release mode. There is a small linker warning in debug mode which is connected to another difference between our and libuv options.

I didn't test it on linux and mac, so please do so if you can, I hope I didn't break anything.


ruby0x1:
This PR is no longer needed :) It's appreciated though. 
The Wren CLI repo has a new release and that includes all the build fixes for Windows and all platforms.
https://github.com/wren-lang/wren-cli

PR #674, opened 2019-04-05
implement TODO for checking result of path stat

dalibor91:
In src/cli/vm.c, function isDirectory, we need to check uv_fs_t.result
and print apropriate message if result is < 0


ruby0x1:
Thanks for the PR, the Wren CLI has been separated to https://github.com/wren-lang/wren-cli/ as part of [0.3.0](https://github.com/wren-lang/wren/releases/tag/0.3.0). 

If you're interested, you can submit the PR there for discussion and consideration. Unfortunately I can't transfer PRs, only issues.

PR #677, opened 2019-04-07
Implement getpid and getppid

dalibor91:
Implement getpid and getppid for Process class, to get process id and parent process id. 
Currently it's only working on UNIX based OSes


underscorediscovery:
Wouldn't the lib uv api be the one to ask for portability?

dalibor91:
`uv_os_getpid` (added in 1.18.0), `uv_os_getppid` (added in 1.16.0) are added in latter versions. 

Version that wren is using is 1.10.
I will try to update libuv in separate PR.

dalibor91:
created pull request with updated libuv https://github.com/wren-lang/wren/pull/678 when/if that is approved then I can use libuv here also.

PR #679, opened 2019-04-08
add dockerfile

dalibor91:
Just a Dockerfile that builds image with compiled wren, and when you run it it starts wren console

Build an image
```
docker build . -t wren 
```

start console 
```
docker run --rm -it wren
```

run tests 
```
docker run --rm wren make test 
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,7 @@
+FROM debian:9.8
+
+RUN apt-get update && apt-get install -y make gcc python
+ADD . /wren
+WORKDIR /wren
+RUN make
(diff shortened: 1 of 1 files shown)


ruby0x1:
Thanks for the PR! 

This particular change is no longer valid due to [Wren 0.3.0](https://github.com/wren-lang/wren/releases/tag/0.3.0) changes, but I would also rather recommend adding this information to the [Wiki](https://github.com/wren-lang/wren/wiki), rather than be a part of the repo. It might be good under tools and utilities.

PR #687, opened 2019-05-13
Implement sum method for Sequence class

vladyio:
This PR:

- Implements the `sum` method of the `Sequence` class; includes tests
- Adds documentation for the `sum` method. 

Examples:

```c
System.print([1, 3.14].sum)  //> 4.14
System.print((1..3).sum)  //> 6
System.print([].sum)  //> 0

System.print(["string", 1, 2].sum)  //> runtime error "Can't sum non-numeric values"
```


bjorn:
Note that this is the same as:
```dart
seq.reduce(0) { |sum, item| sum + item }
```
Though it could be useful as a shortcut.

Do we need the explicit runtime error checking, though? Doesn't a reasonable runtime error already happen once it actually tries to add up something non-numeric?

vladyio:
This is my first PR in a project of this scale, so pardon my knowledge gaps in architecture. I have to questions then:

- Do you think using `reduce` in `sum` implementation would be a better thing than a for-loop which I used? 
- As for explicit errors, probably you're right and it is redundant there.

mhermier:
Does not seems very useful, seems to much specific too me. Would be more
interesting in something like NumSequence or NumList.

vladyio:
I'm closing this PR until someone implements either of `NumSequence` or `NumList` then

bjorn:
To be honest I don't really see the point of `NumSequence` or `NumList`. Wren is not Java. I personally see little harm in adding `sum` to `Sequence`, and letter it produce an error when it encounters something that can't be added, the same way it happens when you try to use the `+` operator anywhere else.

We can also add other little helpers like `min`, `max` and `average`. And while browsing the Java 8 Stream API there's more "inspiration" to be found, for example the [`distinct`](https://docs.oracle.com/javase/8/docs/api/java/util/stream/Stream.html#distinct--) function, which could build up a set of seen-values on-demand to prevent processing them again.

In the end it's up to the maintainers to decide. No need to just the PR just cause someone suggests something else. :-)

mhermier:
Personally I see more these utility functions as external ones.
The proposed change was well defined/correct/... but to me sounds to not
fit in the picture. Relying on + and num 0 to perform the first iteration
is not well behaving. Likewise with min/max it relies on *magnitude* aka
ordering operation of items and don't fit such generic container.

underscorediscovery:
hey @vladyio thanks for the PR, it's a good suggestion and a reasonable implementation.
@bjorn is on point with regards to extraneous types like `NumSequence`, as wren is minimal in design and intention, a bunch of top level types like that don't belong.

The idea of `sum`, `min`, `max` and `average` being part of the sequence API is an interesting question. 

In a way, `any` and `reduce` are not that different from `sum`, but there is one important distinction. In the case of the existing API, the utility is asking you to clarify the nature of the sequence through functions (predicate), or providing a utility based on a fundamental with how wren works. e.g `all` with a list of values will return false if any value is null - it doesn't care about the types, only the falseness. 

This is true for `sum` as well. It doesn't _need_ to care about numeric values, it only need to cares about the `+` operator. This means you can do:

```js
var lists = [ [0], [1], [2] ]
var combined = lists.sum // [0,1,2]
var strings = ["hello", "world"] 
var joined = strings.sum //"helloworld"
```

This works for strings, custom + operators, and more. 
This makes sum a generic and useful tool, similar to reduce and others. The use case reasons about usage on the user side. It doesn't make many assumptions except 1. iterable 2. uses + operator (unsurprising for sum).

Now, there is the issue of `var result = 0`, and with what @bjorn suggested, as it moves the assumption deeper into the core that the sequence is numeric. `reduce(0)` and `result = 0` is enforcing that. To avoid that, it should be along the lines of `var result = sequence.first` [which can copy how the reduce function implements it](https://github.com/wren-lang/wren/blob/5ab9ace7562297008b7e35e8eab8dc5c965c4830/src/vm/wren_core.wren#L98-L103). 

**But** since this is essentially just reduce now, it can be implemented like this:

```js
sum { reduce {|current, item| current + item } }
```

So if this is all that sum amounts to, let's question the 'built in' use cases:
- strings - `sequence.join` works here, as does `reduce`
- lists - A: `var result = [] / sequence.each {|list| result+list }`,  B: `sequence.reduce([]) {|cur, list| cur + list }`
- custom - reduce is the only option

What's interesting is that `B:` reduce form is technically a small convenience over the manual accumulator form `A:`. This is preferable, reduce is much cleaner and nicer. Also then, sum is just a minor convenience over reduce, both cleaning up the code and making the sequence API useful. `seq.sum` is much simpler than `seq.reduce() {|cur, item| cur + item }` and is more convenient.

Based on that, with some minor changes, **I do think sum is a useful addition to the sequence API**, and fits alongside the other conveniences provided by sequence already.

`min`, `max`, and `average` are trickier, they don't necessarily need to enforce numericness internally, but that means they're more generic/useful as predicates, something like `seq.min {|a, b| a < b }`. That's a separate discussion and this is already long - so let's focus on sum first.

bjorn:
> ```js
> sum { reduce {|current, item| current + item } }
> ```

Right, I didn't suggest this because this way the sum of an empty sequence is not 0 but produces an error, which is rather unexpected. That said, I do like that it's more generic.

> `min`, `max`, and `average` are trickier, they don't necessarily need to enforce numericness internally, but that means they're more generic/useful as predicates, something like `seq.min {|a, b| a < b }`. That's a separate discussion and this is already long - so let's focus on sum first.

While I think you're certainly right that `average` is trickier, I think `min` and `max` follow the same generic rule of `sum` in that they would map to the following shortcuts:

```dart
min { reduce {|current, item| current < item ? current : item } }
max { reduce {|current, item| item < current ? current : item } }
```

And the `<` operator will determine in a type-specific way exactly what that comparison means. And in this case, it even makes sense for the function to error out on an empty sequence.

Of course, as is usually the case when sorting, it makes sense to add a way to customize the comparison beyond what the `<` operator provides, as you suggested.

underscorediscovery:
> but produces an error

And in this case would be an error about reduce, which is confusing when calling sum. Maybe the empty sequence error similar to reduce can be added for clarity. This line could go at the top of sum, then:

```js
sum { 
  if (isEmpty) Fiber.abort("Can't sum an empty sequence.")
  return reduce {|current, item| current + item }
}
```

@vladyio feel free to make these changes to your PR, once it's aligned closer to the ideal we can consider merging it.

> I think min and max follow the same generic rule of sum in that they would map to the following shortcuts:

Sum leans on multiple types in the core having `+` while only `Num` has `<`. or `>`, given that only has meaning with numbers. What is < of a list without assumptions? It's hard to say...

Though it could be a variant, i.e provide both `min(fn) {}` and `min { }` such that if it's a numeric-like sequence with types that implement < then it applies, and if not the predicate is required.

But I want to still think about this a bit more.

mhermier:
All these, are utility wrappers around reduce. It is easy to add them to
the core language/module but they are not necessary to make it work.

But you can add them to a dedicated module, in a static method. As a bonus,
it would be more reusable since it would/should work with every iterable
and not only with Sequence. As a side note it is also true with a bunch of
methods from Sequence, even reduce itself.

I don't deny the value of such code, but you are augmenting the Sequence
class for the wrong reasons. Look at how these utility functions are
implemented in Java/c++/... They all implement them in a static method or
function. I can try to explain the reasons why, but it is hard to express.

bjorn:
> They all implement them in a static method or function. I can try to explain the reasons why, but it is hard to express.

I see this is the case in the STL, but how is it true for Java?

I think Wren does not lend itself very well for static functions, since they will always need their class prefix. In addition it reverses the order of the code. So it becomes both verbose and hard to read, if such methods are not directly available on the Sequence class.

> Sum leans on multiple types in the core having `+` while only `Num` has `<`. or `>`, given that only has meaning with numbers. What is < of a list without assumptions? It's hard to say...

Hmm, I had thought at least String would also provide this operator. In any case, we don't need to define < for types where it makes no sense. Trying to use these functions will then raise a `X does not implement '<(_)'` error, which is fine.

mhermier:
Le mer. 15 mai 2019 à 11:14, Thorbjørn Lindeijer <notifications@github.com>
a écrit :

> They all implement them in a static method or function. I can try to
> explain the reasons why, but it is hard to express.
>
> I see this is the case in the STL, but how is it true for Java?
>
They was dedicated classes like (java.utils.)ArrayUtils, but since the
introduction of generics one should use (java.util.)Collections.

It doesn't mean class shouldn't offer shorthand to them as methods, but
behavior should be identical to the static ones.

> I think Wren does not lend itself very well for static functions, since
> they will always need their class prefix. In addition it reverses the order
> of the code. So it becomes both verbose and hard to read, if such methods
> are not directly available on the Sequence class.
>
Verbosity is not the biggest problem here. It makes the language more
functional like. It sounds less intuitive, but by doing less, we can reuse
(and compose) more.

> Sum leans on multiple types in the core having + while only Num has <. or
> >, given that only has meaning with numbers. What is < of a list without
> assumptions? It's hard to say...
>
> Hmm, I had thought at least String would also provide this operator. In
> any case, we don't need to define < for types where it makes no sense.
> Trying to use these functions will then raise a X does not implement
> '<(_)' error, which is fine.
>
Well we are so used to the string < operator that we forgot that it is non
sense, but just a useful helper that rely on encoding of the the string.

>

vladyio:
@underscorediscovery, do you think this behaviour is the right thing? 

```
> [].sum
Runtime error: Can't sum an empty sequence.
```

I mean, Python can (`sum([]) # => 0`) and Ruby can (`[].sum # => 0`), and I think it also would make sense in Wren. 

Then also I noticed this:

```
> [null].sum
null

> [null, null].sum
Runtime error: Null does not implement '+(_)'.

> [1, null, null].sum
Runtime error: Right operand must be a number.
```

Which apparently makes sense on the implementation level, but in terms of developer experience doesn't feel obvious and can be confusing. Well, this is just my opinion about your last suggested implementation of `sum`.

mhermier:
This has to be reduce with the accumulator set to 0 to reduce a number of
these errors.

But it does also means it will only works with what Num + accept.

bjorn:
Maybe packing the appending of strings and arrays in a `sum` helper is going too far.

@vladyio Since you're comparing to Python and Ruby, what do they do on non-numeric input?

vladyio:
@bjorn in Ruby you can't really `sum` much except numeric values. 

```ruby
[nil].sum
# => TypeError (nil can't be coerced into Integer)

["str", "ing"].sum
# => TypeError (String can't be coerced into Integer)
```

The same thing happens in Python:

```python
sum([None])
# => TypeError: unsupported operand type(s) for +: 'int' and 'NoneType'

sum(["str", "ing"])
# => TypeError: unsupported operand type(s) for +: 'int' and 'str'
```

Both languages don't support the sum of arrays via the `sum` method, but it's allowed to join arrays by `+` operator. 

I believe they use `sum` only for the numeric sequences since each error points to an integer, which I guess because they both start sums from a zero.

mhermier:
What happens with more exotic numeric values like big numbers, complex, or
polynomials.

vladyio:
@mhermier, Both are able to sum complex and big numbers in arrays. Polynomial isn't included in their standard libraries, so no point in checking them out.

bjorn:
Well, my preference for `sum` would hence go to:
```dart
sum { reduce(0) { |sum, item| sum + item } }
```
Since this seems the most intuitive and most common behavior. I hope we can move on to more interesting work on Wren. :-)

underscorediscovery:
> I mean, Python can (`sum([]) # => 0`) and Ruby can (`[].sum # => 0`), and I think it also would make sense in Wren.

If we are treating sum as a generic sequence operation and not a numeric operation it doesn't matter much contextually what the other languages do, since their parameters are different. Comparing what they do with e.g bignum is also not relevant specifically, as this is Wren, which doesn't have those concepts. What Wren does have is the + operator, which a custom bignum or polynomial would implement, and gain access to sequence sum all the same,

> Then also I noticed this:
> 
> ```
> > [null].sum
> null
> 
> > [null, null].sum
> Runtime error: Null does not implement '+(_)'.
> 
> > [1, null, null].sum
> Runtime error: Right operand must be a number.
> ```
> 
> Which apparently makes sense on the implementation level, but in terms of developer experience doesn't feel obvious and can be confusing. Well, this is just my opinion about your last suggested implementation of `sum`.

All of these are sort of outside of reasonable bounds either way. 

Even if forcing numeric onto the sum implementation, inserting null into your sequence will yield these errors. The input is invalid, regardless of whether it is numeric or generic. I think what you are referring to may be your original `is Num` test? Touching every element in a sequence up front, every time, is a costly operation to pay every time, when the invariant rules are being broken. (This I think was also what bjorn was originally contesting, a high cost every time).

If the issue is invalid inputs then there may be better options that don't incur that cost by default, e.g `sequence.isNumeric` which the user can use to avoid null in their inputs.

vladyio:
Folks, I've updated everything according to what we discussed. 
The docs probably can be improved as well as the tests, but I need your advice on that.

mhermier:
Better, but I'm still not happy with that change, because it doesn't allow
iterables to reuse that code.

bjorn:
I think this version fails to be actually useful. When do you need a generic `sum` method, that can't deal with empty sequences? What use is an alternative to `join`, which can't deal with empty sequences?

I think this is going to be one of those methods where people think they know what it'll do, but after they have tried to use it, they will invariably run into this limitation and need to litter their code with `isEmpty` checks. :-/

One idea I'd have to maybe have everyone happy (but can't please @mhermier ;-)), is to define `sum` as follows:
```dart
sum(start) { reduce(start) { |sum, item| sum + item } }
sum { sum(0) }
```
So now, we have a convenient default behavior for `sum`, but if you need to sum up a custom type, you can provide the starting value to have Wren use your type's + operator.

mhermier:
Le jeu. 16 mai 2019 à 14:33, Thorbjørn Lindeijer <notifications@github.com>
a écrit :

> I think this version fails to be actually useful. When do you need a
> generic sum method, that can't deal with empty sequences? What use is an
> alternative to join, which can't deal with empty sequences?
>
> I think this is going to be one of those methods where people think they
> know what it'll do, but after they have tried to use it, they will
> invariably run into this limitation and need to litter their code with
> isEmpty checks. :-/
>
> One idea I'd have to maybe have everyone happy (but can't please @mhermier
> <https://github.com/mhermier> ;-)), is to define sum as follows:
>
> sum(start) { reduce(start) { |sum, item| sum + item } }
> sum { sum(0) }
>
> Replace start with 'acc' or 'accumulator' to be consistent with 'reduce'
function.

And again I have no problem with the definition, only with where and how it
is defined.

I'm preparing a patch set to show what I mean, because a lot of 'Sequence'
utility methods needs the same treatment. Because 'Sequence' is a concrete
class of the 'Iterable' concept that is used in 'for', we should be able to
have access to most/all of them for 'Iterable' types.
As a bonus since they become static functions we don't need to unit test
them again and again per subclass.

> So now, we have a convenient default behavior for sum, but if you need to
> sum up a custom type, you can provide the starting value to have Wren use
> your type's + operator.
>

mhermier:
This is a preliminary version of the patch set I propose. It is rough, 
but it is more future/generic proof then simply expending 'Sequence'.

In extra, executing 'Iterable.sum([1, 2, 3, 4])' from the interactive 
interpreter seems to trigger some wellknown frame corruption in the VM, 
while the adding the accumulator parameter is not ... really odd.

bjorn:
> Replace start with 'acc' or 'accumulator' to be consistent with 'reduce' function.

Sure, I didn't think too long about that name. :-)

> This is a preliminary version of the patch set I propose. It is rough, but it is more future/generic proof then simply expending 'Sequence'.

I guess you attached a patch? You're replying to a GitHub issue, which sadly does not support attaching files via email.

mhermier:
Seems I'm doomed, because I renamed my local repo to canary. Last time I
checked I was not able to create another repo named wren because of the
rename...

mhermier:
Basically the change is pretty trivial. You take most of the code from
'Sequence' and vectorize it in static method. In my change set, I put
everything in an Iterable class (probably not the best name, but was a
placeholder to show what I meant).

bjorn:
> Seems I'm doomed, because I renamed my local repo to canary. Last time I checked I was not able to create another repo named wren because of the rename...

The name of your Wren fork is kind of irrelevant. You can still push a branch there based on the master in this repository with your suggested changes.

> In my change set, I put everything in an Iterable class

Rather than introducing a new class, couldn't the static methods just be added to Sequence?

mhermier:
> Rather than introducing a new class, couldn't the static methods just be
> added to Sequence?
>
While there is nothing technically against it, I don't really like it.
Since Iterable is mostly a concept/convention/protocol/interface/whatever
in the language, it should not be be mixed with a concrete class, in my
opinion.

>

mhermier:
Here is the flattened/minimal patch so we can discuss, and before I found a better way to contribute. Github is a pain by not allowing multiple forks, and I don't want to mess my canary fork ... so for the time being, here it is:

```
diff --git a/src/vm/wren_core.wren b/src/vm/wren_core.wren
index c2d0af92..613295aa 100644
--- a/src/vm/wren_core.wren
+++ b/src/vm/wren_core.wren
@@ -4,55 +4,106 @@ class Fn {}
 class Null {}
 class Num {}
 
-class Sequence {
-  all(f) {
+class Iterable {
+  static all(iterable, f) {
     var result = true
-    for (element in this) {
+    for (element in iterable) {
       result = f.call(element)
       if (!result) return result
     }
     return result
   }
-
-  any(f) {
+  
+  static any(iterable, f) {
     var result = false
-    for (element in this) {
+    for (element in iterable) {
       result = f.call(element)
       if (result) return result
     }
     return result
   }
-
-  contains(element) {
-    for (item in this) {
+  
+  static contains(iterable, element) {
+    for (item in iterable) {
       if (element == item) return true
     }
     return false
   }
-
-  count {
+  
+  static count(iterable) {
     var result = 0
-    for (element in this) {
+    for (element in iterable) {
       result = result + 1
     }
     return result
   }
-
-  count(f) {
+  
+  static count(iterable, f) {
     var result = 0
-    for (element in this) {
+    for (element in iterable) {
       if (f.call(element)) result = result + 1
     }
     return result
   }
-
-  each(f) {
-    for (element in this) {
+  
+  static each(iterable, f) {
+    for (element in iterable) {
       f.call(element)
     }
   }
+  
+  static isEmpty(iterable) { iterable.iterate(null) ? false : true }
+  
+  static reduce(iterable, acc, f) {
+    for (element in iterable) {
+      acc = f.call(acc, element)
+    }
+    return acc
+  }
+  
+  static reduce(iterable, f) {
+    var iter = iterable.iterate(null)
+    if (!iter) Fiber.abort("Can't reduce an empty sequence.")
+    
+    // Seed with the first element.
+    var result = iterable.iteratorValue(iter)
+    while (iter = iterable.iterate(iter)) {
+      result = f.call(result, iterable.iteratorValue(iter))
+    }
+    return result
+  }
+  
+  static max(iterable) { reduce(iterable) {|accumulator, element| accumulator < element ? element : accumulator } }
+  
+  static min(iterable) { reduce(iterable) {|accumulator, element| accumulator < element ? accumulator : element } }
+  
+  static sum(iterable) { sum(iterable, 0) }
+  
+  static sum(iterable, accumulator) { reduce(iterable, accumulator) {|accumulator, element| accumulator + element } }
+  
+  static toList(iterable) {
+    var result = List.new()
+    for (element in iterable) {
+      result.add(element)
+    }
+    return result
+  }
+}
 
-  isEmpty { iterate(null) ? false : true }
+class Sequence {
+  all(f) { Iterable.all(this, f) }
+
+  any(f) { Iterable.any(this, f) }
+
+  contains(element) { Iterable.contains(this, element) }
+
+  count { Iterable.count(this) }
+
+  count(f) { Iterable.count(this, f) }
+
+  each(f) { Iterable.each(this, f) }
+
+  isEmpty { Iterable.isEmpty(this) }
 
   map(transformation) { MapSequence.new(this, transformation) }
 
@@ -74,25 +125,9 @@ class Sequence {
 
   where(predicate) { WhereSequence.new(this, predicate) }
 
-  reduce(acc, f) {
-    for (element in this) {
-      acc = f.call(acc, element)
-    }
-    return acc
-  }
+  reduce(acc, f) { Iterable.reduce(this, acc, f) }
 
-  reduce(f) {
-    var iter = iterate(null)
-    if (!iter) Fiber.abort("Can't reduce an empty sequence.")
-
-    // Seed with the first element.
-    var result = iteratorValue(iter)
-    while (iter = iterate(iter)) {
-      result = f.call(result, iteratorValue(iter))
-    }
-
-    return result
-  }
+  reduce(f) { Iterable.reduce(this, f) }
 
   join() { join("") }
 
@@ -109,13 +144,7 @@ class Sequence {
     return result
   }
 
-  toList {
-    var result = List.new()
-    for (element in this) {
-      result.add(element)
-    }
-    return result
-  }
+  toList { Iterable.toList(this) }
 }
 
 class MapSequence is Sequence {
```

minirop:
> Here is the flattened/minimal patch so we can discuss, and before I found a better way to contribute. Github is a pain by not allowing multiple forks, and I don't want to mess my canary fork ...

Github might not allow it (which is sad), but you can still do it manually by creating an empty project then add wren as a remote. (you just won't have the "forked from..." message on the repo.

mhermier:
Le 21/05/2019 à 14:48, minirop a écrit :
>
>     Here is the flattened/minimal patch so we can discuss, and before
>     I found a better way to contribute. Github is a pain by not
>     allowing multiple forks, and I don't want to mess my canary fork ...
>
> Github might not allow it (which is sad), but you can still do it 
> manually by creating an empty project then add wren as a remote. (you 
> just won't have the "forked from..." message on the repo.
>
Well there is a *deeper* consequence to this, you can't simply merge 
from the site and I think the site will even forbid me to reuse the name 
(which needs to be confirmed).

bjorn:
> Or you can add a new branch based of `wren/master`.

@mhermier The above suggestion is the way GitHub should work. You can just call it `wren` or `wren/master` in your fork to avoid confusion. Or `wren/[topic]` of course.

Now please let us close this thread. I think @underscorediscovery should make a call regarding whether we will have `sum` and what form (with or without explicit initial 0) it will have for now. It's important we move on from this small contribution and hopefully @vladyio will find more useful changes to contribute.

underscorediscovery:
Yes, I think veering from a minor convenience method into a major refactor of iterable is not the goal here regardless. The discussion was useful but ultimately it is just a minor convenience being discussed, which aligns with the other existing api. I'll take a closer look when I have a gap and we can finalize a choice.

ruby0x1:
image

Unfortunately I can't reopen this PR as it thinks the source repo doesn't exist.

PR #688, opened 2019-06-13
Update path.c

pyway:
may be a memory leak here?


bjorn:
@RobLoach Ha, I didn't know anybody could post reviews, that's pretty neat!

PR #693, opened 2019-07-27
Fix horrendously bad bit hashing function.

munificent:
I was writing a little test script that coincidentally built a map containing a few million integers keys when I discovered the performance tanked. It's from the hash function. hashBits() is used to generate a hash code from the same 64 bits used to represent a Wren number as a double. When building a map containing a large number of integer keys, it's important for this to do a good job scattering the bits across the 32-bit key space.

Alas, it does not. Worse, the benchmark to test this happens to stop just before the performance falls off a cliff, so this was easy to overlook. I updated the map_numeric benchmark to use larger maps to catch this. I can't show before/after numbers because at the larger size, the old version is too slow to finish running. I gave it a few minutes, but it still hadn't completed a single round. Instead, I used this little test script:

```
for (j in 0..10) {
  var map = {}
  var size = 1000000 + j * 100000

  var start = System.clock
  for (i in 1..size) {
    map[i] = i
  }

  var elapsed = System.clock - start
  System.print("size %(size), time %(elapsed)")
}
```

With the previous code, on my Mac laptop this outputs:

```
size 1000000, time 0.141727
size 1100000, time 3.649353
size 1200000, time 23.404542
size 1300000, time 81.003917
size 1400000, time 100.350219
```

Then I gave up and aborted it. The time here should only slightly increase about 10% each step since the maps get 10% bigger each time. Instead, by the time it's 40% larger, the time is almost 1000 *times* slower. Oops.

With this PR, the results are:

```
size 1000000, time 0.189849
size 1100000, time 0.220847
size 1200000, time 0.230764
size 1300000, time 0.244137
size 1400000, time 0.271335
size 1500000, time 0.288265
size 1600000, time 0.378582
size 1700000, time 0.379689
size 1800000, time 0.383916
size 1900000, time 0.399345
size 2000000, time 0.412544
```

Which is exactly what you'd expect. I wanted to understand *why* it was so bad, so I wrote some diagnostic code to get a picture, literally, of what's going on. [This gist](https://gist.github.com/munificent/b4a733baab60ecb3a3cbd275f6e27446) renders a PPM image of a given map showing what's in each bucket.

With the current hash function, on one of example maps above, you get:

before

That's like, *really* bad. All of the entries are clustered into one small region. Those brighter colors inside entries that are far away from their ideal bucket, which means there is a mountain of collisions occurring. Millions of entries are fighting for the same few buckets and spilling over.

Here is what you get after applying this PR:

after

This is basically an ideal image: a pattern-less noise of empty cells and cells containing keys right where they want to be or very close to it.

### DIFF
--- a/src/vm/wren_value.c
+++ b/src/vm/wren_value.c
@@ -356,24 +356,18 @@ ObjMap* wrenNewMap(WrenVM* vm)
   return map;
 }
 
-static inline uint32_t hashBits(DoubleBits bits)
+static inline uint32_t hashBits(uint64_t hash)
 {
-  uint32_t result = bits.bits32[0] ^ bits.bits32[1];
-  
-  // Slosh the bits around some. Due to the way doubles are represented, small
-  // integers will have most of low bits of the double respresentation set to
-  // zero. For example, the above result for 5 is 43d00600.
-  //
-  // We map that to an entry index by masking off the high bits which means
-  // most small integers would all end up in entry zero. That's bad. To avoid
-  // that, push a bunch of the high bits lower down so they affect the lower
-  // bits too.
-  //
-  // The specific mixing function here was pulled from Java's HashMap
-  // implementation.
-  result ^= (result >> 20) ^ (result >> 12);
-  result ^= (result >> 7) ^ (result >> 4);
-  return result;
+  // From v8's ComputeLongHash() which in turn cites:
+  // Thomas Wang, Integer Hash Functions.
+  // http://www.concentric.net/~Ttwang/tech/inthash.htm
+  hash = ~hash + (hash << 18);  // hash = (hash << 18) - hash - 1;
+  hash = hash ^ (hash >> 31);
+  hash = hash * 21;  // hash = (hash + (hash << 2)) + (hash << 4);
+  hash = hash ^ (hash >> 11);
+  hash = hash + (hash << 6);
+  hash = hash ^ (hash >> 22);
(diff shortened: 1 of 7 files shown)


underscorediscovery:
Looks like a good fix! 
I'll give it a try in a big code base tonight,
but I don't see any issues with merging as is.

mhermier:
The proofs showed that the previous hash was broken after a given size, and
that it needed to change.

The only reserve I have is about the signature change from DoubleValue to
uint64. Do you have proofs that it does help?

bjorn:
> The only reserve I have is about the signature change from DoubleValue to uint64. Do you have proofs that it does help?

I think this signature change was just to simplify the code in the `hashValue` function.

munificent:
> The only reserve I have is about the signature change from DoubleValue to
uint64. Do you have proofs that it does help?

I don't, it's just a simplification. Previously, one of the two callsites to `hashBits()` took a 64-bit int, wrapped it in a DoubleBits, only to have `hashBits()` immediately unwrap it. That seemed pointless to me, so I changed the signature to take a raw `uint64_t` instead.

mhermier:
Good enough for me.

As a side note the URL is dead. I wanted to see the mathematical
informations about that hash.

underscorediscovery:
try web.archive.org, I had to quite a way back. [Here's one that's readable.](http://web.archive.org/web/20120728112436/http://www.concentric.net/~Ttwang/tech/inthash.htm)

PR #696, opened 2019-09-05
import module variables without "for"

pyway:
in ./test/language/module/change_imported_value/change_imported_value.wren:

//removed 'for Module, Other'
import "./module"
System.print(Module) // expect: before
// Reassigning the variable in the other module does not affect this one's
// binding.
Other.change
System.print(Module) // expect: before

// But it does change there.
Other.show // expect: after

===result====
before
before
after

note:
    there need to be tested or optimized...


pyway:
the directory for updated files is wrong...corrected...but how clean these comments?

pyway:
I don't suggest Wren to take this pull, so close it now...
but suggestion for compiling modules and checking the imported variables  in compiling stage...

Issue #697, opened 2019-09-15
Calling wrenAbortFiber during foreign class allocation does not abort the fiber


matusnovak:
### Basic working example

I have a C++ class that I am using in Wren. This is the following code I am using to allocate it:

```cpp
void allocateFoo(WrenVM* vm) {
    auto* memory = wrenSetSlotNewForeign(vm, 0, 0, sizeof(Foo));
    new (memory) Foo();
}

...

WrenForeignClassMethods methods;
methods.allocator = allocateFoo;
```

Wren code:

```js
var foo = Foo.new()
```

### The problem

Nothing special so far, but what if **Foo throws an exception** from its **constructor**? I am trying to wrap it in a try/catch statement and then simply using `wrenAbortFiber` as so:

```cpp
void allocateFoo(WrenVM* vm) {
    auto* memory = wrenSetSlotNewForeign(vm, 0, 0, sizeof(Foo));
    try {
        new (memory) Foo();
    } catch (std::exception& e) {
        wrenEnsureSlots(vm, 1);
        wrenSetSlotString(vm, 0, e.what());
        wrenAbortFiber(vm, 0);
    }
}
```

The following try/catch with `wrenAbortFiber` **works when calling a foreign method** from Wren. I will get handled gracefully and I will get the exception message in the `WrenErrorFn` functiom. But it **does not work when allocating a foreign C++ class**.

Instead, I will get:

```
String does not implement 'init new()'
```

I assume Wren is trying to call `init new()` on the the string that should have been handled by wrenAbortFiber.

Am I doing it wrong?

### The bad workaround

As a workaround, when an exception is thrown when calling the C++ constructor, I am setting the slot to null as: `wrenSetSlotNull(vm, 0);`. Then, I am simply using a custom `WrenErrorFn` callback which has a "next error" string. So when an exception is caught, the message will get stored as the "next error" variable, and this will get picked up by the custom `WrenErrorFn` callback. A horrible workaround but it works for me.

mhermier:
What is the Foo class in your wren code?

matusnovak:
@mhermier That's just a random class. It does not have to be anything specific. The problem is with `wrenAbortFiber`.

mhermier:
The problem report that the class don't have a new, so the problem is in
the class

matusnovak:
[...]

PR #698, opened 2019-09-15
Fix wrenAbortFiber does not work inside of foreign class allocator

matusnovak:
This fixes issue #697

### Description

Consider a foreign class that has to abort a fiber in the constructor. An example of this would be a C++ class that throws an exception in its constructor. To do that, one must call `wrenAbortFiber` inside of the allocator function, but at the moment Wren ignores the error. This pull request fixes that problem. More information in issue issue #697 

Quick example:

```
foreign class BadClass {
  construct new() {}
}

error = Fiber.new {
  BadClass.new()
}.try()
System.print(error) // expect: Something went wrong
```

Without this change, the above `System.print` would output: `String does not implement 'init new()'` (because it thinks the fiber error message is the class instance!)

### Changes

* A single change in `src/vm/wren_vm.c` that adds a check if an error has been raised. 
* Added API test for this change.

### DIFF
--- a/src/vm/wren_vm.c
+++ b/src/vm/wren_vm.c
@@ -1198,6 +1198,7 @@ static WrenInterpretResult runInterpreter(WrenVM* vm, register ObjFiber* fiber)
     CASE_CODE(FOREIGN_CONSTRUCT):
       ASSERT(IS_CLASS(stackStart[0]), "'this' should be a class.");
       createForeign(vm, fiber, stackStart);
+      if (wrenHasError(fiber)) RUNTIME_ERROR();
       DISPATCH();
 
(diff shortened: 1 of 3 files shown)


underscorediscovery:
Thanks @matusnovak looks good.

One minor nitpick before merging, please make the change in wren_vm.c consistent [with the rest of the code](https://github.com/wren-lang/wren/blob/08368069c3f1cfba673d4b859a62e31ed1b42887/src/vm/wren_vm.c#L1005-L1012) :)

matusnovak:
@underscorediscovery All done.

underscorediscovery:
nice, thanks @matusnovak

PR #701, opened 2019-09-28
[0.4.0] Implement compound assignments (+= and friends)

ruby0x1:
So I got this working well with fairly low code impact!
Posting a PR as is so @munificent and others can have a look and people can mess with it in the mean time.

~~**todo: add tests.**~~
~~I've got a big change locally that's making that harder for me right now, 
if someone wants to dive into adding tests feel free.~~

**Don't merge this yet**. 
This is a 0.4.0 change.

**Example:**
devenv_M0RaX9Zfpd

**Benchmarks with master as baseline**
image

### DIFF
--- a/src/vm/wren_compiler.c
+++ b/src/vm/wren_compiler.c
@@ -68,6 +68,10 @@ typedef enum
   TOKEN_PERCENT,
   TOKEN_PLUS,
   TOKEN_MINUS,
+  TOKEN_PLUSEQ,
+  TOKEN_MINUSEQ,
+  TOKEN_STAREQ,
+  TOKEN_SLASHEQ,
   TOKEN_LTLT,
   TOKEN_GTGT,
   TOKEN_PIPE,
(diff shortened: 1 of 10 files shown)


src/vm/wren_compiler.c
```diff
@@ -2695,6 +2727,92 @@ void expression(Compiler* compiler)
...
+}
+
+static void setterAssignment(Compiler* compiler, Code instruction, Signature signature)
+{
+  TokenType infixToken = compoundAssignmentStart(compiler);
+  if(infixToken == TOKEN_EQ) return;
+
+  //We're doing self.setter = self.getter + ...
+  //so to call the getter, we duplicate the self on the stack
+  emitOp(compiler, CODE_LOAD_LOCAL_0);
```
underscorediscovery: This fails when 0 is not self (i.e inside a Fn closure). It should basically be stack -1, will revise.

mhermier:
For the record, the implementation doesn't provide all the binary
operators. So as this it is a no go for me.

underscorediscovery:
one step at a time @mhermier.

mhermier:
Please cleanup the change set, it starts to become unreadable.
Introduce the new opcode in a single patch, and rename it. It is not a PUSH, it is part of the DUP familly, something like DUP_CALLFRAME.
Also try to make commits a little bit more atomic, for ease of readability and future maintenance.
Thanks.

bjorn:
> Please cleanup the change set, it starts to become unreadable.

@mhermier I believe those commits are just his WIP. We can look at the entire patch by clicking on "Files changed", which I think is still relatively small. Personally I'd squash it all into a single commit when merging.

mhermier:
From my past experience I prefer to separate pure vm changes from languages
ones at minimum. It makes it easier to bissect errors. Likewise if entries
are more atomic.

mattiasljungstrom:
If you're doing this, how about adding a++, and a-- as well? Should be doable with this new op code? I'm missing a++ much more than a += 4 when writing code in Wren.

ruby0x1:
Going to reopen a new PR soon with a better implementation (one that works, this one doesn't work safely and will mess up the stack).

PR #706, opened 2019-10-14
Support positive sign in scientific notation

Alloyed:
That is to say, numbers like 1.0e+10. This was found as a result of porting a benchmark from another language to Wren.

This case is already handled by strtod(), so we only need to teach the tokenizer that it's ok.


mhermier:
Add sample for + exponent in the documentation.

ruby0x1:
Thanks for the PR!

PR #712, opened 2019-12-02
Fix #710 – Calling wrenEnsureSlots before wrenInterpret causes apiStack to become invalid

### LINKED ISSUE #710: Calling wrenEnsureSlots before wrenInterpret causes apiStack to become invalid
## Description

```c
WrenConfiguration config;
WrenVM *vm = wrenNewVM(&config);

/* let's allocate some slots for later use */
wrenEnsureSlots(vm, 32);

/* then, let's execute a Wren script */
wrenInterpret(vm, "main", myScript);

/* now, however, we face a problem: while the fiber's stack got updated,
   the API stack (aka slot array) didn't, so we're stuck with the old slots that
   got allocated during the first wrenEnsureSlots call */
```
Calling `wrenEnsureSlots` at this point will not yield any effect, because of how slots are ensured:

```c
{
  if (vm->apiStack == NULL) /* ← problematic line */
  {
  }
  
  int currentSize = (int)(vm->fiber->stackTop - vm->apiStack);
  if (currentSize >= numSlots) return;
  
  // Grow the stack if needed.
  int needed = (int)(vm->apiStack - vm->fiber->stack) + numSlots;
  wrenEnsureStack(vm, vm->fiber, needed);
  
  vm->fiber->stackTop = vm->apiStack + numSlots;
}
```
In this case, `vm->apiStack` is not null, but `vm->fiber->stack` is also a completely different stack than `vm->apiStack`. Updating `vm->apiStack` through `wrenSetSlotT` calls will not have any effect, because internally, these functions update `vm->apiStack`, which now points to some unrelated stack.

That by itself wouldn't be a problem, unless we try and `wrenCall` a method:
```c
wrenEnsureSlots(vm, 1);
wrenGetVariable(vm, "main", "MyClass", 0);
/* ↑ will succeed and get us the correct variable */
WrenHandle *class_MyClass = wrenGetSlotHandle(vm, 0);
WrenHandle *method_doThings = wrenMakeCallHandle("doThings()");

wrenSetSlotHandle(vm, 0, class_MyClass);
/* ↑ will trigger a bug, because the old stack is used */
WrenInterpretResult result = wrenCall(vm, method_doThings); /* uh oh! */
/* the above call will fail, because we updated the old stack instead of the new one.
   the VM will try to read from the new stack, and the value read is `null`,
   which causes the call to fail unexpectedly. */
```

## Possible fix
I'm no expert at the internals of the VM, but I feel like a simple `vm->apiSlots` update at the end of `wrenInterpret` would do the job just fine. The overhead would be practically unnoticable, so it wouldn't matter much.

If it's not possible to update `vm->apiSlots` with a correct value, we can at least invalidate it by setting `vm->apiSlots = NULL;`, which would cause subsequent `wrenEnsureSlots` calls to allocate a new stack and use it in any VM invocations. But as I said, I am no expert, so I don't know if it would even work.

liquid600pgm:
This PR fixes #710. 

I am not sure if this causes memory leaks or not. As far as I know, the fiber should somehow be freed, but I can't tell how. Valgrind does report some problems, but they seem unrelated to this issue.

This does not include any tests, because I have no idea how the test system works. If anyone can explain it to me, I'd be more than happy to create a test for this.

### DIFF
--- a/src/vm/wren_vm.c
+++ b/src/vm/wren_vm.c
@@ -1447,7 +1447,9 @@ WrenInterpretResult wrenInterpret(WrenVM* vm, const char* module,
   ObjFiber* fiber = wrenNewFiber(vm, closure);
   wrenPopRoot(vm); // closure.
   
-  return runInterpreter(vm, fiber);
+  WrenInterpretResult result = runInterpreter(vm, fiber);
+  vm->fiber = fiber;
+  return result;
 }
 
 ObjClosure* wrenCompileSource(WrenVM* vm, const char* module, const char* source,


liquid600pgm:
Superseded by #730.

PR #714, opened 2019-12-11
Fixed Random.sample(list, count) for small number of samples

### LINKED ISSUE #713: Random.sample always the same result
```
var RND=Random.new()
var a=[0,1,2,3,4,5,6]
System.print(RND.sample(a))
System.print(RND.sample(a))
System.print(RND.sample(a))
```

Result:
```
0
0
0
```

bjorn:
Closes #713


underscorediscovery:
I haven't looked yet, but are there tests for the random stuff, and if so can we validate this?

bjorn:
> Thanks! I haven't looked yet, but are there tests for the random stuff, and if so can we validate this?

That's true, I should have added the snippet triggering the broken branch as a new test. Maybe @andraaspar would like to do that, otherwise I'll try to get around to it.

bjorn:
> That's true, I should have added the snippet triggering the broken branch as a new test.

Done in pull request #715.

PR #715, opened 2019-12-17
Extended test for Random.sample to cover both branches

bjorn:
I've verified that this test fails without the fix in change 186a8c7c132529b03291d3db1d942abd35bf8316.

See issue #713

### DIFF
--- a/test/random/sample_one.wren
+++ b/test/random/sample_one.wren
@@ -6,15 +6,22 @@ var random = Random.new(12345)
 System.print(random.sample(["single"])) // expect: single
 
 // Should choose all elements with roughly equal probability.
-var list = ["a", "b", "c", "d", "e"]
-var histogram = {"a": 0, "b": 0, "c": 0, "d": 0, "e": 0}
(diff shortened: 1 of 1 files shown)


bjorn:
Hmm, when I wrote this patch, #716 was not merged yet. This test uses two lists, one with size 5 and one with size 7, because previously a different algorithm was used based on the condition `count * count * 5 < list.count`, which is false for the list of 5 elements and true for the list of 7 elements.

But #716 changed the condition to `count * 4 < list.count`, so now both lists are considered long. To make sure we're still testing both branches, we need to reduce one of the lists to have no more than 4 elements.

ruby0x1:
Oh I see, I misremembered what was going on in these and thought it was one I left for after 0.3.0.

ruby0x1:
I reverted this commit for now @bjorn and we can address it in a separate updated PR as needed.

bjorn:
Hmm, there was no reason to revert. One of the lists should just be changed to have only 4 elements instead of 5.

PR #716, opened 2019-12-20
Optimize Random.sample(_, _) for performance

kawa-yoiko:
There is [a surpringly elegant random sampling algorithm](https://dl.acm.org/citation.cfm?id=315746) described in _Programming Pearls_, which outperforms the current implementation on almost every case, especially with medium-sized (say, sqrt(total)) samples.

I have also updated tests to treat samples taken as unordered sets. In rare cases where a randomly taken _and permuted_ list is desired, the shuffle function can always be used.

That aside, Algorithm R can still be used for sampling from an iterator of unknown length, but that probably belongs to another PR.

Below are benchmark results on my machine for reference. Operations are repeated arbitrary times to amplify the differences.

| Total | Samples | Before   | After    |
|:----- |:------- |:-------- |:-------- |
| 6e6   | 4e6     | 1.19467  | 1.172222 |
| 6e6   | 2e6     | 0.991355 | 0.624902 |
| 6e6   | 4000    | 0.727563 | 0.001411 |
| 6e6   | 2000    | 0.725537 | 0.000765 |
| 6e6   | 30      | 1.188042 | 1.13313  |
| 6e6   | 3       | 2.062269 | 2.038365 |
| 600   | 400     | 1.935752 | 1.257749 |
| 600   | 200     | 1.811921 | 0.628874 |
| 600   | 30      | 7.84491  | 0.851716 |
| 600   | 3       | 1.513245 | 1.514514 |

Code:

```
import "random" for Random
var r = Random.new()

var run = Fn.new { |n, m, its|
  var a = List.filled(n, 0)
  var t = System.clock
  for (i in 0...its) r.sample(a, m)
  System.print(System.clock - t)
}

run.call(6e6, 4e6, 1)
run.call(6e6, 2e6, 1)
run.call(6e6, 4e3, 1)
run.call(6e6, 2e3, 1)
run.call(6e6, 30, 1e5)
run.call(6e6, 3, 1e6)
run.call(600, 400, 2e4)
run.call(600, 200, 2e4)
run.call(600, 30, 1e5)
run.call(600, 3, 1e6)
```

Would be glad to hear your opinions!


bjorn:
I've recently extended the tests to cover both existing branches, because one of them was broken before (see #715). Since your new version also implements two approaches, please make sure both are tested.

Please note that sampling a single element is currently also using this function. To me that seems a little much overhead in the name of elegance, so if you're looking into optimizing you could tweak that one as well.

Other then that, I think the array of booleans version is indeed brilliant.

kawa-yoiko:
I have made the test cover all sample sizes possible. A bitmask is used instead of a map to reduce time spent testing.

As for the single element case, I have observed negligible difference (within 2%) so that should not be a concern. However, directly returning `list[int(list.count)]` is significantly faster, so I added that as well.

underscorediscovery:
Thanks for the PR. My only concern/curiosity would be the license from the referenced work. I've stumbled on good books only to later find out they licensed all code/implementations as GPL for example. We'd need to be sure the license is compatible.

kawa-yoiko:
The original journal article predated GPL, the published book's preface mentioned no particular licence, and unlike entire source packages coming with books, it was in the form of an algorithm description (pseudocode), just like the other famous one named after Floyd! So an attribution like that seems enough.

PR #717, opened 2019-12-24
When using wrenGetVariable, it now returns an int to inform you of failure

avivbeeri:
`wrenGetVariable` used to return `void`, and failed an `ASSERT` when the VM was compiled in debug mode.

This isn't helpful for regular applications, when it is useful to know that a variable isn't defined.

Therefore, this breaks the API a little so that `wrenGetVariable` now returns an `int` to indicate success or failure.

It returns a `0` for success and a `-1` for failure.


mhermier:
I would more see it returning something like a wren_errno_t so we could
distinguish errors like no such module, no such variable, invalid argument,
invalid slot... Comments?

avivbeeri:
I did consider that. I just wasn't sure if it was worth including an enum just for this method. I know we have WrenInterpretResult, so I guess it'd be similar. 

But there's also the decision of whether trying to fetch a variable from a non-existent module is an abortable error, vs a runtime issue that the user should handle, so it's up for discussion.

The first pass is really "Is it worth adding to the API?", then the details can be worked out.

mhermier:
The idea behind the scenes would be to unify error values. But the
implementation/idea should start somewhere. I'm waiting for maintainer
comments on this.

avivbeeri:
Added an enum for the result, although i'm not sure I like the name I've used for it.

mhermier:
That last change introduce spurous white space diff. Please edit last
commit to remove white space changes.

avivbeeri:
Oops , my editor is set to remove left over whitespace from the end of lines. I'll fix it when I get a moment.

avivbeeri:
There, should be fixed and clean now :)

src/include/wren.h
```diff
@@ -471,8 +478,8 @@ void wrenGetListElement(WrenVM* vm, int listSlot, int index, int elementSlot);
 void wrenInsertInList(WrenVM* vm, int listSlot, int index, int elementSlot);
 
 // Looks up the top level variable with [name] in resolved [module] and stores
-// it in [slot].
-void wrenGetVariable(WrenVM* vm, const char* module, const char* name,
+// it in [slot]. Returns -1 if the variable couldn't be found.
```
bjorn:
```suggestion
// it in [slot]. The slot is only set when WREN_VARIABLE_SUCCESS is returned.
```

underscorediscovery:
I'll have a closer look at this soon, it makes more sense as a 0.4.0 inclusion so will be post 0.3.0 anyway.

avivbeeri:
I'm hoping for this PR to land in time for 0.4.0. I just spotted a small merge conflict, so I cleaned that up.

vangroan:
Just want to add that I'm hoping for this in the next version too.

ruby0x1:
I don't think a single function with variant returns is the ideal way to answer the use cases we usually have. Like a lot of the wren API, the API makes assertions about your inputs, what is missing is the tools to ensure those invariants on the way in. 

This is now possible with `wrenHasVariable` and `wrenHasModule`, landed in this commit: https://github.com/wren-lang/wren/commit/182ca90b8c9e14a88837699f1fa95f61d19b0f0a 

@vangroan @avivbeeri this should be the stepping stone you're looking for.

vangroan:
@ruby0x1 Yip, that solves my issues. Thanks 👍

PR #719, opened 2020-01-03
Basic file system functions

busterwood:
I wanted to try `wren` out in a real problem but found that I could not create directories.
This PR adds some basic file opertions to the `io` package:

- `Directory.create(path)`, which returns `true` on success and `false` when the folder already exists
- `Directory.delete(path)`
- `File.rename(oldpath, newpath)`


mhermier:
AFAIK this change is problematic, because it is not portable. To achieve
portability, libuv should be updated to provide permissions symbols.

busterwood:
> AFAIK this change is problematic, because it is not portable. To achieve portability, libuv should be updated to provide permissions symbols.

Okay, I have removed the `permissions` parameter from the `Directory.create` method, and the associated `UserPerm`, `GroupPerm` and `OtherPerm` classes.

mhermier:
I checked the libuv API and the problem is that libuv don't provide a mode
flag abstraction. So previous patch is as correct as the simplified version.

It is also to be noted that libuv_fs_mkdir does not use mode flag on
Windows, making its behavior platform dependant.

busterwood:
I think removing `permissions` from `Directory.create` is the right thing to do to keep the API simple.

Later, `File.chmod` and `Directory.chmod` methods can be added to support changing permissions.

mhermier:
The point I discuss here is not about the API form, but about the usage of
S_* constants, and libuv API usage.

bjorn:
Maybe just pass `0777` as mode, avoiding not portable constants and making it simply depend on the current umask on systems where this applies.

ruby0x1:
Thanks for the PR - some notes here:
- Wren CLI is now here, https://github.com/wren-lang/wren-cli/ 
- Unfortunately I can't transfer PRs (only issues). If you'd like to revisit this, please PR on that repo
- We will probably want to address the TODOs around missing return values
- There's some issues with taking a handle from a slot and not releasing it.

Happy to take a look at the PR again if it's reopened on the other repo!

PR #723, opened 2020-01-07
Allow option semicolon at the end of statements

mingodad:
Probably this will not be accepted but it's here anyway.
Also:
- Partial fixes for building libuv with python3.
- Lowering garbage collector threshold to 4MB.

### DIFF
--- a/deps/libuv/build/gyp/pylib/gyp/__init__.py
+++ b/deps/libuv/build/gyp/pylib/gyp/__init__.py
@@ -34,8 +34,8 @@ def DebugOutput(mode, message, *args):
       pass
     if args:
       message %= args
-    print '%s:%s:%d:%s %s' % (mode.upper(), os.path.basename(ctx[0]),
-                              ctx[1], ctx[2], message)
+    print( '%s:%s:%d:%s %s' % (mode.upper(), os.path.basename(ctx[0]),
+                              ctx[1], ctx[2], message) )
 
 def FindBuildFiles():
   extension = '.gyp'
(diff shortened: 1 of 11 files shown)


bjorn:
The fix for Python 3 changes code that is copied from https://chromium.googlesource.com/external/gyp. Such fixes should be made there, but it seems this has already happened. So the action to take here would be to update Wren's copy of gyp.

The reduction from 100 MB to 4 MB seems arbitrary. A meaningful change would be to add a command-line parameter to set this value.

To allow optional semicolons at end of statements just raises the question, why? In Lua they are optional too but they are also needed sometimes to prevent ambiguity. Is the same true for Wren?

Also, each of these changes should get their own pull request.

mingodad:
Helo bjorn !
Thank you for reply !
The main reason to this pull request is to make other people aware of it.
Also I'm doing it thinking that it's not that difficult to do so and it'll allow easy code reuse from other projects (javascript, C, C++, Java, C#).
Cheers !

mhermier:
While there are currently no particular ambiguity to soove, I see a use
case for source formatting and ease of code reading.

The only issue I see is how should it behave for single line auto return
block, should we allow ';' there and only return the final expression?

Unless I'm misstaken consumeIfMatch is equivalent to match and ignoring
return value, so that function seems unnecessary.

mingodad:
Hello mhermier !

Thank you for reply, I didn't noticed the function that you mention, I'll look at it and if it does work I'll review my implementation detail.

Cheers !

underscorediscovery:
Thanks for the PR,
Aside from the PR containing too many unrelated/combined changes, Wren has no interest in optional semicolons at this time.

PR #725, opened 2020-01-17
Functions for operating on Maps from C

avivbeeri:
I found myself wanting to pass structured data from Wren to C, but I didn't want to define a foreign class specifically to do that, so I thought having access to Maps from the C slot API made sense.


mhermier:
This patch is broken/insufficient. It might leak UNDEFINED_VAL and also
does not handle/provide enough error handling.

avivbeeri:
Added something to handle UNDEFINED_VAL.

I suppose the major error case is trying to use an invalid type of object as a map key? Preventing that shouldn't be too difficult, I just need to establish which types are allowed.

avivbeeri:
Although for what it's worth, I can't see anything in the Wren VM which does error checking like that... so does it even enforce what can be used as a key?

bjorn:
> does it even enforce what can be used as a key?

https://github.com/wren-lang/wren/blob/6ab4abe9e3a2767ced01589e9a3d583c840f54c3/src/vm/wren_primitive.c#L50-L59

avivbeeri:
Ah, pardon me. I guess I can call `validateKey` before allowing the value to be set.

avivbeeri:
Seem that function isn't exposed to wren_vm.h. is it okay to import wren_primitive.h just for this function call?

I don't want to duplicate it, in case the criteria for valid keys changes in the future.

avivbeeri:
Added in the key-type validation, which hopefully helps cover the potential error scenarios.

src/vm/wren_vm.c
```diff
@@ -5,6 +5,7 @@
 #include "wren_common.h"
 #include "wren_compiler.h"
 #include "wren_core.h"
+#include "wren_primitive.h"
```
bjorn: Doesn't seem like the right position for this include, assuming the alphabetical order was intensional.

src/vm/wren_value.h
```diff
@@ -76,6 +76,7 @@
 #define IS_FOREIGN(value) (wrenIsObjType(value, OBJ_FOREIGN))   // ObjForeign
 #define IS_INSTANCE(value) (wrenIsObjType(value, OBJ_INSTANCE)) // ObjInstance
 #define IS_LIST(value) (wrenIsObjType(value, OBJ_LIST))         // ObjList
+#define IS_MAP(value) (wrenIsObjType(value, OBJ_MAP))         // ObjMap
```
bjorn: All the comments were so nicely aligned, until this line was added... :-)

src/vm/wren_vm.c
avivbeeri: Oh yes, so it seems. Fixed :)

src/vm/wren_value.h
avivbeeri: Oops, sorry!

src/include/wren.h
```diff
@@ -470,6 +474,14 @@ void wrenGetListElement(WrenVM* vm, int listSlot, int index, int elementSlot);
 // an element, use `-1` for the index.
 void wrenInsertInList(WrenVM* vm, int listSlot, int index, int elementSlot);
 
+// Retrieves a value with the key in [keySlot] from the map in [mapSlot] and stores it in
+// [valueSlot].
+void wrenGetMapValue(WrenVM* vm, int mapSlot, int keySlot, int valueSlot);
+
+// Takes the value stored at [valueSlot] and inserts it into the map stored
+// at [mapSlot] with key [keySlot].
+void wrenInsertInMap(WrenVM* vm, int mapSlot, int keySlot, int valueSlot);
```
bjorn: Opinions may differ, but why not just wrenSetMapValue?

avivbeeri: I was following the example set by the List api, but both are good.

bjorn:
I think it's a nice addition.

src/include/wren.h
avivbeeri: There's a lot of API-naming stuff which is difficult to answer. Whose job is it really to make that kind of decision?

bjorn: We can all make suggestions, but the decision would currently fall upon @underscorediscovery.

mhermier:
The fact that all these functions doesn't return an error code at minimum
frighten me.

avivbeeri:
By "all these functions", do you mean (1) the new ones introduced by this PR, or (2) the whole Wren<->C api?

For (1), the functions I've added here are based on the existing API, which is why they have limited fail-modes.

If your concern is (2), there's a high amount of assumption built into the FFI that you know what you are doing, or are performing your own verification, which I believe is done for performance purposes. Trying to address those shortcoming would be out of scope for this PR.

underscorediscovery:
Yes you're correct @avivbeeri, this PR is consistent with the code base as is and that's reasonable.

src/include/wren.h
underscorediscovery: My impression is that `wrenSetMapValue` is more consistent with `wrenGetMapValue` and less surprising than `insert` when looking for the inverse operation.

avivbeeri: I'll change the function name to match that opinion :)

vladimyr:
@avivbeeri I've accidentally duplicated your effort in #760 🤦 
After @ruby0x1 pointed me here I went through your code and figured you might want to add few things:
```c
/* wren.h */

// Returns the number of entries in the map stored in [slot].
int wrenGetMapCount(WrenVM* vm, int slot);
```
```c
/* wren_vm.c */

int wrenGetMapCount(WrenVM* vm, int slot)
{
  validateApiSlot(vm, slot);
  ASSERT(IS_MAP(vm->apiStack[slot]), "Slot must hold a map.");

  ObjMap* map = AS_MAP(vm->apiStack[slot]);
  return map->count;
}
```

The reason I'm suggesting it because there is already `wrenGetListCount` so we keep parity between List and Map API.

Aside from that, kudos for great work 🏆

test/api/maps.h
```diff
@@ -0,0 +1,5 @@
+#include "wren.h"
+
+WrenForeignMethodFn mapsBindMethod(const char* signature);
+void mapBindClass(
+    const char* className, WrenForeignClassMethods* methods);
```
vladimyr, line 5: You are missing a new line here.

vladimyr:
@avivbeeri I've rebased your work on top of current master and added `wrenGetMapCount`. I had to regenerate projects to include new test sources. You can see result here: https://github.com/wren-lang/wren/compare/master...vladimyr:mapFFI

It would be awesome if you could take a look and update this PR accordingly.

mode777:
Am I missing something or is it not possible to iterate keys of a map with the implemented functionality? This would be important e.g. for serialization. 
We at least need a another method `void wrenGetMapKey(WrenVM* vm, int mapSlot, int index, int keySlot);` right?

avivbeeri:
At the time I wrote this, you were assumed to know which keys you wanted to extract. The side-step solution is to pass `map.keys` to your foreign method along with the map you want to act upon.

ruby0x1:
I had some more functions for iteration coming, I just merged the PR with the foundation :)

mode777:
@ruby0x1 I'm highly interested in this. Can you give me a hint of how the api will look like. My example from above won't work from what I've learned by now.

PR #726, opened 2020-01-29
Added description of sign to the Num class docs

avivbeeri:
Realised there was a `sign` function in the runtime, but not explained in the docs, so I've rectified it.

### DIFF
--- a/doc/site/modules/core/num.markdown
+++ b/doc/site/modules/core/num.markdown
@@ -106,6 +106,10 @@ Rounds the number to the nearest integer.
     System.print((-3.2).round) //> -3
     System.print((-3.7).round) //> -4
 
+### **sign**
+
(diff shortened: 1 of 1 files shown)


MrZeusTheCoder:
Seeing as there have been quite a few docs additions going on right now, maybe we should include this in 0.3.0? It’s pretty small anyway, so I don’t know why we would wait till 0.4.0 to merge it.

ruby0x1:
Primarily history tracking and contribution attribution, I’d like to preserve that where possible!

MrZeusTheCoder:
Ah, ok. I knew I was missing something, 😄.

ruby0x1:
I’ll merge it into 0.3 after that merges into master, it can still be part of the release. Though, the docs are a live form and update after releases anyway so not a big deal.

Issue #722, opened 2020-01-06
Unsupported platform:  FreeBSD


darkoverlordofdata:
Doen't make on bsd:

darko@NomadBSD:~/GitHub/wren] gmake
gmake[1]: Entering directory '/home/darko/GitHub/wren'
        cc src/optional/wren_opt_random.c  -O3 -std=c99 -fPIC
        cc src/optional/wren_opt_meta.c    -O3 -std=c99 -fPIC
        cc src/vm/wren_primitive.c         -O3 -std=c99 -fPIC
        cc src/vm/wren_vm.c                -O3 -std=c99 -fPIC
        cc src/vm/wren_debug.c             -O3 -std=c99 -fPIC
        cc src/vm/wren_value.c             -O3 -std=c99 -fPIC
        cc src/vm/wren_core.c              -O3 -std=c99 -fPIC
        cc src/vm/wren_compiler.c          -O3 -std=c99 -fPIC
        cc src/vm/wren_utils.c             -O3 -std=c99 -fPIC
        cc lib/libwren.so                  -O3 -std=c99 -fPIC -Wl,-soname,libwren.so
        ar lib/libwren.a                  rcu
       run util/build_libuv.py            
Unsupported platform: FreeBSD
gmake[1]: *** [util/wren.mk:242: build/libuv.a] Error 1
gmake[1]: Leaving directory '/home/darko/GitHub/wren'
gmake: *** [Makefile:16: release] Error 2

darkoverlordofdata:
I was able to fix the build - see https://github.com/darkoverlordofdata/wren

Do you want a pull?

maxdeviant:
@darkoverlordofdata I don't see a reason why a fix wouldn't be accepted. Never hurts to make a PR 😄

darkoverlordofdata:
Ok, I sent it. Thanks!

PR #727, opened 2020-01-30
Unsupported platform: FreeBSD #722

darkoverlordofdata:
It's pretty simple. For BSD, you need to use gmake for the gnu makefile.
Thanks.

### DIFF
--- a/util/wren.mk
+++ b/util/wren.mk
@@ -116,6 +116,8 @@ else
 	# Link in the right libraries needed by libuv on Windows and Linux.
 	ifeq ($(OS),mingw32)
 		LIBUV_LIBS := -lws2_32 -liphlpapi -lpsapi -luserenv
+	else ifeq ($(OS),freebsd12.0)
+		LIBUV_LIBS := -lpthread -lrt -lkvm
 	else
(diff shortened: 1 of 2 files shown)


util/wren.mk
```diff
@@ -116,6 +116,8 @@ else
 	# Link in the right libraries needed by libuv on Windows and Linux.
 	ifeq ($(OS),mingw32)
 		LIBUV_LIBS := -lws2_32 -liphlpapi -lpsapi -luserenv
+	else ifeq ($(OS),freebsd12.0)
```
underscorediscovery: is the `12.0` here very specific? wouldn't it limit other versions of freebsd (I'm unfamiliar with the os semantics, sorry!)?

underscorediscovery:
@darkoverlordofdata here's a semi related question, but you might be able to help answer it. Can you easily run premake on freebsd? I've seen it on freshports and other places so it seems like it's not a stretch. I'd be curious if you can (without much effort) build and run the latest premake 5 alpha there.

https://premake.github.io/download.html#v5

ruby0x1:
This was resolved in 0.3.0 with a bsd specific make file, see projects/make.bsd

PR #729, opened 2020-02-03
Fix duplicate symbol compile error in amalgamation build.

gw3583:


gw3583:
Let me know if you'd prefer to fix this another way, e.g. renaming those functions but keeping them file-local.

mhermier:
Don't know what the maintainer will say, but I would say to keep the
function in the compiler. The rationale is that it an language detail.

avivbeeri:
There's a fairly tight coupling between the compiler and the VM though, and it doesn't make much sense to duplicate the functionality, especially if it's important both methods are kept in sync.

mhermier:
This is not really a tight coupling, it is an implementation choice of the
language with that VM.
Anyway, it concerns a grammatical rule of the language, it should stick
next to the language definitions.

underscorediscovery:
Thanks @gw3583 I think the fixes make sense here. I'm going to merge this into master once I bring the 0.3.0 branch over. Part of the goal 0.3.0 is fixing all the build processes and this is one of those fixes, so thanks!

ruby0x1:
The fix went into the 0.3.0 branch and was merged in. I appreciate the contribution even though this one wasn't merged!

PR #730, opened 2020-02-10
Fix slot array corrupted by wrenInterpret()

### LINKED ISSUE #710: Calling wrenEnsureSlots before wrenInterpret causes apiStack to become invalid

kawa-yoiko:
Fixes #710 (discussions happen there) and add a test case. Credit largely goes to @liquid600pgm for providing detailed analysis.

### DIFF
--- a/src/vm/wren_vm.c
+++ b/src/vm/wren_vm.c
@@ -1446,8 +1446,12 @@ WrenInterpretResult wrenInterpret(WrenVM* vm, const char* module,
   wrenPushRoot(vm, (Obj*)closure);
   ObjFiber* fiber = wrenNewFiber(vm, closure);
   wrenPopRoot(vm); // closure.
-  
-  return runInterpreter(vm, fiber);
+
+  WrenInterpretResult result = runInterpreter(vm, fiber);
+
+  vm->fiber = NULL;
+  vm->apiStack = NULL;
+  return result;
 }
 
 ObjClosure* wrenCompileSource(WrenVM* vm, const char* module, const char* source,
--- a/test/api/new_vm.c
+++ b/test/api/new_vm.c
@@ -13,9 +13,43 @@ static void nullConfig(WrenVM* vm)
   wrenFreeVM(otherVM);
 }
 
+static void multipleInterpretCalls(WrenVM* vm)
+{
+  WrenVM* otherVM = wrenNewVM(NULL);
+  WrenInterpretResult result;
+
+  bool correct = true;
+
+  // Handles should be valid across calls into Wren code.
+  WrenHandle* absMethod = wrenMakeCallHandle(otherVM, "abs");
+
+  for (int i = 0; i < 5; i++) {
+    // Calling `wrenEnsureSlots()` before `wrenInterpret()` should not introduce
+    // problems later.
+    wrenEnsureSlots(otherVM, 2);
+
+    result = wrenInterpret(otherVM, "main", "1 + 2");
+    correct = correct && (result == WREN_RESULT_SUCCESS);
+
+    wrenEnsureSlots(otherVM, 2);
+    wrenSetSlotDouble(otherVM, 0, -i);
+    result = wrenCall(otherVM, absMethod);
+    correct = correct && (result == WREN_RESULT_SUCCESS);
+
+    double absValue = wrenGetSlotDouble(otherVM, 0);
+    correct = correct && (absValue == (double)i);
+  }
+
+  wrenSetSlotBool(vm, 0, correct);
+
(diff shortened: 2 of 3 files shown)


mhermier:
AFAIK this change only fix a small subset of errors related to
wrenEnsureSlot. Otherwise could be enough to fix this particular issue.

kawa-yoiko:
Yes, but slot manipulation between multiple `wrenInterpret()` calls could be a common usage, so a dedicated test can be a good idea. Other similar issues I know involve some kind of re-entrancy (invoking Wren API inside foreign methods).

mhermier:
Yes but theses cases are in fact pretty common, but a little bit hard to
predict to produce reliable test cases.

Le lun. 10 févr. 2020 à 11:27, kawa-yoiko <notifications@github.com> a
écrit :

kawa-yoiko:
Indeed, it can be difficult to ensure test coverage for similar issues.

Maybe tests for unfixed issues should be added when they get understood more and fixed. Before that, black-box tests emulating common usage patterns can partly fulfill the need.

mhermier:
Sure, we can delay resolution of other bugs.
The same question always arise at the end of the question, is the API good
to correctly fix the situation, and AFAIK no one achieve to convince at
least me that the API needs to be changed.

mhermier:
I meant that the API needs to be changed. Sorry for my bad phrase
construction.

kawa-yoiko:
> I meant that the API needs to be changed.

The API related to slot array or something else? Would be interested to hear your reasons.

mhermier:
There are various small problems in the API, but the most visible one is
the managed reference to the execution context from the WrenVM object.
A simple use case to show the problem is: how do you retrieve a returned
object after wrenInterpret?
Also some problems comes from speed trades of, that are influenced by that
WrenVM design. The most visible issue, is the apiStack value caching that
is easy to corrupt. And despite the effort to hide/limit it, the VM is
reentrant in some places and tends to corrupt apiStack, in some situations.

ruby0x1:
Thanks @kawa-yoiko this is a good step forward, and thanks to @liquid600pgm too.

PR #737, opened 2020-03-29
Add wrenSetListElement to API

mode777:
I was writing a bulk method for vector transformations and I was looking to pass a lot of values to C and back to wren without re-allocations.

I noticed that while wren has wrenGetListElement there is no wrenSetListElement.
I quickly hacked it in for now - this pull request is more like a discussion base and probably not ready to be merged.


ruby0x1:
Thanks for the PR! This particular one had been caught up in some issues (see Files changed being too many), so I implemented manually since I had planned to before.

There was also a few minor issues in the PR btw. An assertion was copied from `wrenInsert..`  which isn't valid, since insert allows an out of bounds index on the end, where `wrenSet..` wouldn't.

I've implemented this here, and while doing so noticed that the get variant doesn't allow negative indices which isn't consistent with the way wren indices work. So now get and set both allow a negative index, and set exists.

https://github.com/wren-lang/wren/commit/97ebcc72c36f8ea81ed8bbfac8e165f7ef74495f

PR #739, opened 2020-03-30
Add vcpkg installation instructions

NancyLi1013:
`Wren `is available as a port in [vcpkg](https://github.com/microsoft/vcpkg), a C++ library manager that simplifies installation for `wren `and other project dependencies. Documenting the install process here will help users get started by providing a single set of commands to build `wren`, ready to be included in their projects.

We also test whether our library ports build in various configurations (currently only support static) configurations on various platforms (OSX, Linux, Windows: x86, x64) to keep a wide coverage for users.

I'm a maintainer for vcpkg, and [here is what the port script looks like](https://github.com/microsoft/vcpkg/blob/master/ports/wren/portfile.cmake). We try to keep the library maintained as close as possible to the original library.

### DIFF
--- a/util/vs2017/README.md
+++ b/util/vs2017/README.md
@@ -1,3 +1,14 @@
 # Building on windows
 
 Building the `wren` project requires python2 to available in the path as `python`, in order to download and build the libuv dependency. The download and build of libuv is done as a pre-build step by the solution.
+
+You can download and install `wren` using the [vcpkg](https://github.com/Microsoft/vcpkg) dependency manager:
+
+    $ git clone https://github.com/Microsoft/vcpkg.git
+    $ cd vcpkg
+    $ ./bootstrap-vcpkg.sh
+    $ ./vcpkg integrate install
+    $ vcpkg install wren
+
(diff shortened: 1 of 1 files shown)


underscorediscovery:
Hi @NancyLi1013 thanks for the info. You might want to wait until 0.3.0 is wrapped up as the build process is changing.

NancyLi1013:
Hi @underscorediscovery  thanks for your update.

ruby0x1:
Hi @NancyLi1013 
[Wren 0.3.0](https://github.com/wren-lang/wren/releases/tag/0.3.0) was released recently, so you are aware that [the build process is different](https://wren.io/getting-started.html#building-wren).  The file that this PR was made against is no longer present, but we can maybe find a place to mention it that makes sense.

NancyLi1013:
@nathanhourt 
Thanks for your above info. We will tr to update wren to 0.3.0 these days.
Also need your help to add vcpkg installation instructions in some places that you think it more suitable.

Really appreciate for your help and support.

Issue #742, opened 2020-04-03
Tests failing with Python 3.8


MageJohn:
I'm on Manjaro Linux, and the tests are failing for me due to issues with Python trying to parse invalid utf-8 being used as part of a test. In Manjaro `/usr/bin/python` is Python 3 by default (specifically Python 3.8 currently), not Python 2, and I confirmed that the tests don't fail when I used `python2.7 utils/test.py` instead. 

The output from running with Python 3:
```
(bin/wrend) Passed: 0 Failed: 0 Skipped: 0 Traceback (most recent call last):
  File "util/test.py", line 383, in <module>
    walk(join(WREN_DIR, 'test'), run_test, ignored=['api', 'benchmark'])
  File "util/test.py", line 304, in walk
    walk(nfile, callback)
  File "util/test.py", line 306, in walk
    callback(nfile)
  File "util/test.py", line 365, in run_test
    run_script(WREN_APP, path, "test")
  File "util/test.py", line 346, in run_script
    if not test.parse():
  File "util/test.py", line 67, in parse
    for line in file:
  File "/usr/lib/python3.8/codecs.py", line 322, in decode
    (result, consumed) = self._buffer_decode(data, self.errors, final)
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xab in position 65: invalid start byte
```

With a little print() debugging, I found the test causing the error, namely [test/regression/442-000088.wren](https://github.com/wren-lang/wren/blob/c7087652ee34ff9d2d1fd697bd8a4eac22a96182/test/regression/442-000088.wren)

underscorediscovery:
The tests have been fixed in the 0.3.0 branch and when that is released, it'll be resolved. Thanks for the report.

MageJohn:
I should probably have waited a few minutes then before making a patch. 😂

PR #743, opened 2020-04-03
fix: use bytes objects throughout Tests class

MageJohn:
A fix for #742

This fix works in both Python 2.7 and Python 3.8. It could use some testing in other versions of Python and on Windows however (assuming it's meant to work on Windows). 

This fixes the issue by avoiding utf-8, rather than just getting Python to not crash on the invalid byte. The latter is actually simpler, but I don't think there's a way that works on both Python 2 and 3.

### DIFF
--- a/util/test.py
+++ b/util/test.py
@@ -27,15 +27,15 @@ WREN_DIR = dirname(dirname(realpath(__file__)))
 WREN_APP = join(WREN_DIR, 'bin', 'wren' + args.suffix)
 TEST_APP = join(WREN_DIR, 'build', config_dir, 'test', 'api_wren' + args.suffix)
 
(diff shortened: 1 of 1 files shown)


util/test.py
```diff
@@ -267,6 +258,7 @@ def validate_output(self, out):
 
   def fail(self, message, *args):
     if args:
+      args = [arg.decode("utf-8") if type(arg) is bytes else arg for arg in args]
```
MageJohn, line 261: This is the bit I'm least happy about. It's not critical, but without it errors are printed as python bytes literals, e.g. `b'foo'`.

MageJohn:
https://github.com/wren-lang/wren/issues/742#issuecomment-608584513

I was too fast, this is already fixed.

PR #745, opened 2020-04-22
Fix example C code

nickgravelyn:
Hi there! Thanks for this lovely language. I'm really enjoying using it. I was working through the docs to embed Wren in my program and noticed the example for classes incorrectly used `->` when it should be `.` to assign the methods. Thought I'd send a little PR to fix that for you.


underscorediscovery:
nice thanks, I was gonna do a solid pass on the docs for 0.3.0 so this is a good find.

PR #756, opened 2020-06-07
Uniformise macro multi-line style.

mhermier:
Hi,
The code use various different code-styles for multi-lines macros.
This patch uniforms macro style to move the continuation line escape to stick to 80 columns.
While at it, also uniform to "while(false)" and remove extra spurious ';' (warns in pedantic mode) for the expression trick.
Cheers

### DIFF
--- a/src/include/wren.h
+++ b/src/include/wren.h
@@ -15,8 +15,8 @@
 
 // A monotonically increasing numeric representation of the version number. Use
 // this if you want to do range checks over versions.
-#define WREN_VERSION_NUMBER (WREN_VERSION_MAJOR * 1000000 + \
-                             WREN_VERSION_MINOR * 1000 + \
+#define WREN_VERSION_NUMBER (WREN_VERSION_MAJOR * 1000000 +                    \
+                             WREN_VERSION_MINOR * 1000 +                       \
                              WREN_VERSION_PATCH)
 
 // A single virtual machine for executing Wren code.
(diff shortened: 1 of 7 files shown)


bjorn:
For a change that is all about uniforming things, I don't like that there are both "while (false)" and "while (0)" used and sometimes there's a space in between and sometimes not. :-)

mhermier:
Nice catch, I didn't thought about the white spaces and missed one (0)... should be fixed now. Some day it should be possible to have a .clang-format, but the road is quite long, I tried it without much success, some part of the code are are destroyed by the heavy macros usages.

ruby0x1:
@mhermier I noticed you collapsed the `do` braces, which makes them inconsistent with the rest of the code formatting. Was that intended for any reason than avoiding an extra line?

```
do
{
```
became
```
do {
```

mhermier:
Ultimately personal taste, but mainly it enforce the block nature of the construct. Additionally, it also avoid a new line on something that is just a spurious construct to force the user to put a semi column. If we ultimately go down the rabbit hole, the while should also be put in a new line, but I don't find that it does help readability at all.

To detail a little bit more on my tastes, I really like formats like:
```
void foo() {
  if (bar ||
      baz) {
    baz = 42;
  }
}
```
I find:
- It makes quite compact code
- It use indent to exposes blocks
- It make easy to navigate in debug mode
- It produce quite small diffs if you need to expand small braces
- ultimately make it look like wren code to some extends (I agree the argument is a little bit made up...)

But this is my personal taste, and I'll stick to any .clang-format you would provide for reference, before being applicable every where.

ruby0x1:
I don't see a rabbit hole. The question was simply about making this code in particular inconsistent with the rest of the code base. Thanks for the PR! I just made them consistent after merging.

PR #757, opened 2020-06-08
wren: Remove a magic number in the compiler.

mhermier:
Hi,
Remove a magic number in field function, the 255 is meant to be MAX_FIELDS so that it doesn't produce code and enforce an error.
Cheers


ruby0x1:
good find, thanks!

PR #758, opened 2020-06-10
removed localhost

CrazyInfin8:
Noticed the site had a link that leads to localhost instead of being a relative link like the others on the page

Not too familiar how to use github... does this change do anything?


ruby0x1:
good catch

PR #759, opened 2020-06-10
Update gitignore

vladimyr:
Add VSCode project files and macOS specific stuff.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,6 +16,9 @@ Release/
 # The baseline file is machine-specific
 /test/benchmark/baseline.txt
 
+# VSCode project files.
+.vscode
+
 # XCode user-specific stuff
 xcuserdata/
 
@@ -33,3 +36,6 @@ ipch/
 *.user
 *.userosscache
 *.sln.docstates
+
+# macOS
+.DS_Store


bjorn:
Why close without any message? I guess the reasoning is, that these ignore rules are not Wren-specific and would be more suitable to put in the global `core.excludesfile`?

ruby0x1:
No @bjorn, this was GitHub closing ALL pull requests due to a branch rename. We're waiting to hear back from support about redirecting the PRs to the right branch.

vladimyr:
@bjorn All PRs got closed automatically by Github because the project owner renamed the main branch from `master` to `main`.

bjorn:
Ok, thanks for the enlightenment! I'm confused though, why would `main` be a better name than `master`? Was this discussed anywhere? I would suggest sticking with the default.

ruby0x1:
It was discussed in the discord yes. The default is going to change soon too, don't worry.

bjorn:
> The default is going to change soon too, don't worry.

Huh, interesting. I better not share my opinion on this lest I be called a racist...

I do think `main` is a better name in general anyway, though I would have waited for the default to change rather than proactively changing the name and even deleting the `master` branch before moving the pull requests over. ;-)

PR #733, opened 2020-03-04
More types for embedding in C

fabricemarie:

PR #421, opened 2017-03-11
New C API map support and various enhancements

fabricemarie:

PR #760, opened 2020-06-10
Expose Map API

vladimyr:
This exposes public slot-based API for working with maps.


ruby0x1:
This has been proposed several times! Here's a few:

#733 
#725
#591
#421

As always the PR is appreciated. I would recommend checking existing PRs first, and if appropriate opening a discussion before doing work.

PR #762, opened 2020-06-14
Small tweaks to error handling.

mhermier:
Hi,
The following changes allow RETURN_ERROR_FMT to take any number of arguments and not just one, and make use of the macro in a missed place.
Also removes a validateApiSlot call in wrenGetVariable. The function already check the slot index later because of the setSlot call. These kind of changes degrade a little the stack traces on debugging, but since these functions are pretty trivial, I don't think it is trivial to bloat them with duplicated checks.
Cheers,
   Michel

### DIFF
--- a/src/vm/wren_primitive.c
+++ b/src/vm/wren_primitive.c
@@ -23,9 +23,7 @@ static uint32_t validateIndexValue(WrenVM* vm, uint32_t count, double value,
 bool validateFn(WrenVM* vm, Value arg, const char* argName)
 {
   if (IS_CLOSURE(arg)) return true;
-
-  vm->fiber->error = wrenStringFormat(vm, "$ must be a function.", argName);
-  return false;
+  RETURN_ERROR_FMT("$ must be a function.", argName);
 }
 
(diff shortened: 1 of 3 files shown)


ruby0x1:
Isn't `__VA_ARGS__` a gcc extension? I'm wary of this reducing portability.

mhermier:
AFAIK the concatenation of empty argument list is an extension but __VA_ARGS__ is C. I'm trying to confirm minimal C version to have it in pre-processor.

ruby0x1:
Yea I looked at official docs, seems clang, msvc, gcc and even smaller compilers like tcc support it.
There's some nuance apparently about expansion within other macros but this use case is the basic one.
If it turns out to be a problem we can add a define to shield it but this seems fine for now.

ruby0x1:
Thanks, seems good otherwise.

mhermier:
Verified, __VA_ARGS__ is in C99 standard, and I specifically avoided extension usage '##' for comma elision, and forced that there is at least one argument by removing the msg argument. So it should be compliant to C99.

PR #763, opened 2020-06-14
Revert "Fix slot array corrupted by wrenInterpret()"

mhermier:
Reverts wren-lang/wren#730
Hi,
After further analysis, this change does not really fix the root cause of the problem.
In addition, it forbids the access of results of the invocation of wrenInterpret.

The issue could also be solved by applying:
```
diff --git a/src/vm/wren_vm.c b/src/vm/wren_vm.c
index e8dad4ab..bc30cdfe 100644
--- a/src/vm/wren_vm.c
+++ b/src/vm/wren_vm.c
@@ -1447,11 +1447,7 @@ WrenInterpretResult wrenInterpret(WrenVM* vm, const char* module,
   ObjFiber* fiber = wrenNewFiber(vm, closure);
   wrenPopRoot(vm); // closure.
 
-  WrenInterpretResult result = runInterpreter(vm, fiber);
-
-  vm->fiber = NULL;
-  vm->apiStack = NULL;
-  return result;
+  return runInterpreter(vm, fiber);
 }
 
 ObjClosure* wrenCompileSource(WrenVM* vm, const char* module, const char* source,
@@ -1565,7 +1561,7 @@ int wrenGetSlotCount(WrenVM* vm)
 {
-  if (vm->apiStack == NULL)
+  if ((vm->fiber != NULL && vm->fiber->numFrames == 0) || vm->apiStack == NULL)
   {
```
which has the advantage of not resetting the fibers (which would allow to get the result of wrenInterpret, but needs to be tested/fixed), but has another issue that force the recreation of fiber when invoked from C, leading to horrible performance in the benchmarks.

### DIFF
--- a/src/vm/wren_vm.c
+++ b/src/vm/wren_vm.c
@@ -1446,12 +1446,8 @@ WrenInterpretResult wrenInterpret(WrenVM* vm, const char* module,
   wrenPushRoot(vm, (Obj*)closure);
   ObjFiber* fiber = wrenNewFiber(vm, closure);
   wrenPopRoot(vm); // closure.
-
-  WrenInterpretResult result = runInterpreter(vm, fiber);
-
-  vm->fiber = NULL;
-  vm->apiStack = NULL;
-  return result;
+  
+  return runInterpreter(vm, fiber);
 }
 
 ObjClosure* wrenCompileSource(WrenVM* vm, const char* module, const char* source,
--- a/test/api/new_vm.c
+++ b/test/api/new_vm.c
@@ -13,43 +13,9 @@ static void nullConfig(WrenVM* vm)
   wrenFreeVM(otherVM);
 }
 
-static void multipleInterpretCalls(WrenVM* vm)
-{
-  WrenVM* otherVM = wrenNewVM(NULL);
-  WrenInterpretResult result;
-
-  bool correct = true;
-
-  // Handles should be valid across calls into Wren code.
-  WrenHandle* absMethod = wrenMakeCallHandle(otherVM, "abs");
-
-  for (int i = 0; i < 5; i++) {
-    // Calling `wrenEnsureSlots()` before `wrenInterpret()` should not introduce
-    // problems later.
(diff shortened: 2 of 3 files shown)


mhermier:
After more digging, we simply need to reset apiStack before calling runInterpreter:
```
 src/vm/wren_vm.c | 7 ++-----
 1 file changed, 2 insertions(+), 5 deletions(-)

diff --git a/src/vm/wren_vm.c b/src/vm/wren_vm.c
index e8dad4ab..5f07b4fa 100644
--- a/src/vm/wren_vm.c
+++ b/src/vm/wren_vm.c
@@ -1446,12 +1446,9 @@ WrenInterpretResult wrenInterpret(WrenVM* vm, const char* module,
   wrenPushRoot(vm, (Obj*)closure);
   ObjFiber* fiber = wrenNewFiber(vm, closure);
   wrenPopRoot(vm); // closure.
-
-  WrenInterpretResult result = runInterpreter(vm, fiber);
-
-  vm->fiber = NULL;
   vm->apiStack = NULL;
-  return result;
+
+  return runInterpreter(vm, fiber);
 }
 
 ObjClosure* wrenCompileSource(WrenVM* vm, const char* module, const char* source,
```
It seems to be more correct since that way the whole evaluation is performed without a dandling apiStack pointer.

But that does not fix the question about how to retrieve the result of interpretation. I tried some things without success yet.

mhermier:
I reminded myself that wrenInterpret acts at module level so it always evaluates to a Null function, so my point about accessing the result is invalid, but the last patch is still required be properly fix the issue, instead of reverting.
Tell me it you want me to close the issue and issue a pull request instead, I can easily issue it since it is already in my waiting to be pushed change queue.

mhermier:
Last patch is ready for review (contains test case that check that is possible to run foreign functions in wrenInterpret after a wrenEnsureSlot) is ready for review. Waiting for my other patch set to be reviewed/merged before publishing that one. Closing this push request in the meantime.

PR #765, opened 2020-06-17
Change `wren` to `wren_cli` in usage doc

jeffesp:
I noticed that the download is `wren_cli` and the docs just say `wren`. Not a big deal, but hopefully this helps.

### DIFF
--- a/doc/site/cli/usage.markdown
+++ b/doc/site/cli/usage.markdown
@@ -7,7 +7,7 @@ You can [download a build for your OS from the releases page](https://github.com
 
 ### Interactive mode
 
-If you just run `wren` without any arguments, it starts the interpreter in
+If you just run `wren_cli` without any arguments, it starts the interpreter in
 interactive mode, where you can type in a line of code, and it immediately executes
 it. You can exit the interpreter using good old Ctrl-C or Ctrl-D.
 
@@ -26,7 +26,7 @@ for (i in 1..10) System.print("Counting up %(i)")
 ### Running scripts
 
 The standalone interpreter can also load scripts from files and run them. Just
-pass the name of the script to `wren`. Create a file named "my_script.wren" in
+pass the name of the script to `wren_cli`. Create a file named "my_script.wren" in
 your favorite text editor and paste this into it:
 
 <pre class="snippet">
@@ -52,4 +52,4 @@ for (yPixel in 0...24) {
 
 Now run:
 
-    $ ./wren my_script.wren
+    $ ./wren_cli my_script.wren


bjorn:
Hmm, I'd prefer the CLI to be renamed back to `wren`. `wren_cli` looks pretty silly.

jeffesp:
I can probably get a PR together that would do that too, but just wanted things to be consistent.

mhermier:
Considering how *featured* wren_cli is, I would rename it to something else
than *wren*. To my idea, I found current wren_test to be more near to what
I think is the a base wren CLI, and wren_cli is much more something like
*nodewren*. But that just my opinion.

ruby0x1:
The split between wren the language and wren cli, a user of the language, was very intentional. 

The project is called “Wren CLI” and as such, this is the name of the binary as well. Calling it ‘wren’ muddies the water and undoes some of the clarity which was the whole point of making them distinct. It’s not wren, it has modules that wren does not have. 

Thanks @jeffesp , good find 👍

vladimyr:
> Hmm, I'd prefer the CLI to be renamed back to `wren`. `wren_cli` looks pretty silly.

https://github.com/Homebrew/homebrew-core/pull/55819/files#diff-b6987c08910028cea9ae989cf9968dc6R26 😉

ruby0x1:
@vladimyr since that is your PR can you please make it align with the project goals as stated above? Willfully choosing to go against the work we are trying to do here is not a good thing, and is not a positive contribution to the community. Thanks.

vladimyr:
There might be people in this world who have deep and complex feelings about naming CLIs (and FWIW I'm not trying to be sarcastic at all). I try to not get too much emotionally evolved in such activities and stay on practical side of things:
```
~
➜ ls /bin /sbin /usr/bin /usr/sbin /usr/local/bin | grep -i 'cli$' | wc -l
       0
```
with exactly 198 install Homebrew packages plus base Mac OS userland...

As you can clearly see from commit dates in that PR, my work predates this discussion so it is not that I'm (or was) trying to act against anything. I perfectly understand and support this split, it totally makes sense to me to refer to it as Wren CLI in the documentation but calling the actual binary `wren_cli` is wildly unusual. Not in the context of your development workspace, there you can call it whatever you like ofc, but install it under `$PREFIX/bin/wren_cli` would make it unique flower in mine and many more fs trees. Correct me if I'm wrong but this community believes there are more sophisticated selling points wren could and should use than making it stick out in a weirdly named way?

To prove you and everyone else that I'm not active contributor by day and the rogue devil by night I'll simply close that PR because I disagree with going against conventions. That ofc doesn't mean I can or would try to prevent anyone to take my work, modify that tiny bit and resubmit it if there is community-wide consensus around it.

ruby0x1:
I'll refer you to your previous comment before mine about what made me say that (I didn't check timestamps).

What we're referring to here is two separate entities:
Wren is an embeddable programming language, it is NOT a CLI. 
What we are working on is a language, this is the main project.

The fact that the CLI is a part of the wren-lang organization and used to be a part of the wren repo are inconsequential. The fact that it has _incompatible and entirely different APIs_ to Wren itself, and then calling itself `wren`, is a problem.  It should not be the case.

The fact that `node` is an arbitrary name and the fact that `python` named it's cli the same as the language don't really weigh much to me. I don't see how naming it `pineapple` is any different from `wren_cli` except with potentially less underscores, and less discoverability via search. 

For now, the binary is `wren_cli`, because that's the name of the project.

bjorn:
> Calling it ‘wren’ muddies the water and undoes some of the clarity which was the whole point of making them distinct. It’s not wren, it has modules that wren does not have.

Could we then re-introduce a basic command-line `wren` tool in the Wren repository that is essentially just vanilla Wren? I'm not interested in "`wren_cli`" but would just like to have a convenient way to play around with the Wren language without needing to first embed it into my application.

PR #774, opened 2020-07-09
runFile: Free file source before exiting the function

CohenArthur:
This prevents a memleak, noticeable when running `wren_test` under
`valgrind`. For example, the following command would leak

```sh
./bin/wren_test_d any_example.wren
```


ruby0x1:
Thanks!

PR #775, opened 2020-07-11
[0.4.0] Import as

ruby0x1:
I have still found the ability to rename an imported variable vital over time. 
I plan to introduce this in 0.4.0 with the `as` keyword. This PR replaces #464


mhermier:
I have mixed fillings about that change. While I'm not particularly against "as" to be a new keyword, I don't find it particularly superior to:
```
import "..." for OtherName = Variable 
```
I also think there is a little ambiguity that is introduced by using `as` here. From what I remember variables are imported by copying and the `as` feels more like some kind of name alias/overload, inducing the variable is not copied, which lead to a little oddity in the language.

The `as` keyword can also be used to introduce variable names aliases at expression level, and I'm not sure it is really a desirable language feature. I can see a usage for refactoring, but I wonder about complexity in the compiler, the practicality...

ruby0x1:
See the test cases, the values are copied because the variable is loaded onto the stack and then stored into a slot, all this does is make the slot name different if as is present. mechanically there is no change.

mhermier:
Ok so I'm not mistaken, the ambiguity is introduced by this change then. It is a minor issue thought, but I found it counter intuitive that this language choice can not convey the actual behavior. So the operator = usage here makes more sense to me as it enforce the idea of copying. But it may be only me.

doc/site/modularity.markdown
```diff
@@ -42,6 +42,17 @@ if (thirsty) {
 }
 </pre>
 
+If you need to import a variable via a different name, you can use
```
bjorn, line 45:
```suggestion
If you need to import a variable under a different name, you can use 
```

doc/site/modularity.markdown
```diff
@@ -42,6 +42,17 @@ if (thirsty) {
 }
 </pre>
 
+If you need to import a variable via a different name, you can use 
+`import "..." for Name as OtherName`. This looks up the top-level variable
+`Name` in *that* module, but declares a variable called `OtherName` in *this* module
+with it's value.
```
bjorn, line 48:
```suggestion
with its value.
```

ruby0x1:
Note that `as` (and the wording) wasn't an arbitrary choice. 
Dart, JS, Python and other languages use `as OtherName` because it's a simple english expression that is easy to understand.
Other languages use the same wording for casts, like rust and swift have `let a = 0 as u8;`.

mhermier:
It can be likewise decided to make `import "module"` also available as a regular expression that is returns the module variable (to allow assignment), and make `for` also a binary operator, that allow extracting variables from a map to local scope, so that it is compatible with module import. That way the as operator is also not required.

To add to the cons, and comparing to C++, the `as` operator also has the disadvantage to be a right definition like `typedef`, and readability seems to have improved since the adoption on `using`. As practice seems to had proven, we need the defined first for fast reading, not the possibly ugly/dirty definition.

ruby0x1:
I agree `import "something" Name OtherName` would be confusing. This is not the case, though.

mhermier:
To clarify my words, I don't care about the choice of the `as` keyword, could be anything, and I know `as` is used in a lot of other language. I point that assignment syntax is also an interesting possibility with quite some pros and cons, that I prefer over introducing a new keyword.

DethRaid:
```wren
import "..." for OtherName = Variable 
```
is less clear to me than
```wren
import "..." for Variable as OtherName
```

The first statement reads as "import this module so we can use OtherName to refer to Variable", while the second reads as "import this module to we can use Variable and call it OtherName". It's just more clear, and more conventional, and I don't see a real reason to use the assignment operator here

mhermier:
It is the same reason to why `using` is superior to `typedef` in C++.
When I read:
```
import "..." for OtherName = Variable
```
I read "Import this module, that give me this symbol followed by =" and I stop reading unless I really need to know what is the underlying type.
With the other syntax you have to read allllll the text before reaching the usable symbol name that doesn't has a fixed position.

If you import many symbols, you can do something like:
```
import "..." for
  OtherName = Variable,
  FooOtherName = FooVariable,
  ...
  BarOtherName = BarVariable
```
And the defined symbols is always visible at first sight, and does not require reading the whole expression, hacky text alignment or editor search.

PR #776, opened 2020-07-11
[0.4.0] allow import strings to use string interpolation

ruby0x1:
Adds the ability to do:
```js
import "configs/%(which_config)" for Config
```

I wanted to open this as a PR for interest sake and discussion purposes.
Disclaimer: I'm not sure yet if this is a good idea due to the thoughts below.

The biggest downside to this is that the determinism of import paths at compile time is lost. 
This means for things like code completion, static analysis etc you don't have a simple string any more, 
but rather expressions that can only be answered through execution of the code. **BUT**.

Note that this only applies to the string itself, not the result/contents of said string. 
For example, `import "http://127.0.0.1/api/random_quote" for Quote` is not somehow deterministic to code completion. The implementation details of the result of the import are entirely up to the host, not the code completion handler. 

The import path string is opaque, and in theory, has no meaning without a host giving it some. 
**There is no escaping that the host has to answer these questions for those systems to function, somehow.**

Another example from my own implementation details. 
Module imports follow a pattern where `package: file` is used, which means the host needs to explain:
- how to find `package: file`
- which includes knowing where `package` is located but also
- which `version` is used, as well as other project specific details

Just some things to think on, interested in hearing thoughts.

(Edit: not sure why it says the tests are failing, they all pass locally. Will figure it out later.)

### DIFF
--- a/src/vm/wren_compiler.c
+++ b/src/vm/wren_compiler.c
@@ -2721,6 +2721,7 @@ static int getByteCountForArguments(const uint8_t* bytecode,
     case CODE_FOREIGN_CONSTRUCT:
     case CODE_FOREIGN_CLASS:
     case CODE_END_MODULE:
+    case CODE_IMPORT_MODULE:
       return 0;
 
     case CODE_LOAD_LOCAL:
@@ -2761,7 +2762,6 @@ static int getByteCountForArguments(const uint8_t* bytecode,
     case CODE_OR:
     case CODE_METHOD_INSTANCE:
     case CODE_METHOD_STATIC:
-    case CODE_IMPORT_MODULE:
     case CODE_IMPORT_VARIABLE:
       return 2;
 
@@ -3343,11 +3343,18 @@ static void classDefinition(Compiler* compiler, bool isForeign)
 static void import(Compiler* compiler)
 {
   ignoreNewlines(compiler);
-  consume(compiler, TOKEN_STRING, "Expect a string after 'import'.");
-  int moduleConstant = addConstant(compiler, compiler->parser->previous.value);
 
+  if(match(compiler, TOKEN_STRING)) {
+    int constant = addConstant(compiler, compiler->parser->previous.value);
+    emitShortArg(compiler, CODE_CONSTANT, constant);
+  } else if(match(compiler, TOKEN_INTERPOLATION)) {
+    stringInterpolation(compiler, false);
+  } else {
+    error(compiler, "Expect a string after 'import'");
+  }
+ 
   // Load the module.
-  emitShortArg(compiler, CODE_IMPORT_MODULE, moduleConstant);
+  emitOp(compiler, CODE_IMPORT_MODULE);
 
   // Discard the unused result value from calling the module body's closure.
   emitOp(compiler, CODE_POP);
--- a/src/vm/wren_vm.c
+++ b/src/vm/wren_vm.c
@@ -1281,7 +1281,7 @@ static WrenInterpretResult runInterpreter(WrenVM* vm, register ObjFiber* fiber)
       // value. It will be popped after this fiber is resumed. Store the
       // imported module's closure in the slot in case a GC happens when
       // invoking the closure.
-      PUSH(importModule(vm, fn->constants.data[READ_SHORT()]));
+      PUSH(importModule(vm, POP()));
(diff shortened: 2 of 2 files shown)


mhermier:
Go for it, I suggested/predicted that usage when @munificent was implementing imports.

ruby0x1:
@munificent I implemented this as an alternative to the issue I mentioned over DM with import running a closure (which doesn't trivially happen outside of the interpret loop, like from a PRIMITIVE or foreign). 

Around 80% of what makes my engine work relies on the ability to import modules from a variable, so I consider the ability to do that vital. Even the user project is a wren script. Previously I had a `System.importModule` as a local primitive that let me do it, but the way imports work now vs the version I was on before (where import returned a fiber, not a closure), isn't working.

So, right now my engine is relying on this particular implementation until I can figure out how to run the closure outside the interpret loop safely/correctly.

I agree with the concerns about it, hence a discussion.

mhermier:
@munificent It solve/remove the trivial macro switch problem, without the need to have macros. It is the same thing as allowing expressions as parent class when creating a class. It is a must have so macros are not needed. It is also elegant since it is integrated in the language.

I don't understand the fear of your analyzability, if it is a plain string it is analyzable, if not then it is the user responsibility to do it. Exported symbols are defined clearly, so the compilation unit is round and square. Only the linking issue that happens at runtime can be a source of error, but this doesn't change from what we currently have.

The other alternative, I see right now, would be to have conditional imports, which I think is inferior for various reasons.

For reference, AFAIK C++ already allow that in the C pre-processor language at least since includes path are evaluated before included, shell sourcing allow any valid string, and probably many more languages have such feature.

ruby0x1:
As an update, with fresh eyes I looked at getting my original importModule implementation to work and found the way for it to happen correctly. Maybe it can live in Meta or similar, but it's necessary to be able to do it, even if the `import` keyword isn't the mechanic by which it's done :)

mhermier:
That smell like feature duplication. Waiting to see how it goes.

ruby0x1:
hm? The feature of importing a string from a variable _does not exist_ at present, I had added an implementation (which I added to `System.importModule`) out of necessity, which I've used for years. That broke when I was updating to latest Wren, which led me to explore this idea, but I have fixed the broken one. If this lands, I don't need that one. And if it doesn't, I still do. No duplication.

mhermier:
I'm waiting to comment ;)

mode777:
From a language design point of view, I agree with @munificent. I think the benefits do not outweigh the costs of having potentially unreadably code and problems with static analysis (not only language services but also things like bundling, tree shaking).

I would prefer to have different construct for dynamic imports similar to typescripts `import()` https://mariusschulz.com/blog/dynamic-import-expressions-in-typescript

@ruby0x1 Your System.importModule seems like an ideal solution to me. I would also be **very** interested how you got it back up and running 😇

mhermier:
The argument of unreadability is a falcy, else it means string
interpolation should be removed either. In addition, it remove some need of
some kind of factory pattern which is even more code/and spread import all
around the code, which does not help reading code either.

If one really needs string interpolation includes, you are probably in a
generic land, so bundling and tree shaking is not your main concern, since
you are either dealing with versioning or platform specifics. In addition,
if you change string interpolation to fixed string stressing the code or
what ever optimization..., it become a one liner diff, and not restructure
the code because the include is made somewhere in the code.

In addition, if you need string interpolation support, you probably want to
limit it's usage to a proxy/adapter file (set). I suspect no sane code can
use this feature all around the code in a controlled manner.

Anyway if you try to do the same thing without it, you end up needing to do
the factory pattern, possibility loading a huge bulk of code that might not
even apply to your system...

System.importModule in some form is required as an implementation
detail/backend, so this one is a no brainier. But it has the same tree
shacking issue, since you can't predict imported code.

I agree, it is required to have some dynamic loading a expression level,
but having it at import level, is a nice and elegant way to do it
seemlesly. The tooling argument, to me, is a developer being *lazy* at
updating it's code (not that it is a bad thing either).

ruby0x1:
The nature of this change is a can of worms and not a good fit for Wren atm. Thanks for the thoughts everyone! 

Anyone interested in doing it in their own fork has this a reference implementation and can explore that freely.

Issue #736, opened 2020-03-28
Magic Numbers in Exit Codes


clsource:
It seems that exit codes are following the [`BSD Standard`](https://github.com/openbsd/src/blob/master/include/sysexits.h).

As shown here (74 For input/output error EX_IOERR)

https://github.com/wren-lang/wren/blob/master/src/cli/vm.c#L56

And here (70 For Internal sofrware error EX_SOFTWARE)

https://github.com/wren-lang/wren-cli/blob/master/src/module/scheduler.c#L29

More info about exit codes here

https://krueger.report/programming/exit-status-codes/

Code List

```

    0 - Successful termination
    1 - Unspecified error
    2 - (grep) Incorrect command usage | (bash) I/O error
    64 - Command line usage error
    65 - Data format error
    66 - Input is not openable
    67 - Addressee unknown
    68 - Unknown host name
    69 - Service unavailable
    70 - Internal software error
    71 - System error
    72 - Critical OS file missing
    73 - Can no create output file
    74 - I/O error
    75 - Temporary failure
    76 - Remote error in protocol
    77 - Permission denied
    78 - Configuration error
    126 - Command found but not executable
    127 - Command not found
    128 - Invalid exit code
    128 + n - Fatal error signal n, i.e. kill -9 $PPID (128 + 9 = 137)
    130 - Command terminated by Ctrl-C
    255+ - Exit code out of range
```

I would recommend to include a header file like BSD sysexits.h. So these magic numbers can replaced with a more meaningful alternative.

## BSD sysexits.h

```c
/*	$OpenBSD: sysexits.h,v 1.5 2003/06/02 19:34:12 millert Exp $	*/
/*	$NetBSD: sysexits.h,v 1.4 1994/10/26 00:56:33 cgd Exp $	*/

/*
 * Copyright (c) 1987 Regents of the University of California.
 * All rights reserved.
 *
 * Redistribution and use in source and binary forms, with or without
 * modification, are permitted provided that the following conditions
 * are met:
 * 1. Redistributions of source code must retain the above copyright
 *    notice, this list of conditions and the following disclaimer.
 * 2. Redistributions in binary form must reproduce the above copyright
 *    notice, this list of conditions and the following disclaimer in the
 *    documentation and/or other materials provided with the distribution.
 * 3. Neither the name of the University nor the names of its contributors
 *    may be used to endorse or promote products derived from this software
 *    without specific prior written permission.
 *
[...]

PR #777, opened 2020-07-11
Remove magic values as exit codes

CohenArthur:
References #736


ruby0x1:
Thanks @CohenArthur. One minor change, could you move the changes from `wren.h` into `test.h`, as they don't belong as a part of the public Wren API and are instead local to the test application. 

Otherwise, looks good.

PR #782, opened 2020-07-16
Hashset in Symboltable

CohenArthur:
I'm looking for feedback on that PR. I got it working but there are still a few things to fix, namely a couple memleaks. I'll definitely fix them before taking the PR out of "Draft" mode.

I'd like to know if this is going in a way that you like. There's definitely a speedup on the lookup, but I'd like to know if you'd consider adding it or if I should drop it.

I have a few questions as well, mostly regarding the hash function. A hash is computed in the `ObjString` struct, but I have to implement my own hashing function since `hashString()` is a static function, and I need to hash in order to look up objects. Would you rather I used hashString() ? Is that fine ?

### DIFF
--- a/src/vm/wren_utils.c
+++ b/src/vm/wren_utils.c
@@ -64,7 +64,7 @@ static void wrenSymbolTableGrow(WrenVM *vm, SymbolTable *symbols);
 
 // Insert a symbol in the bitset with a specific index
 static size_t wrenSymbolTableInsertHashIdx(WrenVM *vm, SymbolTable *symbols,
-        ObjString *symbol, size_t idx)
+                                           ObjString *symbol, size_t idx)
 {
     size_t hashIdx = wrenHashDjb2(symbol->value, symbol->length) % symbols->hCapacity;
 
--- a/src/vm/wren_utils.c
+++ b/src/vm/wren_utils.c
@@ -62,8 +62,6 @@ int wrenSymbolTableAdd(WrenVM* vm, SymbolTable* symbols,
 {
     ObjString* symbol = AS_STRING(wrenNewStringLength(vm, name, length));
 
-    wrenPushRoot(vm, &symbol->obj);
-
     // Compute the true index and the hash index
     size_t buffIdx = symbols->objs.count;
     size_t hashIdx = wrenHashDjb2(name, length) % symbols->hCapacity;
@@ -73,7 +71,7 @@ int wrenSymbolTableAdd(WrenVM* vm, SymbolTable* symbols,
         hashIdx++;
 
     symbols->hSize++;
-    if (symbols->hSize == symbols->hCapacity)
+    if (symbols->hSize == symbols->hCapacity || hashIdx >= symbols->hCapacity)
         wrenSymbolTableGrow(vm, symbols);
 
     // "Initialize" the element in the set
@@ -81,6 +79,7 @@ int wrenSymbolTableAdd(WrenVM* vm, SymbolTable* symbols,
     symbols->bitset[hashIdx].idx = buffIdx;
 
     // Add the name to the buffer
+    wrenPushRoot(vm, &symbol->obj);
     wrenStringBufferWrite(vm, &symbols->objs, symbol);
     wrenPopRoot(vm);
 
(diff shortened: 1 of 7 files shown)


mhermier:
This seems a too much complicated/specific code for marginal gain. Can you you provide number of what it tries to improves ?

CohenArthur:
It tries to improve this: https://github.com/wren-lang/wren/blob/f81cb5d23cbf3a6bb5b83511cf6a47dcf2f130e9/src/vm/wren_utils.c#L43

As said in the comments, the lookup is in O(n). Since the code almost exclusively uses `wrenSymbolTableEnsure()` which then calls `wrenSymbolTableFind()`, and not `wrenSymbolTableAdd()`, it should give some increase in performance.

If you're worried about the number of allocations, it can be reduced as well. This is just a draft

mhermier:
Being O(n) is not really a problem for one time things like compiling, but it is if it does become a runtime issue.

For cheaper/re-usability of code, instead of this custom Set, why not add a Map of Symbol to index ?

CohenArthur:
It's used in the `wrenCompile()` function, and I don't think it's a one time thing, but I might be wrong.

What do you mean by Map of Symbol to Index ? Do you mean a regular hashmap or something along those lines, where each symbol would have its index based on the hash of its name ?

This implementation is useful because the compiler goes through all variables and their variableNames sequentially. So having it ordered is nice and easy

mhermier:
I mean compilation is marginal to runtime. Improving it, is interesting but it has to be wisely done.

I mean a regular hashmap, where key would be the symbol and value its index. And maybe for completeness also register values and keys (can't remember the name of this variant of map).

It is nice and easy for what kind of usage ?

CohenArthur:
It is useful here https://github.com/wren-lang/wren/blob/f81cb5d23cbf3a6bb5b83511cf6a47dcf2f130e9/src/vm/wren_compiler.c#L3496 in that loop

mhermier:
I didn't said it is not use-full, I said it will provide marginal gain. It implies that hacking it should be either to reduce the code size or to improve it in significant way, and preferably both.

CohenArthur:
Oh, I see what you mean. I agree that this definitely doesn't reduce the code size but it should make the interpreter faster, which is what I was going for.

CohenArthur:
And while it is not a very realistic example, generating a wren source file composed only of variable declarations and then running it gets way faster. With a file generated with the following script:

```python
for i in range(64000):
    print(f"var val_{i} = {i}")
```

The execution time goes from 5.5s on average to 0.04s on average.

Even though it's not really realistic, it shows the impact it could have on important wren codebases

mhermier:
I gave a try at using Map, and this patch series needs a preliminary patch to introduce something like:
`const char *wrenSymbolTableGet(const SymbolTable* symbols, int symbol)`
to reduce some of the noise that make this patch-set more hard to read than necessary. I can provide the patch or try to propose it once my current pull request is accepted.

CohenArthur:
Sure, I can definitely work on this. As I said, this is just preliminary. Right now, it does more allocation than I want and needs some cleanup

src/vm/wren_utils.c
```diff
@@ -91,7 +91,7 @@ static void wrenSymbolTableGrow(WrenVM *vm, SymbolTable *symbols)
     wrenSymbolTableClearBitset(symbols);
 
     size_t oldCap = symbols->hCapacity;
-    symbols->hCapacity *= 2;
+    symbols->hCapacity = wrenPowerOf2Ceil(symbols->hCapacity);
```
mhermier, line 94: This change is broken: `wrenPowerOf2Ceil` does not return the next power of 2 if the current number is a power of 2. https://godbolt.org/z/P3vGnd

CohenArthur: Oh good catch. So `wrenPowerOf2Ceil(symbols->hCapacity + 1)` would work. I think *= 2 is simpler, I call `wrenPowerOf2Ceil()` simply for consistency but I feel like it might make it harder to read

CohenArthur:
@mhermier I'm not sure having a `get()` function would really clear things up. From my understanding, the code relies a lot on the fact that the SymbolTable is ordered. Which is why functions such as `wrenSymbolTableEnsure()` return the symbol's index. For example, the compiler uses the same index for the variables and the variableNames.
Using this hacky structure allows to keep an ordered buffer.
If we decide to go the "regular hashmap" route and rewrite this, then it also requires a rework in the compiler. I'm not sure the benefits outweigh the cost, but again, I might be wrong.

I fixed the remaining memory leaks and am happy with the current implementation, but I will obviously listen to feedback. This PR speeds up code interpretation in exchange for more memory allocation

mhermier:
I have have to confirm it and make some adaptations (because of local changes), but I have a map implementation that make a time of 0.000307s with your benchmark, on my machine.

CohenArthur:
Nice ! That's definitely an insane speedup

mhermier:
Well it needs to be verified and polished. I have some local changes that might alter the results, or botched the benchmark, but the number is pretty nice.

mhermier:
I made some changes and discovered a flaw in my benchmark but improvement to baseline is good:
```
origin:  compiler_many_var - wren       .......... 9.88s 0.0340
patched: compiler_many_var - wren       .......... 0.05s 0.0019
```

mhermier:
My change set for my implementation is ready to land. Waiting for my current pull request to be resolved before I can propose it.

CohenArthur:
Is there anything I should do to make this PR move forward?

mhermier:
Here is your bus ticket... XD

PR #787, opened 2020-07-22
Segmentation Fault when calling unknown foreign constructors

CrazyInfin8:
I noticed that when calling constructors of foreign classes that don't have constructors set on the C side, the vm may crash with a segmentation fault rather than abort.
```C
#include "wren.h"
int main(int argc, char const* argv[]) {
    WrenConfiguration cfg;
    wrenInitConfiguration(&cfg);
    WrenVM* vm = wrenNewVM(&cfg);
    wrenInterpret(vm, "main",
                  "foreign class Foo {\n"
                  "    construct unknownConstructor() {}\n"
                  "}\n"
                  "\n"
                  "Foo.unknownConstructor()"); // <- Segmentation fault from calling constructor
    wrenFreeVM(vm);
    return 0;
}
```
I don't know much about the VM's internals so I don't really intend for this PR to merge but this little hack aborts the vm instead of crashing it.

I didn't do any extensive testing so not sure about any side effects but passes all tests ran from `.travis.sh` (at least on WSL)

### DIFF
--- a/src/vm/wren_vm.c
+++ b/src/vm/wren_vm.c
@@ -626,10 +626,11 @@ static void createForeign(WrenVM* vm, ObjFiber* fiber, Value* stack)
   // TODO: Don't look up every time.
   int symbol = wrenSymbolTableFind(&vm->methodNames, "<allocate>", 10);
   ASSERT(symbol != -1, "Should have defined <allocate> symbol.");
-
-  ASSERT(classObj->methods.count > symbol, "Class should have allocator.");
+  
+  // ASSERT(classObj->methods.count > symbol, "Class should have allocator.");
   Method* method = &classObj->methods.data[symbol];
-  ASSERT(method->type == METHOD_FOREIGN, "Allocator should be foreign.");
+  // ASSERT(method->type == METHOD_FOREIGN, "Allocator should be foreign.");
+  if(method->as.foreign == NULL) return;
 
   // Pass the constructor arguments to the allocator as well.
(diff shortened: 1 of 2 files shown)


CohenArthur:
The ASSERTS() are useful because they tell you more about what is happening in debug mode. I don't think they should be removed, but the check is probably important for release mode.

mhermier:
Foreign class are not intended to not be allocated by normal means, to
allow C bindings. The fact that your patch allows to create a foreign class
without the allocator foreign function is an error.

CrazyInfin8:
@mhermier what I’m worried about is if I run sandboxed wren code in a program, I don’t want it to crash the entire program with a segfault but would rather the fiber be aborted instead.

@CohenArthur I did remove some ASSERTs but I don’t intend this edit to be a proper fix. I don’t think I know enough about there internal to do a proper fix.

mhermier:
Ideally, `wrenInterpret` should return `WREN_RESULT_COMPILE_ERROR` in this case, since the program is failing to *link*. It would require a little bit more refactoring than the current proposed patch.

CrazyInfin8:
Wait, I’m trying to learn how to use GitHub properly, should I have created an issue instead? (And is so, should I just open a new issue then?)

mhermier:
Depends on what you are trying to do. If you have some code a push request
is fine. If you only need to talk about a problem or something with
potentially consequences issue is the way.

CrazyInfin8:
Opened an issue [here](https://github.com/wren-lang/wren/issues/811) instead.
closing this thread.

PR #788, opened 2020-07-22
Add userData pointer to reallocateFn

koljakube:
In the spirit of [this PR](https://github.com/wren-lang/wren/pull/406), I am trying to write a wrapper for Wren in [Zig](https://ziglang.org). Zig is very explicit about everything, including allocators.

Wren is currently written with a global `malloc` in mind. With this (sadly, breaking) change, `reallocateFn` accepts a `void* userData`. This can be used to locate an allocator instance, so that each VM can use its own allocation method (e. g. fast ones with limited memory for smaller VMs and general-purpose ones for larger VMs).

I only added `void* userData` and not `WrenVM* vm` since I thought that messing with the VM itself during allocations might not be too wise.

**Breaking API changes**
Added additional `userData` parameter to WrenReallocateFn.
```
typedef void* (*WrenReallocateFn)(void* memory, size_t newSize, void* userData);
```


mhermier:
I would always put user data in front, since it is some kind of anonymous
of object style call.

koljakube:
I was wondering if I should do that. I put it in the back to differentiate it from the usual `WrenVM*`, and because I thought that memory allocation functions feel different than the other ones. But I can change it and put the pointer in front.

ruby0x1:
It's fine on the right too, don't change it just yet.

CohenArthur:
Personally I think it's better on the right. It's similar to functions like `pthread_create` and others. Plus, it makes more sense for an allocator to not have that extra user data in front. This way you're just gonna use it like `realloc()`, with an extra `NULL` at the end most of the time

mhermier:
The fact that you want to differentiate it, is exactly an anti pattern that
C love, in the sense of using and abusing pointer cast.

About realloc, the argument is also a falcy if you look how realm API looks
like. And in the general case, the important thing is that all/most other
arguments are aligned so that they can be copied fast in worse case. But
anyway we are in a slow path, if we loose few cycle rearranging the stack
it is always minor to the heap management.

koljakube:
Is there anything I can change or do to get this accepted or rejected with a reason? I just [published the initial code](https://github.com/koljakube/zapata) for the mentioned wrapper, and would love to be able to build it with the official source.

mhermier:
Wait, seems maintainer is busy. Bus factor is in effect.

ruby0x1:
Thanks @koljakube! It's reasonable to need access to the userData (that's why it exists).

Issue #801, opened 2020-08-19
System.print causes different output


mathewmariani:
I came across this strange error were adding a `System.print` statement changes the expected output of a function.

```
class SudoList {
  construct new() {
    _a = [10, 7, 8, 9, 1, 5]
  }
  
  sort(fn) {
    quicksort_(0, _a.count - 1, fn)
    System.print(_a)
  }

  quicksort_(a, b, fn) {
    if (a < b) {
      var p = partition_(a, b, fn)
      quicksort_(a, p - 1, fn)
      quicksort_(p + 1, b, fn)
    }
  }

  partition_(a, b, fn) {
    var p = _a[b]
    var i = a - 1    
    for (j in a..(b-1)) {
      // commenting and uncommenting this line yields different results
      System.print("bug stopper") 
      if (fn.call(_a[j], p)) {
        i = i + 1
        var t = _a[i]
        _a[i] = _a[j]
        _a[j] = t
      }
    }
    var t = _a[i+1]
    _a[i+1] = _a[b]
    _a[b] = t
    return i+1
  }
}

var w = SudoList.new()
w.sort{|a, b| a < b }
```

I've run this code in both the VM and in the browser at [wren.io/try](https://wren.io/try/); additionally, the error is not present when running the script on [wren-nest](http://ppvk.github.io/wren-nest/).

When `System.print("bug stopper")` is uncommented the program runs as expected.
When commented out the program crashes giving a `Right operand must be a number.` error.

Complete error:
```
Right operand must be a number.
[compile line 41] in sort(_) block argument
[compile line 25] in partition_(_,_,_)
[compile line 14] in quicksort_(_,_,_)
[compile line 16] in quicksort_(_,_,_)
[compile line 8] in sort(_)
[compile line 41] in (script)
```

mhermier:
If you are on Linux try to run wren inside valgrind. I suspect you are
hitting a variation of the well-known stack corruption.

ruby0x1:
This should hopefully be fixed soon, just waiting on some feedback. Thanks for the report and examples!

mathewmariani:
> If you are on Linux try to run wren inside valgrind. I suspect you are hitting a variation of the well-known stack corruption.

I'm running on macOS currently.

ruby0x1:
Hi @mathewmariani, try with #807 merged, I tested your code and the results are consistent.

Issue #761, opened 2020-06-10
Stack corruption


mhermier:
Here is a reduced test to the long standing stack corruption issue:

```
// Level of required nesting is dependant on INITIAL_CALL_FRAMES
var Fn1 = Fn.new {|acc| Fiber.yield(acc) }
var Fn2 = Fn.new {|acc| Fn1.call(acc, null) }
var Fn3 = Fn.new {|acc| Fn2.call(acc, null) }
var Fn4 = Fn.new {|acc| Fn3.call(acc, null) }
var Fiber1 = Fiber.new { Fn4.call(42) }

System.print(Fiber1.call()) // expect: 42
System.print(Fiber1.call("caramba")) // expect: caramba
```
Instead of returning caramba, second call to it returns 42. I have not been able to weaponize it to make the test crash, but at least it can be tested somehow.

mhermier:
Another one based on #770
```
var Fib1 = Fn.new {|n|
  if (n <= 2) {
    return 1
  } else {
    return Fib1.call(n-1) + Fib1.call(n-2)
  }
}

var Fib2 = Fn.new {|n|
  if (n <= 2) {
    return 1
  } else {
    var left = Fib2.call(n-1)
    var right = Fib2.call(n-2)
    return left + right
  }
}

System.print(Fib2.call(12)) // expect: 144
System.print(Fib1.call(12)) // expect: 144
```

Just in case, reducing it to:
```
var Fib = Fn.new {|n|
  if (n <= 2) {
    return 1
  } else {
    return Fib.call(n-1) + Fib.call(n-2)
  }
}

System.print(Fib.call(12)) // expect: 144
```
Trigger the bug but does not make it visible to user for some reasons.

ruby0x1:
Thanks, I think we have a fix for this, at least in my engine where I have a fix applied for another bug, this one is giving expected results too.

mhermier:
Could you issue a pull request of that fix before merging? Since that stuff is in the critical path, and can have possible huge impacts in speed.

I also have a crude patch that fix that issue by doing a STORE_FRAME/LOAD_FRAME around the `switch (method->type)` block that work quite well. It is a little bit expensive (in term of time), but seems to cover all the case.

Another solution would be to revert, the function call change.

So before you push, please allow to evaluate your solution.

ruby0x1:
Don't forget the benchmarks exist.

[...]

PR #802, opened 2020-08-19
Added List.sort(comp) to List module

mathewmariani:
Added `sort(comp)` function to List using an in-place quicksort implementation.


CohenArthur:
Your doc example shows it, but it would be nice to have some documentation on the expected behavior of `comp()`. For example, saying it must take two arguments and return a bool, etc etc

mathewmariani:
> Your doc example shows it, but it would be nice to have some documentation on the expected behavior of `comp()`. For example, saying it must take two arguments and return a bool, etc etc

Thanks for pointing that out. I just updated the docs and added a default `comp` functions

mhermier:
From my experience at using the language, I would recommend to change `sort()` to be a uni-line. Stub functions are only meant to provide default values or simply computed values, they should behave exactly like their parent and also return the same thing. While technically equivalent in your case, making it already there, express intent more, avoid possible future bugs and late fixing.

So as a rule of thumb; if you write a one liner, it is highly likely you want to return the computed value, so use the real one liner expression (that implicitly returns the computed value). I know, it is not really intuitive or does not scale producing changes, but the language is like this...

mathewmariani:
> From my experience at using the language, I would recommend to change `sort()` to be a uni-line. Stub functions are only meant to provide default values or simply computed values, they should behave exactly like their parent and also return the same thing. While technically equivalent in your case, making it already there, express intent more, avoid possible future bugs and late fixing.
> 
> So as a rule of thumb; if you write a one liner, it is highly likely you want to return the computed value, so use the real one liner expression (that implicitly returns the computed value). I know, it is not really intuitive or does not scale producing changes, but the language is like this...

Just to clarify you mean a uni-line (stue function) like so `sort() { sort {|a, b| a < b} }` or is there a Wren style that im missing?

mhermier:
Yes this one, it act as a stub with implicit returning.

test/core/list/sort.wren
```diff
@@ -0,0 +1,10 @@
+var list = [4, 1, 3, 2].sort()
+System.print([4, 1, 3, 2].sort()) // expect: [1, 2, 3, 4]
```
bjorn, line 2: Does this test pass? I don't see `List.sort()` returning any value, so I would expect this to print `null`. Did you mean to `return this` at the end of `sort`?

mathewmariani: With this [issue](https://github.com/wren-lang/wren/issues/801) open the tests don't pass.

mhermier: @bjorn one liner body/method/function does implicitly return.

mathewmariani: @mhermier the tests were returning null

bjorn: @mhermier I know, but since the `sort(_)` function was returning `null`, so did the one-line version of `sort()`.

src/vm/wren_core.wren
```diff
@@ -323,6 +323,43 @@ class List is Sequence {
     return other
   }
 
+  sort() {
+    sort {|a, b| a < b }
+    return this
```
bjorn, line 328: I think it's rather weird for `sort()` to return `this` but not `sort(comparer)`. I'd suggest changing this back to the one-line version and putting the `return this` in `sort(comparer)`.

test/core/list/sort.wren
mathewmariani: So `sort(_)` should return `this` and the one-liner `sort()` should be reverted back to a stub function?

mhermier:
Does the tests pass now ?

mathewmariani:
> Does the tests pass now ?

Not this the [stack corruption](https://github.com/wren-lang/wren/issues/761) issue

mhermier:
I think the test should be a bit more atomic to be more interesting.

mathewmariani:
I rebased from `main` and can confirm the tests are passing

bjorn:
@mathewmariani It seems you did a rebase followed by a merge. Please get rid of the merge for force-push just the rebased version of your branch.

mathewmariani:
> @mathewmariani It seems you did a rebase followed by a merge. Please get rid of the merge for force-push just the rebased version of your branch.

I believe that's it

mathewmariani:
Should I resolve all the comments, or leave that up to the commenters?

bjorn:
@mathewmariani I see most of my comments marked as "Outdated" and I see no way to resolve them (neither on ones that haven't been marked outdated), so I guess that's on you.

mathewmariani:
Anything else needed for this to be merged?

projects/xcode/wren_test.xcodeproj/project.xcworkspace/contents.xcworkspacedata
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<Workspace
```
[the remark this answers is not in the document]
bjorn: You resolved this comment but the file is still being added by the pull request.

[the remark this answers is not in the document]
mathewmariani: @bjorn should be done now

ChayimFriedman2:
Nice work. Three things, however:

 1. I think it's preferred to implement this in C. Wren prefers to implement the core in C (speed).
 2. You implemented a quick sort. Do we need a stable sort instead?
 3. Should we add a non in-place sort, too? (Maybe a method in `Sequence`). It can be just:
    ```wren
    class Sequence {
      // ...
      sort() { sort {|a, b| a < b } }
      sort(comparer) { toList.sort(comparer) }
    }
    ```

avivbeeri:
I don't think it's semantically appropriate to have a Sequence.sort method, because a sequence could in theory be infinite in length (and they can be lazily evaluated), so a sorting operation would be impossible or intractable.

ruby0x1:
Also a sort function in c runs into reentrancy (specifically, it's not possible when a user supplied compare function is involved)

ChayimFriedman2:
@ruby0x1 I think the reentrancy is about to be solved, am I wrong?

@avivbeeri I know that, be that still holds in Python for example, and it does have a `sorted()` function that operates on iterables.

mathewmariani:
@ChayimFriedman2 I implemented Quick Sort since it's the algorithm used in Lua's `table.sort`
@ruby0x1 is there anything left that I should be adding for this MR?

ruby0x1:
I'll look closely soon, busy with some other Wren stuff locally.

src/vm/wren_core.wren
```diff
@@ -323,6 +323,41 @@ class List is Sequence {
...
+  sort(comparer) {
+    if (!(comparer is Fn)) {
+      Fiber.abort("Comparer must be a function.")
+    }
+    quicksort_(0, count - 1, comparer)
+    return this
+  }
+
+  quicksort_(a, b, comparer) {
+    if (a < b) {
```
ruby0x1, line 337: @mathewmariani this is using `<` despite comparer being provided. 
I would expect `if(comparer.call(a, b)) {` to be the condition instead?

doc/site/modules/core/list.markdown
```diff
@@ -94,6 +94,25 @@ System.print(["a", "b", "c"].removeAt(1)) //> b
...
+### **sort**(), **sort**(comparer)
+
+Sorts the elements of a list in-place; altering the list.
+
+<pre class="snippet">
+var list = [4, 1, 3, 2].sort()
+System.print(list) //> [1, 2, 3, 4]
+</pre>
+
+A comparison function `comparer` can be provided to customise the element sorting. The comparison function must return a boolean value specifying the order in which elements should appear in the list. The default behaviour uses a `<` comparison.
```
ruby0x1, line 106: This doesn't explain the boolean result, or the nature of the compare function :) We'd ideally do both, I can adjust the documentation after merge though, just wanted to give feedback.

> The comparison function accepts two arguments, which are the two values to compare, and must return a boolean value to indicate if the first value is smaller than the second value. If the function returns true, the first argument will be before the second in the sorted results. 
> 
> A compare function like `{|value, other| true }` will always put `value` before `other`. The default compare function is `{|a, b| a < b }`.

src/vm/wren_core.wren
bjorn: Actually `a` and `b` are the start and end indexes of the range that's being sorted. Since the range is inclusive, the comparison is checking whether the range has more than 1 element.

Maybe there could be more intuitive names for these variables.

src/vm/wren_core.wren
```diff
@@ -323,6 +323,41 @@ class List is Sequence {
     return other
   }
 
+  sort() { sort {|a, b| a < b } }
```
ruby0x1, line 326: I suspect this might want to be `<=` for stable sort results, otherwise equal items will rearrange themselves on successive sorting, and won't be deterministic.

src/vm/wren_core.wren
ruby0x1: Oh right, I misread there. I agree with better naming.

src/vm/wren_core.wren
ChayimFriedman2: Quick sort isn't stable anyway.

ChayimFriedman2: https://stackoverflow.com/questions/13498213/quicksort-algorithm-stability

avivbeeri: It can be implemented in a manner that makes it stable, at the cost of memory consumption: https://www.geeksforgeeks.org/stable-quicksort/

As an alternative, we could use a merge sort, which has similar properties, plus stability. However, it would also have some memory consumption requirements.

ChayimFriedman2: I know that, but unless we use a stable sort (which I suggested earlier), it's useless to discuss about `<` vs. `<=`.

mhermier:
To help comprehension, could you replace 'a' and 'b' with something like
'i' and 'j' where iterators are expected.

ChayimFriedman2:
I think `start` and `end` would be better.

mathewmariani:
@ChayimFriedman2 @mhermier I'm leaning towards `i` and `j` since that's what you find in most books. I also like more compact code.

src/vm/wren_core.wren
mathewmariani: Is there a specific reason we want a stable sort? My idea for using a non-stable quick-sort was that it's highly memory efficient which is perfect for, in my opinion, virtual memory environments like the WrenVM. It might have a worst-case scenario of running at O(n^2) but on average it should run O(n logn).

If a more efficient algorithm is needed it can be left to the user.

src/vm/wren_core.wren
mathewmariani: I was thinking `i` and `j` since that's what most algorithms & data-structure books might use. But, `start` and `end` can be used as well.

src/vm/wren_core.wren
avivbeeri: I think we can get away with it being unstable, so long as we document it clearly. From an embeddability point of view, the memory usage argument is quite compelling, in my opinion.

mathewmariani: @avivbeeri Thank you! and yes, I will have to redo the documentation

ruby0x1:
@mathewmariani code is for humans, not machines, legibility > compactness. especially for wren, which strives to help people learn and understand, making the code clearer is part of that goal. In my code I used `low` `high` but `start` and `end` is also clear enough. I would prefer not to use single letter variables here.

ruby0x1:
Thanks @mathewmariani , I'll probably tweak some things after merging, but looks good from here.

ruby0x1 (approved):

PR #803, opened 2020-08-30
make sure Sequence.any and Sequence.all always return boolean values

totallyRonja:
previous behaviour was returning the first falsy value or the last value for all / opposite for any


mhermier:
If you want to go this way, you can simplify more removing the result
variable. And don't forget to update the .inc.
Previous behavior was interesting as it was combining searching and mapping
the result for *free*. This need behavior, seems a little bit incomplete,
returning an iterator to the first falling iterator is also an interesting
behavior.

totallyRonja:
the new behaviour is more predictable in my eyes and mirrors more closely how its documented in the docs
and yeah, I moved the variable in the for loop since its not needed afterwards anymore, but I believe its more readable to have a intermediate variable here than to squish it into the if condition

mhermier:
Performance wise I think, it should be more efficient to remove the
variable completely. Since the compiler does not do any analysis, in fact
having the variable inside the loop force variable creation/destruction on
each loop. So either let the variable outside or remove it completely.

totallyRonja:
both those ideas make the code slightly less readable and less consistent with the rest of the code
String.trim_ also has a variable in a loop iteration thats only used once and is stored for convenience

PR #846, opened 2020-11-24
Add .editorconfig

billyquith:
- 2 space indents for most things. Defaults tend to be 4.


joshgoebel:
We should also standardize:

- trim_trailing_whitespace
- end_of_line
- insert_final_newline

billyquith:
I'd suggest you merge this and modify it accordingly.

PR #850, opened 2020-11-25
Fix hash mask

billyquith:
The hashing function which uses Thomas Wang's algorithm masks off the top bit, which isn't in Wang's algorithm. Masking the bit off effectively halves the hash space, which is already masked from 64 to 32 bits. Having a larger hash just reduces the chance of collisions.

Also fixed the hyperlink to the article as site link disappeared.

### DIFF
--- a/src/vm/wren_value.c
+++ b/src/vm/wren_value.c
@@ -360,9 +360,8 @@ static inline uint32_t hashBits(uint64_t hash)
 {
   // From v8's ComputeLongHash() which in turn cites:
   // Thomas Wang, Integer Hash Functions.
-  // http://www.concentric.net/~Ttwang/tech/inthash.htm
+  // https://web.archive.org/web/20071223173210/http://www.concentric.net/~Ttwang/tech/inthash.htm
   // V8 version masks off top bit, whilst Wang's version does not:
-  // https://gist.github.com/badboy/6267743
   hash = ~hash + (hash << 18);  // hash = (hash << 18) - hash - 1;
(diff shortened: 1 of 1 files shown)


joshgoebel:
Why would V8 do this? Do you know?

billyquith:
> Why would V8 do this? Do you know?

I don't know. I think the source is available but I haven't checked for comments. At a wild guess I'd say they cast it to an `int` a lot and didn't want a negative number. What other reasons are there?

joshgoebel:
Where did you get this information about V8 masking the top bit - just from our own line 361?   I can't seem to confirm...

- Deno: https://denolib.github.io/v8-docs/functional_8cc_source.html
- Chromium: https://source.chromium.org/chromium/chromium/src/+/master:v8/src/base/functional.cc;drc=e51dd5c377fd47393a171f6bdcd7c1a6a9a609c5;l=97
- GitHub V8 mirror: https://github.com/v8/v8/blob/master/src/base/functional.cc

billyquith:
> Where did you get this information about V8 masking the top bit - just from our own line 361? I can't seem to confirm...

I'm sorry Josh, I can't remember. I did some searching at the time. It may be from the Wren comments. I guess if V8 does not mask, all the more reason to not mask.

joshgoebel:
https://github.com/wren-lang/wren/commit/2a1499b04bc65853fdf71db2126e716168d4c5ae

Look's like Bob's work in 2019.  Any idea/memory on why we're masking that high bit in the first place?

CC @munificent

Issue #893, opened 2020-12-27
Odd behaviour of Sequence.all and Sequence.any


clsource:
There is something odd in Sequence.all and Sequence.any.
https://github.com/wren-lang/wren/blob/3d5e68fc01e5c5a74f8fd3bdd02cea060e2d2444/src/vm/wren_core.wren#L12
The docs says https://wren.io/modules/core/sequence.html

> Iterates over the sequence, passing each element to the function predicate. *If it returns something false, stops iterating and returns the value*. Otherwise, returns true.

```js
  all(f) {
    var result = true
    for (element in this) {
      result = f.call(element)
      if (!result) return result
    }
    return result
  }
```

By "If it returns something false, stops iterating and returns the value" should be `if (!result) return element` instead of result?

mhermier:
The documentation is lying. `all` or `any` are tests and we expect to
return a truthy or falsy value. `element` obviously doesn't match its
criteria.

clsource:
So the behaviour is correct, but the docs need improvement

mhermier:
Yes. The behavior you describe is most likely related to something like a `find` method (which does not exist and would need a PR) that would return an iterator on the `Sequence`.

clsource:
@ruby0x1 told that  PR #803 is related.

Would be this wording more inline with the actual behaviour?

## all(predicate)

Tests whether all the elements in the sequence pass the predicate.

Iterates over the sequence, passing each element to the function predicate. If it returns something false, stops iterating and returns false. Otherwise, returns true.



    Back to Modules

core classes

    Bool
    Class
    Fiber
    Fn
    List
    Map
    Null
    Num
    Object
    Range
    Sequence
    String
    System

Sequence Class

An abstract base class for any iterable object. Any class that implements the core iterator protocol can extend this to get a number of helpful methods.
Methods #
all(predicate) #

Tests whether all the elements in the sequence pass the predicate.

Iterates over the sequence, passing each element to the function predicate. If it returns something false, stops iterating and returns the value. Otherwise, returns true.

System.print([1, 2, 3].all {|n| n > 2}) //> false
System.print([1, 2, 3].all {|n| n < 4}) //> true

## any(predicate)

Tests whether any element in the sequence passes the predicate.

Iterates over the sequence, passing each element to the function predicate. If it returns something true, stops iterating and returns true. Otherwise, returns false.


## find
Maybe the `find` method would be similar to

```js
[...]

PR #895, opened 2020-12-27
improved wording on sequence all and any method docs

clsource:
Fixes https://github.com/wren-lang/wren/issues/893


ChayimFriedman2:
Why did you closed that?

clsource:
> Even if more more valid than before, the statement is technically wrong, the predicate only needs to return truthy/falsy values. This was made so that predicate could also act as/work with a `map` function.
> 
> Don't add `find` documentation, I handled it in the push request I issued.

I personally don't know how to document this behaviour.

PR #898, opened 2020-12-28
wren/core: Add `Sequence::find`.

mhermier:
A followup of #893 about adding `Sequence::find`.


mhermier:
Fixed an issue where it was always iterated first ignoring the begin of the iterator.

PR #908, opened 2021-01-09
Added USEM64 flag in Makefiles for aarch64.

yuppox:
This deals with the -m64 flag problem on aarch64.


ruby0x1:
The make files are generated via [premake](https://github.com/premake/premake-core/wiki) and are not edited directly. See `projects/premake/` for the source premake script, and you can use `util/generate_projects.py` to generate all the necessary outputs. (That assumes you've put the premake binary in the projects/premake folder).

PR #931, opened 2021-03-02
doc/site/functions Fn.new() clarification [minor]

cxw42:
Thank you for Wren, which I just recently discovered (via _Crafting Interpreters_)!  This is my first Wren contribution.  I noticed a typo in the docs for Fn.new() and offer this fix.  Thanks for considering this PR!


ruby0x1:
Thanks!

PR #936, opened 2021-03-08
rename some test functions that clash with libc

alisonatwork:
While attempting to get Wren to compile under Cosmopolitan libc (see https://github.com/jart/cosmopolitan/issues/104), I discovered two test functions whose names clash with libc. This is not a problem when compiling against a standard libc, but with Cosmopolitan there is a "super header" that is included with `-include cosmopolitan.h` where all the functions are defined. This stops some tests from compiling, since they also define `write` and `remove`. This PR just renames those functions.

This is the only change to the Wren source code now required to get it building and all the tests passing under Cosmopolitan libc.

Something that might be interesting is adding a Cosmopolitan Premake action so that we can generate makefiles that just work without any tweaks, but that's something for the future, I think.

### DIFF
--- a/test/api/maps.c
+++ b/test/api/maps.c
@@ -49,7 +49,7 @@ static void insert(WrenVM* vm)
   wrenSetMapValue(vm, 0, 1, 2);
 }
 
-static void removeWren(WrenVM* vm)
+static void removeKey(WrenVM* vm)
 {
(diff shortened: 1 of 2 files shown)


ruby0x1:
I tweaked the names a little, let me know if that still works out?
Otherwise, I'll keep an eye on that issue as well. Thanks!

PR #946, opened 2021-03-31
Add seperators for digits

CrazyInfin8:
Allows adding underscores as seperators for digits. As `strtoll` doesn't support separators I replaced it with a custom function I temporarily called `wrenStrToUL`. Currently this only works with hexadecimals, octals, and binary numbers but I will try to eventually add a replacement to `strtod` to support decimals as well (if this even seems like a good idea).

an example of what it looks like so far:
```js
var hex = 0xAB_CD_EF
var oct = 0o0123_4567
var bin = 0b1010_1010_1010

System.print(hex) // 11259375
System.print(oct) // 342391
System.print(bin) // 2730
```


ruby0x1:
@CrazyInfin8 it's not ideal to include clauses in a user facing codebase like that. I wouldn't include code from anything that imposed that.

mhermier:
Licencing should not be a problem here, the code is quite trivial. The other one is a direct copy of a file with permissive copyright with derivative code that is probably subject to the Python licence, not the same issue.

CrazyInfin8:
Started to replace `strtod` a bit. There are still some kinks and issues to work out so this is not yet ready. 
Two of the tests have not passed.
- `test/language/number/literal_too_large.wren` fails as there is no checking if we make a number too big (yet) 
- `test/core/number/fraction.wren` fails because line 18 and 19 print 1 and -1 instead of 0 and -0.

For the first issue, I'll have to do more research on doubles/floats to determine what is a safe maximum to limit the number to. for the second issue, I'm not exactly sure what is happening here or I might be a little unfamiliar with how floats/doubles work.

CrazyInfin8:
`wrenStrToD` now passes all tests on Linux. Windows still has some issues to solve (at least when using MinGW).

image
-----
In `test/core/number/from_string.wren`, these lines fail:
```js
System.print(Num.fromString("12.34") == 12.34) // expect: true
System.print(Num.fromString("-0.0001") == -0.0001) // expect: true
```
I decided to print out those numbers to see how far off `wrenStrToD` is from windows which looks something like this:
```js
System.print(Num.fromString("12.34")) // 12.34
System.print(Num.fromString("-0.0001")) // -0.0001
System.print(12.34) // 12.34
System.print(-0.0001) // -0.0001
```
So the difference is probably due to a floating point error. This might be able to be fixed by replacing the call to `strtod` with `wrenStrToD` (for function `Num.fromString`) but `strtod` can also do hexadecimals whereas `wrenStrToD` can't (at least not yet)
-----
In `test/core/number/fraction.wren`, these lines fail:
```js
System.print((-123).fraction)   // expect: -0
System.print((-0).fraction)     // expect: -0
```
because Windows (or MinGW) doesn't do negative zeroes (which already failed on windows before so this isn't a new issue)
These next few lines also fail:
```js
System.print((1.000000000000000000000000000000000000000001).fraction)  // expect: 0
System.print((-1.000000000000000000000000000000000000000001).fraction)  // expect: -0
```
Not entirely sure why the number `1.000000000000000000000000000000000000000001` becomes `1.7763568394003e-15` but I'll try and throw this in a windows debugger later on and try to find out why.

CrazyInfin8:
Another question that's rising in my head is whether `wrenStrToD` and `wrenStrToLL` (renamed `wrenStrToUL` after having some issues on Windows) should have the same signatures as `strtod` and `strtoll`. Meaning, should they take a char pointer at the start and set an end char pointer when finished like `strtod` or should I pass in `Parser*` like I do with `wrenStrToLL`? doing it the first way means it is more general purpose since it can operate as a drop-in replacement but doing it the second way means it is a little simpler since we don't need to pre-check for a number token when parsing numbers during compilation.

mhermier:
I would leave the old C naming convention and use something like "wrenStrToU64". Ideally, signature should be something like:
```C
errno_t wrenStrToTYPE(type *value, const char *begin, const char **end, const void *locale);
```
`end` is optional, and could be also used to pass the end of text, and retrieve where the parsing ended.
`locale` should be NULL or a mockup pointer for a Wren locale. This is just in case we need locale support later.
That way parsing and scanning could be done at once, since we could allow broken values to be parsed, till we don't think it is part of the number, and return using `end`.
Also this should leave the "wren_compiler.c" to "wren_math.c", "wren_string.c" or "wren_utils.h"...

CrazyInfin8:
I have created a custom version of `strtod` in a new repo [here](https://github.com/CrazyInfin8/StrToD). This time it feels a lot more stable and spits the results I want from my own testing. The actual tests in the repo compare consistency between `strtod` and `c8_StrToD` but some intentional designs make it fail a few of those tests.

I have also vastly rewritten the strtoll to handle radix settings using prefixs like `0b`, `0o` and just `0`. it is also in it's own repo [here](https://github.com/CrazyInfin8/StrToLL). As people discussed dissatisfaction with treating leading zeros as octals, I put defines to change the behavior of a leading zero to octal or decimal. the comment located [here](https://github.com/wren-lang/wren/pull/945#issuecomment-810851761) intrigued me into trying out smalltalk style radix literals so I also tried experimenting with those as well. turns out they aren't too hard to implement so I guess I decided to add them (also optionally enabled using a define).

I'm still working on them but for now they should be mostly ready to add to wren for testing. Might try to fix some inconsistencies with my `strtoll` function and may try to make `strtod` more powerful later on.

CrazyInfin8:
As I'm talking about adding these functions to the VM, would it be better to close this PR and open a new one with the functions specified in the above comment for neatness or should I add the functions to this PR and mark as ready for review?

I know a few changes have recently been committed to the main branch already so wasn't sure if adding these new changes to this PR would cause them to be ignored due to it being on an outdated branch.

still pretty new to contributing to other projects here so bear with me. 😅

PR #947, opened 2021-03-31
Port Python's strtod implementation

fermian:
This pull request includes a port of Python's strtod implementation for parsing floating point numbers in Wren's source code parser. Using a separate strtod function should remove locale dependency from using the one provided by the standard library.

### DIFF
--- a/projects/make.bsd/wren.make
+++ b/projects/make.bsd/wren.make
@@ -87,8 +87,6 @@ ALL_CFLAGS += $(CFLAGS) $(ALL_CPPFLAGS) -g -std=c99
 ALL_CXXFLAGS += $(CXXFLAGS) $(ALL_CPPFLAGS) -g
 ALL_LDFLAGS += $(LDFLAGS)
 
-else
-  $(error "invalid configuration $(config)")
 endif
 
 # Per File Configurations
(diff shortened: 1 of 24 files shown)


mhermier:
Could someone with more knowledge than me verify licencing of it. The python licence seems troublesome too add to the project.

clsource:
https://docs.python.org/3/license.html

> 2. Subject to the terms and conditions of this License Agreement, PSF hereby
   grants Licensee a nonexclusive, royalty-free, world-wide license to reproduce,
   analyze, test, perform and/or display publicly, prepare derivative works,
   distribute, and otherwise use Python 3.9.2 alone or in any derivative
   version, provided, however, that PSF's License Agreement and PSF's notice of
   copyright, i.e., "Copyright © 2001-2021 Python Software Foundation; All Rights
   Reserved" are retained in Python 3.9.2 alone or in any derivative version
   prepared by Licensee.

> 3. In the event Licensee prepares a derivative work that is based on or
   incorporates Python 3.9.2 or any part thereof, and wants to make the
   derivative work available to others as provided herein, then Licensee hereby
   agrees to include in any such work a brief summary of the changes made to Python
   3.9.2.

Im no lawyer but it seems it would need adding a copyright notice

mhermier:
The real question is will that contaminate the MIT Licencing of the project ?

ruby0x1:
fwiw if we do go down this road (it is not determined yet) there are also MIT implementations available https://git.musl-libc.org/cgit/musl/tree/src/stdlib

PureFox48:
Personally, if I'm looking for something to copy/translate, I don't even consider anything that isn't either MIT licensed (as Wren is itself) or in the public domain. Licensing is just too much of a can of worms to do otherwise in my view.

mhermier:
I will probably hate myself later for proposing that, but maybe we should
implement it ourself, by augmenting our own functions. That way we could do
token scanning and parsing in one pass...

p0nce:
locale dependency in `strtod` in Wren caused problems for us in `wren-port`. I would recommend replacing double parsing in Wren by something indenpendent from C stdlib locale.

mhermier:
Maybe we can conditionally support `strtod_l`. It seems supported on a a fair amount of systems now with some kind of limitations on some like `#define _GNU_SOURCE` on glibc. Dunno if it is a no go for the MIT licence though...

PR #952, opened 2021-04-02
[Experiment] Added do end alternative to braces

clsource:
This is just an experiment on having an alternative to left and right braces.
- This was not hard to achieve but it required two additional keywords to be added
- Lots of tests fails because of the "end" to be a keyword now.

```js

if (true) do
  System.print("hello") // expect: hello
end

Fn.new do | hello | System.print(hello) end.call("world") // expect: world

class Test do
  static hello { "Wren" }
  static print(){ System.print(hello) }
end

Test.print() // expect: Wren

do
  System.print("closure") // expect: closure
end
```


fermian:
I wonder if we could do Ruby-style blocks with an implicit "do" keyword 🤔 

```rb
if true
	puts "Hello World"
end
```

ruby0x1:
In Wren this part of the language is well defined and at the moment there is no interest in adding alternative syntax for this. 

Thanks for the suggestion!

PR #954, opened 2021-04-03
fix possibility of having no reallocate function

Orcolom:
currently you cant provide a config without also providing a `reallocateFn`. Most cases `reallocateFn` gets filled by `wrenInitConfiguration`. This just ensure a reallocate function always exists

### DIFF
--- a/src/vm/wren_vm.c
+++ b/src/vm/wren_vm.c
@@ -58,13 +58,11 @@ WrenVM* wrenNewVM(WrenConfiguration* config)
   if (config != NULL) {
     
     // ensure config has a reallocate function
-    if (config->reallocateFn != NULL) {
-      reallocate = config->reallocateFn;
-    }
-    else {
+    if (config->reallocateFn == NULL) {
       config->reallocateFn = defaultReallocate;
     }
 
+    reallocate = config->reallocateFn;
     userData = config->userData;
   }
   
--- a/src/vm/wren_vm.c
+++ b/src/vm/wren_vm.c
@@ -56,7 +56,15 @@ WrenVM* wrenNewVM(WrenConfiguration* config)
   WrenReallocateFn reallocate = defaultReallocate;
   void* userData = NULL;
   if (config != NULL) {
-    reallocate = config->reallocateFn;
+    
+    // ensure config has a reallocate function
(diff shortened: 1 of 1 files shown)


src/vm/wren_vm.c
```diff
@@ -56,7 +56,15 @@ WrenVM* wrenNewVM(WrenConfiguration* config)
...
   if (config != NULL) {
-    reallocate = config->reallocateFn;
+    
+    // ensure config has a reallocate function
+    if (config->reallocateFn != NULL) {
+      reallocate = config->reallocateFn;
+    }
+    else {
+      config->reallocateFn = defaultReallocate;
+    }
```
sveljko, line 66: I find the code somewhat strange, requiring looking in the surrounding code to make sure it's doing the right thing. I believe a slight refactoring would make it easier to reason about:
```suggestion
    if (NULL == config->reallocateFn) {
      config->reallocateFn = defaultReallocate;
    }
    reallocate = config->reallocateFn;
```

Orcolom: sure, reformatted but with NULL on the other side

ruby0x1:
Thanks @Orcolom for the PR. 

My only thought is that it now has side effects on the user config struct. Not that the user usually holds onto the struct, or that the change from a null function pointer to a default function pointer is that surprising... just something that is interesting.

An alternative that doesn't modify the user value, but instead the vm owned one after the copy, but not sure it's relevant enough in this particular case.

```c++
  WrenReallocateFn reallocate = defaultReallocate;
  void* userData = NULL;
  if (config != NULL) {
    userData = config->userData;
    reallocate = config->reallocateFn ? config->reallocateFn : defaultReallocate;
  }
  
  WrenVM* vm = (WrenVM*)reallocate(NULL, sizeof(*vm), userData);
  memset(vm, 0, sizeof(WrenVM));

  // Copy the configuration if given one.
  if (config != NULL)
  {
    memcpy(&vm->config, config, sizeof(WrenConfiguration));
    vm->config.reallocateFn = reallocate;
  }```

Orcolom:
yeah, don't think this would give unexpected behavior. but probably also a good idea to not modify user's provided struct.
didn't know you could do ternary in c. removes the need for the awkward if else

PR #958, opened 2021-04-04
made wren's version number accessable via function

Orcolom:
this way its accessible in dll via WREN_API


mhermier:
Not that it should matter some day soon, but I would prefer that it returns an uint64_t instead of a generic int. Otherwise trivially legitimate.

ruby0x1:
Thanks @Orcolom

PR #963, opened 2021-04-06
Add an additional escape sequence \e.

PureFox48:
This is an idea I had when discussing #945 to avoid the need to support octal escape sequences.

Probably, the only common use for these nowadays is to encode the ESC character (0o33) itself in ANSI escape codes and adding an escape sequence for this will make life a little easier.

It's not a new idea as gcc, clang and tcc already support it as a non-standard extension to their C compilers.

I haven't added a test as the other escape sequences which represent control characters don't seem to have any.

### DIFF
--- a/doc/site/values.markdown
+++ b/doc/site/values.markdown
@@ -69,6 +69,7 @@ A handful of escape characters are supported:
 "\%" // A percent sign.
 "\a" // Alarm beep. (Who uses this?)
 "\b" // Backspace.
+"\e" // ESC character.
 "\f" // Formfeed.
 "\n" // Newline.
 "\r" // Carriage return.
--- a/src/vm/wren_compiler.c
+++ b/src/vm/wren_compiler.c
@@ -972,6 +972,7 @@ static void readString(Parser* parser)
         case '0':  wrenByteBufferWrite(parser->vm, &string, '\0'); break;
         case 'a':  wrenByteBufferWrite(parser->vm, &string, '\a'); break;
         case 'b':  wrenByteBufferWrite(parser->vm, &string, '\b'); break;
+        case 'e':  wrenByteBufferWrite(parser->vm, &string, '\33'); break;
         case 'f':  wrenByteBufferWrite(parser->vm, &string, '\f'); break;
         case 'n':  wrenByteBufferWrite(parser->vm, &string, '\n'); break;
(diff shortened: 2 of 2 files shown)


ruby0x1:
@PureFox48 is it `\033` or `\33`, I suspect you're missing a preceding 0 if using octal. 

One alternative is adding `\x` instead, then ESC become `\x1b` and it's more general purpose than just \e.

mhermier:
@ruby0x1 the front 0 is optional, the escapes where always octal in all known conventions/implementations.

PureFox48:
Yes, @mhermier is correct that the leading `0` is optional since octal escapes can be either 1, 2 or 3 characters following the `/`.

Wikipedia has a good summary [here](https://en.wikipedia.org/wiki/Escape_sequences_in_C).

Although `\x1b` works fine, the reason I wanted to add `\e` is because you don't then have to remember the code for ESC when writing an ANSI escape code and (apart from `\0` which we already have) it's the only common use for octal escapes these days. We wouldn't therefore need to support them as well as octal literals.

mhermier:
Adding `\x` will need to be done at some point to allow to ease creation of byte arrays, but it is not critical yet (till someone request for it).

PureFox48:
We already have `\x`. See [escaping in the docs](https://wren.io/values.html) towards the end of the secton.

ruby0x1:
Ah ok, I've somehow only ever noticed with a leading 0 but looked it up. 

Also I forgot \x was there below. I'll make that table a bit clearer.

Noting that the [Wren page](https://wren.io/values.html#escaping) also has this line, which hints that escapes aren't too precious or strictly minimal. 
```js
"\a" // Alarm beep. (Who uses this?)
```

mhermier:
Peopole who do console applications, probably wren-cli use it... was not able to run it since a while because of local changes...

PureFox48:
Yes, I sometimes use `\a` myself as it's currently the only way to get a tweet out of Wren without using a game engine :)

ruby0x1:
Minor note: the new escape wasn't documented in this PR but I added it. 

Thanks for the PR, I think I've used escape stuff enough in wren for unit tests, command line tools and logging that it's not unreasonable to making it cleaner considering the low impact of the change.

PureFox48:
Excellent, thanks :)

PR #965, opened 2021-04-08
value: Rename `validateKeyType`to `wrenMapIsValidKey`.

mhermier:
Simplify and move where it is more logical in the process.


mhermier:
Updated to build in debug mode.

PR #976, opened 2021-04-16
Documents 'return from module'.

### LINKED ISSUE #974: Undocumented behavior of the return keyword.
One thing I've noticed recently is that, if you're in the middle of module-level code (i.e. code which is neither inside a function or a method), then the `return` keyword will jump to the end of that module. For example:

```javascript
/* mod_ret1.wren */

for (i in 1..2) {
    if (i == 2) return
    System.print(i) // prints 1 but not 2
}

System.print(3) // doesn't print

/* output:
1
*/
```
However, it only jumps to the end of the module it's in, not the end of the script:
```javascript
/* mod_ret2.wren */

import "./mod_ret1"
System.print(4) // prints 4

/* output:
1
4
*/
```
So it seems to work as if the module-level code is wrapped inside a function and then that function is automatically invoked.

Does anybody know whether this behavior is intended because, if it is, I couldn't find it documented anywhere.

If it is intended then:

1.  It could be useful for exiting a module and possibly the entire script prematurely without the need for ```Fiber.abort```.

2.  I could try and do a PR to document it somewhere.

PureFox48:
Fixes #974.
 
Although @mhermier asked for a test to be added, I'm not sure how we can do one just now as there is no way to capture any returned value and print it.

### DIFF
--- a/doc/site/modularity.markdown
+++ b/doc/site/modularity.markdown
@@ -308,7 +308,7 @@ where you need them.
 
 ## Exiting a module early
 
-Although the return statement is normally used to exit from a [method](classes.html#methods) or a [function](functions.html), it can also be used from a module's top-level code to exit the module. For example, if the script consists of a single module, this code would exit the module (and therefore the script) early:
+Although the `return` statement is normally used to exit from a [method](classes.html#methods) or a [function](functions.html), it can also be used from a module's top-level code to exit the module. For example, if the script consists of a single module, this code would exit the module (and therefore the script) early:
 
 <pre class="snippet">
 for (i in 1..2) {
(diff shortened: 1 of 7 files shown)


mhermier:
Since we can't check for returned value, we can still check for execution flow. Basically you need to add 4 tests for the case we return a value or not, from main or imported module. Basically the tests are (modulo the required adaptation for the scenarios):
```javascript
System.print("foo") // expect: foo
return
System.print("bar")
```

PureFox48:
OK, thanks, I've done all that.

mhermier:
Good enough for me. Would have preferred more consolidated diffs, but this can do.

PR #983, opened 2021-04-20
Add note about closures capturing their surrounding scope by reference

VayuDev:
We had a discussion on Discord related to this commit that was sparked by clsource. He provided the following code:
```lua
var printNumbers = Fn.new {
    var numbersToPrint = []
    var i = 0
    while (i < 10) {
        numbersToPrint.add(Fn.new {
            System.print(i)
        })
        i = i + 1
    }

    return numbersToPrint
}

var numbers = printNumbers.call()

// Prints only 10
numbers.each {|fn| fn.call()}
```

Which (rather surprisingly) only prints 10. To get the expected behavior, you need to copy the variable i in each iteration and use the copied variable in the closure. This patch adds a note in the documentation so that new users aren't surprised by this.

### DIFF
--- a/doc/site/functions.markdown
+++ b/doc/site/functions.markdown
@@ -97,6 +97,11 @@ System.print(counter.call()) //> 2
 System.print(counter.call()) //> 3
 </pre>
 
+Variables outside the scope of the closure are captured by reference and not by
+copy, so writing to a captured variable will mutate it in the outer scope as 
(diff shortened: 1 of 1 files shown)


PureFox48:
In the program comment, I suspect you meant to say that it "Prints ten 10's", one for each function called. 

I don't think this behavior will surprise people who've used closures in other languages but I agree that it's a good idea to document it anyway.

PR #986, opened 2021-04-23
wren/language: Add unary `+` operator.

mhermier:
For the sake of completeness and to ease pretty formatting, allow numbers to be be prefixed with `+`.

### DIFF
--- a/doc/site/modules/core/num.markdown
+++ b/doc/site/modules/core/num.markdown
@@ -205,6 +205,18 @@ System.print(1.5.truncate)    //> 1
 System.print((-3.2).truncate) //> -3
 </pre>
 
+### **+** operator
+
+Returns the identity of the number.
+
+<pre class="snippet">
+var a = +123
+System.print(a) //> 123
+
+var b = -123
+System.print(+b) //> -123
+</pre>
+
 ### **-** operator
 
 Negates the number.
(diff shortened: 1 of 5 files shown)


PureFox48:
Well, it's always surprised me that you can't prefix numbers with `+` anyway.

However, it wouldn't necessarily help with print formatting because you'd need to have some way to tell methods such as `System.print` to print the `+` which currently we don't have.

mhermier:
Well, I meant it would help to pretty formatting numbers in code code, not as in print formatting. Sorry for confusion.
It is meant to help math people to transfer tables from a source where numbers would always get signed (like scientific or financial stuff), or simply avoid to trigger some OCD is some particular situations XD

PureFox48:
Yes, I see what you mean now.

I'd support the change.

tr8dr:
I too want the unary + operator (and produced a separate duplicate pull for it).   This pull has been sitting for quite a while.   Any thoughts on resolving?

nikeedev:
Do the changes from the commit work @mhermier?

tr8dr:
I made exactly the same changes in the interpreter for my pull (and they work).   @mhermier's pull is a better one than the one I submitted in that @mhermier created test cases and adjusted the documentation.

PR #996, opened 2021-05-01
Added missing **'s to Num.toString

Orcolom:


RobLoach:
Other representations of this don't emphasis the heading....
 https://github.com/wren-lang/wren/blob/a33832efcfd86c5ac7c91455373b35924b6c1c73/doc/site/modules/core/num.markdown#tostring

ChayimFriedman2:
It depends on the CSS. The https://wren.io site does:
image
image

But it's on `Bool`'s page! Please fix this.

RobLoach (approved):
Ah, I see. Thanks!

PR #999, opened 2021-05-02
Add octal escape sequences

CrazyInfin8:
This allows the use of octal escape sequences in string literals. Octal escapes are represented with a back slash followed by at least 1 and at most 3 octal digits. Octal escape sequences also cannot exceed 255 (the max value for an unsigned byte).

The two most common use cases for octal escapes are null bytes represented as`\0`, and ANSI escape sequences represented as `\033` or `\33` which is functionally similar to `\e`.

### DIFF
--- a/src/vm/wren_compiler.c
+++ b/src/vm/wren_compiler.c
@@ -874,6 +874,28 @@ static int readHexEscape(Parser* parser, int digits, const char* description)
   return value;
 }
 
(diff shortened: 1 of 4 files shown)


PureFox48:
Well, now that we have both `\0` and `\e`, I don't think that supporting octal escapes is worthwhile, particularly as - unlike the existing escapes - they would be variable length.

I just can't think of any other use cases for them in modern programming.

PR #1000, opened 2021-05-02
Prevents the compiler from printing unprintable chars for escape sequences

CrazyInfin8:
Fixed an issue where invalid escape sequences can print bytes that were not meant to be printed, thus sometimes corrupting error messages.
image
After fix...
image


CrazyInfin8 (commented):
line 1028 may potentially conflict with #999 if it were to get accepted.

PR #1007, opened 2021-05-04
docs; fix example with 'continue'

tsukanov-as:
This is a minor fix in the documentation.


joshgoebel (approved):

ruby0x1:
nice catch, thanks for the PR!

Issue #994, opened 2021-04-28
Request: Support fields on foreign classes


joshgoebel:
**Is your request related to a specific problem you're having?**

Yes.  I'm work on networking for Wren CLI and the server object needs to hold some state on the C side.  It also has some Wren state that is of course stored in fields.  Evidently this is currently impossible so you're forced to create two classes, the Wren class and then the C foreign class - and then the Wren class must wrap the foreign class.  This is because foreign classes currently do not allow fields.  So you end up with something like:

```js
class TCPServer {
    construct new(ip, port) {
        // ...
        _uv = UVServer.new(ip, port, this)
    }
    listen() { _uv.listen_() }
    stop() { _uv.stop_() }
}

foreign class UVServer {
    construct new(ip,port,serverWren) {}
    foreign listen_()
    foreign stop_()
}
```

This starts to feel very unwieldy.  TCPServer now has to keep a reference to UVServer and vice versa (so some things can possibly be passed up the stack).  And many functions (anything that requires the lower-level C data) have to be proxied down from the non-foreign class.

Perhaps UVServer will get more complex but so far I feel like I'm asked to create a whole second class and all this proxying complexity foley to attach a some small amount of data to each Wren instance on the C side.  Am I missing something?  Is there a better pattern here I'm not aware of?

**The solution you'd prefer / feature you'd like to see added...**

It would be nice if this restriction was removed - if a class could both be foreign AND have fields.  A foreign class's RAM allocation might this look like:

```
[
  WREN_DATA (fields, etc),
  C_DATA (ram allocated to C)
]
```

Of course this wouldn't have to be contiguous memory, that would be an implementation detail.  

**Any alternative solutions you considered...**

Passing a pointer from C back into Wren as a number.  This would of course be problematic for GC but if C ALSO holds onto the object in question with a handle - this would ensure it's never GCed.  This pushes some memory management back into Wren as then you'd need to have a `destroy` callback that passed the number back to C again where the memory could deallocated and then the handle returned via `wrenReleaseHandle`.

I haven't tried this, it's just what popped into my head.

**Additional context...**

None.

mhermier:
I have this problem in mind since a while, tried various solutions that where not enougth generic for me. The solution without being completely generic is to fusion `ObjForeign` and `ObjInstance` structures, and change the API accordingly...

joshgoebel:
> to fusion ObjForeign and ObjInstance structures, and change the API accordingly

[...]

PR #1008, opened 2021-05-05
Adding fields support to foreign class #994 by adding ObjMemorySegment

mhermier:
Before it can be applied, this patch set requires to solve the question of the behavior when inheriting from a foreign, when no constructor/destructor are provided, while the parent class had some.

This patch set allow foreign class to have member variables with some restrictions. It still must inherit from a regular class without member variables (should we check that there is no static member variables ?) or from another foreign class.

It does that by adding an `ObjMemorySegment` which is the fusion of `ObjForeign` and `ObjInstance`. The name comes from the fact that such object can technically be used hold any object the VM can produce, and therefore represent a memory segment of the VM.

### DIFF
--- a/src/vm/wren_compiler.c
+++ b/src/vm/wren_compiler.c
@@ -2879,7 +2879,6 @@ static int getByteCountForArguments(const uint8_t* bytecode,
     case CODE_LOAD_LOCAL_8:
     case CODE_CONSTRUCT:
     case CODE_FOREIGN_CONSTRUCT:
-    case CODE_FOREIGN_CLASS:
     case CODE_END_MODULE:
     case CODE_END_CLASS:
       return 0;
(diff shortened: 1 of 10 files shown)


PureFox48:
> should we check that there is no static member variables ?

On the face of it, there seems no need as static methods are not inherited anyway.

mhermier:
True, but you end up in a situation where declaring a generic constructor in wren, results in the in a crash if you call a method from the super class expecting the defined constructor memory. And I don't see a way to `ASSERT` on this properly.

PureFox48:
Just check there are no instance or static fields then.

As it's a technical requirement that the foreign class needs to inherit from a regular class, the latter may just be invented for this sole purpose and not need any kind of state anyway.

mhermier:
I think we don't understand each other, the problem is in the subclass possibly not providing foreign user data to the parent class because of the default behavior when NULL is passed.

PureFox48:
I can see there is problem with static fields now you've explained it.

However, I'm try to look at it from a practical point of view.

In other words do you think it would be a major impediment if the regular super-class couldn't have either instance or static fields?

mhermier:
The problem lies in the fact that the constructor cannot be called in a re-entrant manner, leading to an improperly constructed regular object. But that issue can be circumvented in C, for foreign inheritance.

PR #1010, opened 2021-05-08
Add speed-trial infrastructure

cxw42:
This PR is the test script I wrote in support of #1009.  I thought I'd submit it in case it's useful to others.
- `util/benchmark.py` tests the _same_ Wren code between _different_ commits.
- With this PR, you can test _different_ Wren code within _one_ commit.

What this PR does:
- util/test.py: Exclude directory test/speedtrials.  This gives us a place to hold speed tests that aren't part of the test or benchmark suites.
- New util/speedtrial.pl to run speed trials

For example:
1. Build Wren
1. Create `test/speedtrials/foo-1.wren`, `test/speedtrials/foo-2.wren`, and `test/speedtrials/foo-3.wren`.
1. Run `./util/speedtrial.pl test/speedtrials/foo`.  It will run `test/speedtrials/foo*.wren` and output a performance comparison.

### DIFF
--- a/util/speedtrial.pl
+++ b/util/speedtrial.pl
@@ -1,0 +1,123 @@
+#!/usr/bin/env perl
+
+=head1 NAME
+
+speedtrial.pl - Compare speed of multiple Wren files
+
+=head1 SYNOPSIS
+
+  speedtrial.pl STEM
+
+Runs all C<< <STEM>*.wren >> tests and reports the results.
+
+=cut
+
+use 5.006;
+use strict;
+use warnings;
+use Benchmark qw(cmpthese timethese);
+use Config;
+use File::Spec;
+use FindBin;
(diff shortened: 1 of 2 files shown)


RobLoach:
Could we stick with Python for this? Adding a Perl script seems out of the norm.

cxw42:
Feel free!

joshgoebel:
>  in case it's useful to others.

@cxw42 Thanks for thinking of everyone.  I feel like we should either have a policy that says we accept random helpful scripts (written in a variety of languages) or perhaps we need another repository or Wiki page to collect these instead (rather than living in PRs that will never be merged).

I think I'd be against such a policy for `utils` but if we had a `contrib` folder that we more open I think that might be fine. But then of course one might worry how you prevent that from becoming a "junk drawer" of things no one uses (other than the original author).

PR #1021, opened 2021-05-18
Reduce the scope of multiple utility classes in core module

jube:
This PR reduces the scope of some classes in the core module, namely:

- `MapSequence`
- `SkipSequence`
- `TakeSequence`
- `WhereSequence`
- `StringByteSequence`
- `StringCodePointSequence`
- `MapKeySequence`
- `MapValueSequence`

These classes do not appear in the documenation, so if code relies on their presence, it's some kind of undefined behavior. As a counter-example, `MapEntry` was not removed as its name explicitely appears in the documentation.

Some pros:

- The names do not appear anymore in the core module and can be used safely in user code
- The number of core symbols is reduced
- The creation of classes is paid only when needed, i.e. when the function is called. As a consequence, it reduces the runtime of core loading and hence the startup time.
- If the classes have to be changed in the future, it will be easier as they are not part of the core interface anymore

Some cons:

- These classes have to be created at each call of the functions. But the functions create a whole sequence so the cost can be balanced with the cost of the loop (that does not change).


joshgoebel:
Very interesting.  Someone is going to ask for performance numbers I'm sure. :-)

> The names do not appear anymore in the core module and can be used safely in user code
> The number of core symbols is reduced

Curious, which of these was the primary motivator for this PR?

>  it reduces the runtime of core loading and hence the startup time.

I'd say for most use cases this feels backwards... I'd rather have the runtime performance and let it start up a few milliseconds slower.  I've been doing a lot of test suite work lately and we're already so much faster than so many things at startup.  

> If the classes have to be changed in the future, it will be easier as they are not part of the core interface anymore

Perhaps they should simply be named `TakeSequence_` as another way of solving this concern.

mhermier:
No need to ask for performance numbers. There is a certain issue in the compiler/vm, that makes it not really suitable to have embed class inside function. Each time you call the class you patch each method, and the performance are far beyond one would expect since we recreate a class on each calls.
In simple terms, this is not the way to go, even if it works.

ruby0x1:
Thanks for the PR!

Every one of these is used extensively in regular wren code, it would have a large impact on general performance if you pay the cost for every single call site. I don't think the trade off it the right one, the start up time is negligible and the runtime issue is far more important. 

I also don't think this is an issue to solve, I know it was discussed in other issues but sometimes these discussions are solutions in search of a problem.

The effort is appreciated though.

joshgoebel:
> and can be used safely in user code

@ruby0x1 Is this perhaps a small valid point?  If we use `symbol_` elsewhere to indicate "private", should these classes be marked the same - or are they intended to be part of the public API even if undocumented?

jube:
@ruby0x1 OK, no problem.

Just to mention a small point though. Three of these classes appear in tests and, IMO, they should not appear as they are not part of the interface.

- in `test/core/string/bytes.wren`, `StringByteSequence` appears but the documentation says that `bytes` should return a `Sequence`
- in `test/language/module/implicitly_imports_core/module.wren`, `MapKeySequence` and `MapValueSequence` appear but again, the documentation does not specify the actual type of `keys` and `values` (just `Sequence`)

Do you want a PR just for this?

joshgoebel:
> they should not appear as they are not part of the interface.

I'm not sure I see a problem with parts of our test suite testing the internals... that sounds like more of a policy decision?  I was more concerned with the naming and whether we want people just making their own new instances of `StringByteSequence` for whatever reason or whether we should flag it as private API.

> StringByteSequence appears but the documentation says that bytes should return a Sequence

I think `StringByteSequence` is just an implementation detail here.  It's a Sequence by all intents and purposes because it inherits from Sequence... I'm guessing the fact that it's a subclass isn't relevant to the API usage.

PR #1022, opened 2021-05-20
String inside a sequence is quoted in toString method.

ThakeeNathees:
### test.wren
```lua
var list = [42, 12, "42, 12", false, "false"]
System.print(list)

var map = {12: "List", "12": List}
System.print(map)
```

### Before
```
[42, 12, 42, 12, false, false]
{12: List, 12: List}
```

### After
```
[42, 12, "42, 12", false, "false"]
{12: "List", "12": List}
```


util/wren_to_c_string.py
```diff
@@ -22,6 +22,7 @@
 def wren_to_c_string(input_path, wren_source_lines, module):
   wren_source = ""
   for line in wren_source_lines:
+    line = line.replace("\\", "\\\\")
```
joshgoebel, line 25: This should likely be it's own PR if Ruby has't fixed it already... I believe she mentioned it in another issue.  I've seen this also.

ThakeeNathees: I've used `\"` in `wren_core.wren` and which makes the  `wren_core.wren.inc` generation faill because `\"` become `\\"` and the c string terminates with the quote.

joshgoebel: _Yes, I'm familiar with the problem._  But this feature may or may not be merged - but that fix should be merged no matter what because it's actually a bug fix.

mhermier: For that line only yes.

mhermier:
For the content of this patch I have some concerns, as it tries to do an equivalent of python `repr`. Implementation wise, I would have preferred a `String.quote`, and make it part of join instead. But even with that In place I'm still wondering.

joshgoebel:
My thumbs up was for the functionality itself.  I think this improves the readability.  The implementation could likely be improved further, agreed.

Issue #988, opened 2021-04-24
string literals dont account for CRLF


Orcolom:
![image](https://user-images.githubusercontent.com/10909816/115959231-7b7dd700-a50b-11eb-962f-809fc8eadf13.png)

Orcolom:
Using this code

```js
System.print("a string") // expect: a string

var long = "
  A
  multi line
  regular string
"
System.print(long) // expect: 
                   // expect:   A
                   // expect:   multi line
                   // expect:   regular string
                   // expect: 

var raw = """
  A if*(<invalid>)*
  multi line /{}()
  raw string [\]/
  "json": "value"
"""
System.print(raw) // expect:   A if*(<invalid>)*
                  // expect:   multi line /{}()
                  // expect:   raw string [\]/
                  // expect:   "json": "value"
```

return results are 
```
"a string"
"\n"

"\r\n  A\r\n  multi line\r\n  regular string\r\n"
"\n"

"\r\n  A if*(<invalid>)*\r\n  multi line /{}()\r\n  raw string [\\]/\r\n  \"json\": \"value\"\r"
"\n"
```

observations:
1. both multi line strings and raw strings have the `\r` show up
2. raw multi line string starts with `\r\n`
3. raw multi line strings has a trailing `\r`


I did a small test where I change every instance of `\r\n` to `\n`.  It gave me some failed tests with trimming, but I haven't had the time to check why

So, at the very least point 2 and 3 should be fixed to adhere to wren standards.

But i do have a question if we should keep or remove `\r\n` to simplify working with strings across platforms

joshgoebel:
What are "wren standards" here again?  If we're using \r\n for newlines (inside raw strings) in general then then example 3 seems broken to me, in that it should end in `\r\n` to be consistent with everything else.

Orcolom:
Standard was wrong word, I mean mainly "what is expected behaviour"

From my understanding both the leading and trailing newline should have been ignored in the raw string

joshgoebel:
OH, are you actually putting literal \r and \n characters there and then expecting them to come back verbatim?  I guess I wasn't thinking of invisibles or special chars... So I guess raw to me was "mostly raw" not "any binary characters at all"... so I think the expected behavior may indeed need to be clarified.

>  This means the opening and closing lines are not counted as part of the string when the triple quotes are separate lines, as long as they only contain whitespace (spaces + tabs).

It does seem to say only spaces and tabs are whitespace, not `\r`.

[...]

PR #1023, opened 2021-05-20
wren_core.wren.inc backslash characters escaped

ThakeeNathees:
Cannot find the actual issue it requires but this fix also needed for #1022


ruby0x1:
Likely referring to this: https://github.com/wren-lang/wren/issues/988#issuecomment-842670624

PR #1032, opened 2021-05-29
Clean up functions in header vs source files

### LINKED ISSUE #1030: Can the amalgamater put the implementing code around a preprocessor macro?
It's somewhat canonical for single-file C libraries to have some macro you must define before including it to have it actually implement the implementing code (where it's included), but still permit you to include it for definitions to use. The popular stb libraries are what popularized this and others have followed since (sokol, raylib, zpl, mbedtls, sqlite3, etc)

Here's an example:
```c
#define STB_IMAGE_IMPLEMENTATION
#include "stb_image.h"
```
Now puts the implementing code in the source file you include it from, but other source files can still do the following for function definitions, macros, enums, etc.
```c
#include "stb_image.h"
````

You can checkout some more information of this technique [here](https://github.com/nothings/stb/blob/master/docs/stb_howto.txt)

I see no way the current amalgamation could work unless one only uses wren symbols from the source included which seems overly restrictive and unusable even.

This was easy enough for me to add to the script in `util` though, patch is below
```diff
diff --git a/util/generate_amalgamation.py b/util/generate_amalgamation.py
index a0d8076f..f6977940 100755
--- a/util/generate_amalgamation.py
+++ b/util/generate_amalgamation.py
@@ -63,6 +63,8 @@ out.write('\n')
 # Source files.
 add_file(join(WREN_DIR, 'src', 'include', 'wren.h'))
 
+out.write('#ifdef WREN_IMPLEMENTATION\n')
+
 # Must be included here because of conditional compilation.
 add_file(join(WREN_DIR, 'src', 'vm', 'wren_debug.h'))
 
@@ -71,3 +73,5 @@ for f in iglob(join(WREN_DIR, 'src', 'vm', '*.c')):
 
 for f in iglob(join(WREN_DIR, 'src', 'optional', '*.c')):
   add_file(f)
+
+out.write('#endif\n')
```
Thoughts?

RobLoach:
This cleans up the static declarations of functions that are used outside of their own source files by having the declaration in the .h file, and the definition in the .c file. In turn, this enables the ability to use `WREN_IMPLEMENTATION` in the amalgamated wren.h file.

Fixes #1030

### DIFF
--- a/src/vm/wren_vm.c
+++ b/src/vm/wren_vm.c
@@ -8,6 +8,7 @@
 #include "wren_debug.h"
 #include "wren_primitive.h"
 #include "wren_vm.h"
+#include "wren_value.h"
 
 #if WREN_OPT_META
   #include "wren_opt_meta.h"
@@ -1610,6 +1611,26 @@ int wrenDefineVariable(WrenVM* vm, ObjModule* module, const char* name,
   return symbol;
 }
 
+void wrenCallFunction(WrenVM* vm, ObjFiber* fiber,
+                      ObjClosure* closure, int numArgs)
+{
+  // Grow the call frame array if needed.
+  if (fiber->numFrames + 1 > fiber->frameCapacity)
+  {
+    int max = fiber->frameCapacity * 2;
+    fiber->frames = (CallFrame*)wrenReallocate(vm, fiber->frames,
+        sizeof(CallFrame) * fiber->frameCapacity, sizeof(CallFrame) * max);
+    fiber->frameCapacity = max;
+  }
+  
+  // Grow the stack if needed.
+  int stackSize = (int)(fiber->stackTop - fiber->stack);
+  int needed = stackSize + closure->fn->maxSlots;
+  wrenEnsureStack(vm, fiber, needed);
+  
+  wrenAppendCallFrame(vm, fiber, closure, fiber->stackTop - numArgs);
+}
+
 // TODO: Inline?
 void wrenPushRoot(WrenVM* vm, Obj* obj)
 {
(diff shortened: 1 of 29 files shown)


src/vm/wren_vm.c
```diff
@@ -1610,6 +1611,26 @@ int wrenDefineVariable(WrenVM* vm, ObjModule* module, const char* name,
   return symbol;
 }
 
+void wrenCallFunction(WrenVM* vm, ObjFiber* fiber,
```
RobLoach, line 1614:
``` suggestion
inline void wrenCallFunction(WrenVM* vm, ObjFiber* fiber,
```

util/generate_amalgamation.py
```diff
@@ -4,7 +4,9 @@
 from os.path import basename, dirname, join, realpath, isfile
 from glob import iglob
 import re
+from typing import TYPE_CHECKING
```
RobLoach, line 7:
``` suggestion
```

src/vm/wren_vm.h
```diff
@@ -206,46 +189,11 @@ void wrenPopRoot(WrenVM* vm);
 // Defined here instead of in wren_value.h because it's critical that this be
```
ruby0x1, line 189: some of this is very intentional, btw. undoing it isn't ideal.

mhermier:
I don't understand why moving all these functions are necessary. Isn't it possible to isolate "wren.h" at the top of the amalgation and isolate all the other files inside a giant `#ifdef` guard around all others files inclusions? There should be no good reasons to allow to access internal, so it should be simpler to do like this.

RobLoach:
Would like to figure out how. The static inline functions were causing undefined references. I'm sure there is a way around it though.

Issue #1052, opened 2021-09-04
Are 0.0 and -0.0 equal or not?


Srinivasa314:
https://wren.io/try/?code=G4QwTgBAHgvADAKFJAnjAtI5EAWMDeUAXAIwC+CAyigM4AuApgLYB0ADmAJYB2dAFLBgoAlFVqNWHHvxwBtFAF1hQA
There seems to be two types of equality (because the bit representation of -0 and 0 differ). I am not sure whether this is documented anywhere.

mhermier:
That issue should be value representation dependant. Union tagged values
should not be affected.

PureFox48:
As far as I'm aware, Wren always treats ```0``` and ```-0``` as being equal in value even though their underlying floating-point bit representations differ.

However, in the docs for the [```Object.same```](https://wren.io/modules/core/object.html) method it says:

> For value types, this returns true if the objects have equivalent state. In other words, numbers, strings, booleans, and ranges compare by value.

In the interests of clarity, I suggest we add the following sentence to that paragraph:

"Note that the numbers ```0``` and ```-0``` are treated as being equal in value even though their bit representations differ."

joshgoebel:
Does this depend at all on the  compile options?

PureFox48:
I don't think so though @mhermier would know better than me.

The position seems to be governed by the following lines 1407-1410 in the implementation of ```Object.same``` in  [wren_core.c](https://github.com/wren-lang/wren/blob/main/src/vm/wren_core.c):

```c
// These are defined just so that 0 and -0 are equal, which is specified by
// IEEE 754 even though they have different bit representations.
PRIMITIVE(vm->numClass, "==(_)", num_eqeq);
PRIMITIVE(vm->numClass, "!=(_)", num_bangeq);
```
So internally they remain distinct which often shows up when you're trying to format numbers for display purposes - negative zero is a nuisance, frankly in that particular scenario.

mhermier:
The problem highly lies in https://github.com/wren-lang/wren/blob/accfa598b3cbc54fdfb33cb568055ed6994a4966/src/vm/wren_value.c#L430 and the notion of equality. One would expect that equality implies same hashing, and this is usually the case, but it is not an absolute truth in particular with ieee754 double. Some values are considered equal, but in their representation they can be fundamentally different, and some numbers are not equals despite being the same exact bits.

[...]

PR #1054, opened 2021-09-10
Clarify treatment of negative zero

PureFox48:
Suggested wording to be added to docs for Object class following discussion in #1052.

### DIFF
--- a/doc/site/modules/core/object.markdown
+++ b/doc/site/modules/core/object.markdown
@@ -29,6 +29,8 @@ Compares two objects using built-in equality. This compares [value
 types](../values.html) by value, and all other objects are compared by
 identity&mdash;two objects are equal only if they are the exact same object.
 
+Note however that these operators are overridden in the `Num` class to ensure `0` and `-0` are treated as being equal in value even though they have different bit representations. Object.same(0, -0) still returns `false` because they are different objects.
+
 ### **is**(class) operator
 
 Returns `true` if this object's class or one of its superclasses is `class`.


doc/site/modules/core/object.markdown
```diff
@@ -29,6 +29,8 @@ Compares two objects using built-in equality. This compares [value
 types](../values.html) by value, and all other objects are compared by
 identity&mdash;two objects are equal only if they are the exact same object.
 
+Note however that these operators are overridden in the `Num` class to ensure `0` and `-0` are treated as being equal in value even though they have different bit representations. Object.same(0, -0) still returns `false` because they are different objects.
```
ruby0x1, line 32: I would say "different values" (last two words). I wouldn't want to imply numbers are objects, since they're not (they're a primitive value).

ruby0x1: although Object.same on two distinct values is a weird thing... would we even expect it to return true when you're comparing two distinct _values_?

PureFox48: The reason I used the word `object` is because we're in the Object class docs and, technically, everything in Wren is an object even numeric literals. On the other hand Num is a value type and so they're values as well. If you're more comfortable with `value` we can change it.

You'd expect Object.same to return false and so the last sentence is really superfluous. The only reason I put it in was to make the distinction between that method and the equals operator crystal clear but we can remove it if you like. 

Incidentally, when I was playing around with this the other day, I noticed another weird thing about negative zero. Although (-0).toString returns `-0` as you'd expect, (-0).sign returns `0` rather than `-1`. I'm not suggesting that this is wrong and we should change it but do you think we should mention it in the docs for Num.sign?

ruby0x1: from the `same` docs right above, yea:

> For value types, this returns true if the objects have equivalent state. In other words, numbers, strings, booleans, and ranges compare by value.

I don't think using 'object' to refer to values is clearer, it adds questions. I don't think same needs a mention.

I also feel a bit weird about Num details being in the `==` operator documentation, but regardless I'd simplify this either way. It could be relevant to mention it in the Num docs, where it documents `==` specifically to call this out. That's where I'd look to answer questions about Num.

> Note: `Num` overrides these operators so that `0 == -0` returns `true`.

PR #1068, opened 2022-01-06
Fix unused variable warnings

trufae:


ruby0x1:
hey @trufae thanks for the PR, if you check the build failures on the CI above, you'll notice there's an issue in your proposed change when both OPT and META are enabled. You could to add `(void)nameString;`  after the initial declaration, or to add `{ }` blocks around each of opt and meta to localize that variable further.

PR #1075, opened 2022-01-19
fix: stack smaching with long names

### LINKED ISSUE #1033: stack smashing leads to segfault for malformed class definitions involving unterminated strings and attributes
Reduced this down to the most basic example I can think of, the commented lines exist purely to pad out the space used on the stack which leads to the stack overflow.

```
#!
class " A { }
////////////////////////////////////////
////////////////////////////////////////
////////////////////////////////////////
////////////////////%%%%%%%%%%%%
```

This will lead to a stack overflow, as seen here with `./wren_test`
```
[graphite@graphite bin]# ./wren_test a.wren 
< several lines of errors snipped >
[./a line 7] Error at end of file: Class " A { }
////////////////////////////////////////
////////////////////////////////////////
////////////////////////////////////////
////////////////////%% already defines a method '%(_)'.
*** stack smashing detected ***: terminated
Aborted (core dumped)
```

Adding one more comment to pad it out past the stack check leads to a segmentation fault
```
#!
class " A { }
////////////////////////////////////////
////////////////////////////////////////
////////////////////////////////////////
//////////////////////%%%%%%%%%%%%
```

```
[graphite@graphite bin]# ./wren_test a.wren 
< several lines of errors snipped >
[./a line 7] Error at end of file: Expect ')' after parameter name.
Segmentation fault (core dumped)
```

The code input is pathological but Wren should not crash on pathological inputs either.

### LINKED ISSUE #1036: Long duplicate method names (even under 64 characters) leads to stack overflow in parser
This is similar to issue #1034 except the method names are not actually longer than 64 characters, they're at 58 characters in length. The body of the method is empty too, and there is no invalid tokens necessary.

It looks like length management is messed up under duplicate names and that alone leads to stack overflow.

```
class Foo {
  iiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiii {}
  iiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiii {}
}
```

Like the other issue, adding more `i`s if it does not crash will make it crash too if it isn't already. This one I think is more interesting from a user perspective because the identifier is less than 64 characters, adding more to exceed it is not surprising either, though it shouldn't crash in this fashion. Without stack protector this leads to a segfault in my case.

Output
```
[graphite@graphite bin]# ./wren_test a.wren 
[./a line 3] Error at 'iiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiii': Class Foo already defines a method 'iiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiii'.
*** stack smashing detected ***: terminated
Aborted (core dumped)
```

joshgoebel:
Resolves #1033.
Resolves #1036.

Better handling of very long error messages.

### DIFF
--- a/src/vm/wren_compiler.c
+++ b/src/vm/wren_compiler.c
@@ -48,7 +48,7 @@
 // is kind of hairy, but fortunately we can control what the longest possible
 // message is and handle that. Ideally, we'd use `snprintf()`, but that's not
 // available in standard C++98.
-#define ERROR_MESSAGE_SIZE (80 + MAX_VARIABLE_NAME + 15)
+#define ERROR_MESSAGE_SIZE (80 + MAX_VARIABLE_NAME + MAX_VARIABLE_NAME + 15)
 
 typedef enum
 {
@@ -3331,12 +3331,18 @@ static int declareMethod(Compiler* compiler, Signature* signature,
     if (methods->data[i] == symbol)
     {
       const char* staticPrefix = classInfo->inStatic ? "static " : "";
-      error(compiler, "Class %s already defines a %smethod '%s'.",
+      if (compiler->enclosingClass->name->length > MAX_VARIABLE_NAME) {
+        error(compiler, "Class %.*s... already defines a %smethod '%s'.",
+          MAX_VARIABLE_NAME,
+          &compiler->enclosingClass->name->value, staticPrefix, name);
+      } else {
+        error(compiler, "Class %s already defines a %smethod '%s'.",
             &compiler->enclosingClass->name->value, staticPrefix, name);
+      }
+
       break;
     }
   }
-  
   wrenIntBufferWrite(compiler->parser->vm, methods, symbol);
   return symbol;
 }


mhermier:
It might work as a temporary solution, to mitigate the error.
For a long term solution, a better solution would be to use `wrenStringFormat`, but that would require some effort to support more types.

joshgoebel:
Yeah I was trying to follow the existing established pattern for such things.  A better solution might be what you suggest or just replacing the ASSERT in `printError` with a runtime error instead... truncate the entire string instead of trying to truncate each individual %s.

joshgoebel:
> Ideally, we'd use `snprintf()`, but that's not available in standard C++98.

Ah...

joshgoebel:
Ping.  Can I do anything further to expedite this?  As far as I'm aware this is a valid fix that uses the same approach already put forward in the codebase...

ruby0x1:
I'll look at it closer soon (we moved house during Feb so been mostly afk). I'm curious, since you say expedite, is it an issue you're running into very frequently somehow? or blocking somehow? It looks fine at a glance but I want to check the branch closer is all. 

One thing is it doesn't mention why the length was doubled in the comments, so anyone stumbling on it will have to wonder.

joshgoebel:
> One thing is it doesn't mention why the length was doubled in the comments

I added a mention of why.

> is it an issue you're running into very frequently somehow? 

No, I just love Wren and was trying to find some non-controversial low-hanging-fruit to contribute to core.  Since there were 3 separate issues that could lead to this and the *immediate* short-term fix seems clear (since there is existing precedent - we solve this already, just not completely) I thought this was an easy one to start with.

PR #1078, opened 2022-01-31
Fix formatting error

tavi-vi:
Fixed error that caused `<stdio.h>` to disappear from C `#include` directive


ruby0x1:
Thanks for the fix! (It might not go live to the docs immediately atm).

iacore:
The site is still broken

PR #1083, opened 2022-02-19
api: add wrenCopySlot

erincandescent:
Simple utility function which copies a value from one slot to another, as originally proposed in #836

### DIFF
--- a/doc/site/embedding/slots-and-handles.markdown
+++ b/doc/site/embedding/slots-and-handles.markdown
@@ -97,6 +97,11 @@ length of the string using `strlen()`.
 
 [string]: ../values.html#strings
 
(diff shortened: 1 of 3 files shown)


mhermier:
Instead add `wrenCopySlots` and make `wrenCopySlot` a special case/macro...

erincandescent:
I thought about that, but it opens up memcpy esque cans of worms (what do we do if they're overlapping). 

Nothing insurmountable but I've never actually needed to move multiple

mhermier:
Well this can be as this for now. The main usage for copying in block, is
when one needs to realign the stack for a reentrant call, or when
performing tail call optimisation (which are not supported yet).. That
said, implementation should always be safest, so it should follow `memmove`
semantic.

PR #1088, opened 2022-04-06
check `MAX_UPVALUEES` before adding an upvalue

ThakeeNathees:
Not sure about returning `-1`, but it'll be seen as the upvalue wasn't found by the `findUpValue` to the caller.


mhermier:
Nice catch, mandatory typo in the commit message. Do not know if it easy to
create a test for it.

Le mer. 6 avr. 2022 à 20:59, Thakee Nathees ***@***.***> a
écrit :

PR #1090, opened 2022-05-28
Implement fmin/fmax ourselves when compiling for djgpp

phcoder:


src/vm/wren_core.c
```diff
@@ -13,6 +13,26 @@
 
 #include "wren_core.wren.inc"
 
+#ifdef __DJGPP__
+double fmax(double x, double y)
+{
+  if (isnan(x))
```
joshgoebel, line 19: - Why would double not be a number?
- What is y isn't a number either?
- If one or both aren't "numbers" what would just falling thru to the comparison do?

minirop: 1/ not all 2^64 values possibles for a double are numbers. Like `sqrt(-1)`.
2/ if both are NaN, you can only return NaN.
3/ all binary operations (except `==`) returns `false` if any (or both) is NaN, therefore `x` would be returned. But then, the second `if` isn't strictly necessary.

ruby0x1: You can check musl for a standard lib implementation https://git.musl-libc.org/cgit/musl/tree/src/math/fmax.c

phcoder: > * Why would double not be a number?
e.g. 0.0/0.0, infinity * 0.0, and so on
> * What if y isn't a number either?
By the doc you need to return NaN
> * If one or both aren't "numbers" what would just falling thru to the comparison do?
It's possible to optimize a bit by removing second check but code becomes way more unintuitive and error-prone. Second one is strictly necessary to be correct in case of x=nan, y=1 which needs to return 1.

According to the fmin/fmax doc:
RETURN VALUE
       These functions return the minimum [maximum] of x and y.

       If one argument is a NaN, the other argument is returned.

       If both arguments are NaN, a NaN is returned.

phcoder: > You can check musl for a standard lib implementation https://git.musl-libc.org/cgit/musl/tree/src/math/fmax.c

Thanks. I'll update based on it

phcoder: Updated

phcoder:
Ping?

mhermier:
I would say to put that functionality in wren_math.h as a macro that define
an inline function, guarded as you did.  See
https://ffmpeg.org/doxygen/trunk/math_8c_source.html for inspiration.

PR #1095, opened 2022-07-01
Makes the validateApiSlot function a macro

ConfusedSky:
This PR changes VALIDATE_API_SLOT to be a macro instead of a function to make the error messages that you get from the assertion much more useful for anyone using it in an embedding project.

I ran all the tests to be sure (on both debug and release) I don't think this changes functionality too much, but this does improve the visibility significantly in dev builds


ConfusedSky:
That being said, it's not as necessary now that I've got LLDB working properly with the call stack

mhermier:
Your change as the nice property to be able to report the error in the
console without debugger.

PR #1097, opened 2022-07-17
Added an example of 'else if' to the docs

stelleron:
Currently there isn't any example of an else if statement currently in docs, or an explicit mention of else if statements existing at all, I wanted to add an example of else if statement so new users aren't confused as to whether they exist in Wren.


RobLoach (approved):

ruby0x1:
thanks!

Issue #1098, opened 2022-07-20
`Random.new(seed)` where `seed` : `Num` relies on platform's cstdlib `srand()`/`rand()`


nitz:
Making use of passing a number to `Random.new(seed)` eventually calls [`randomSeed1(WrenVM* vm)`](https://github.com/wren-lang/wren/blob/main/src/optional/wren_opt_random.c#L57-L66).

`randomSeed1` uses that number cast as a `uint32_t` to call cstdlib's `srand()`, and then fills out the initial Well512 state with 16 subsequent calls to `rand()`.

The result of this is that on a given platform, seeding random with a single number will produce an expected, repeatable output. Using the same seed on a different platform however, _may_ not produce the same output as the original.

Example:

```dart
import "random" for Random
var letters = "ABCDEFGHIJKLMNOPQRSTUVWXYZ"
var rng = Random.new(1234)
System.print(rng.sample(letters, 10))
```

On my windows machine outputs: `[P, Q, D, T, G, V, W, C, J, F]`
However, the [try wren](https://wren.io/try/?code=JYWwDg9gTgLgBAIigQwHYBMIgXAZtOAJTUxACgA3ZKOAGwFMYZ6oBnOAXkQEEAhAYQAiAUQBiAcQASASQBSAaQAyAWQByAeQAKARUIBlACoBVAGoB1ABoBNAFoJK1OFFQBzTkRJYAdKnoB3AAoARgAmAGYAFgBKMj0AT1ZmEC8wKGBUGADnFy9WZHAGAIYmFlYAGjgggAYoqKA) page prints: `[G, O, H, E, U, A, W, J, M, Z]`

Forcing `Random.new()` to seed the initial state with non-`rand()` values by passing a `Sequence` instead causes the behavior to be properly deterministic:

```dart
    import "random" for Random
    var letters = "ABCDEFGHIJKLMNOPQRSTUVWXYZ"
    var rng = Random.new([1234]) // note [1234] rather than 1234
    System.print(rng.sample(letters, 10))
```

windows: `[Q, B, S, T, U, C, I, F, W, P]`
[...]

PR #1099, opened 2022-07-20
Remove `randomSeed1()`'s reliance on cstdlib's rand implementation.

nitz:
This adds a simple TLCG to initialize the random state when only a number is supplied as seed.

This has the result of removing the reliance on cstdlib's `srand()`/`rand()` implementations for setting the initial state, meaning that the same seed value will have the same effect regardless of platform.

Proposed as a solution for #1098


ConfusedSky:
For consistencies sake wouldn't it make sense to do the same thing for `randomSeed0()`? Using time as the seed value?

PR #1100, opened 2022-07-20
Update Premake to 5.0.0-beta1; Add VS2022 Solution

nitz:
Just bumped the premake binary to the latest release and generated projects/solution for Visual Studio 2022.


RobLoach (changes requested):
You may want to update `premake5.lua` as well.

nitz:
I didn't need to make any changes to `premake5.lua` to generate the 2022 solution, and the others seemed to generate as expected. Is there something else `premake5.lua` should be doing that it isn't?

mwasplund:
I would like to build using vs2022, can this pr get checked in?

PR #1102, opened 2022-07-31
Specialize `Range.contains(_)`

ConfusedSky:
Creates a specialization for `Range.contains(_)` that is a bound check rather than using `Sequence`'s implementation which uses iteration instead.

This came up when I was implementing some Project Euler stuff when and I noticed the pathological behavior in some of those problems as well. Particularly problem 17 was about 1000x slower using ranges (I had like 6 different ranges).

I didn't add any new tests since this method is pretty extensively covered. I did create a pathological benchmark for it though. 

Benchmark included below uses `./bin/wren_test`. I wasn't sure if I should include it in the benchmarks folder since it looks like that is for comparing against different languages.

Looks like it's roughly 300x faster in this pathological case, but I reran it for `(0...10)` and it still is about 5x faster. 

Anyway, let me know if I should change anything here! This isn't too big of a change and I think it can help all users of the language. 

```wren
var start = System.clock
var total = 0
for (i in 0..1e5) {
    if ((0...1e3).contains(i)) {
        total = total + 1
    }
}
System.print(total)
System.print("elapsed: %(System.clock - start)")

// Before
// 1000
// elapsed: 2.968864

// After
// 1000
// elapsed: 0.010275
```

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -29,4 +29,4 @@ ruby0x1 <ruby0x1@pm.me>
 Kolja Kube <code@koljaku.be>
 Alexander Klingenbeck <alexander.klingenbeck@gmx.de>
 Aviv Beeri <avbeeri@gmail.com>
-
+Masa Maeda <masamaedae@gmail.com>
--- a/src/vm/wren_core.c
+++ b/src/vm/wren_core.c
@@ -919,6 +919,24 @@ DEF_PRIMITIVE(range_max)
   RETURN_NUM(fmax(range->from, range->to));
 }
 
+DEF_PRIMITIVE(range_contains)
+{
+  ObjRange* range = AS_RANGE(args[0]);
+
+  if (!validateNum(vm, args[1], "Element")) return false;
+  double num = AS_NUM(args[1]);
+
+  // If it's an empty range it can't contain any value
+  if (range->from == range->to && !range->isInclusive) RETURN_FALSE;
+  // Handle the inclusive case
+  if (!range->isInclusive && num == range->to) RETURN_TRUE;
+
+  double min = fmin(range->from, range->to);
+  double max = fmax(range->from, range->to);
+
+  RETURN_BOOL(num > min && num < max);
+}
+
 DEF_PRIMITIVE(range_isInclusive)
 {
   RETURN_BOOL(AS_RANGE(args[0])->isInclusive);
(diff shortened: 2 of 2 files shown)


RobLoach:
Looks like some great performance improvement here! Thanks a lot.

ConfusedSky:
@mhermier What do you mean? It's passing all the tests. So as far as the tests are concerned the behavior is the same.

mhermier:
@ConfusedSky I mean that the tests are not sufficient. `Range` is an iterable, not the continuous mathematical object that represents all the numbers within 2 numbers that are the limits.
As a general rule, always consider what you are writing and testing against. If a test does pass or not always look at the tests you are referring to. They are human products, and can be broken/insufficient/whatever...

ConfusedSky:
@mhermier I guess I hadn't considered that, however in that case really the only thing that would need to be checked is if the number was an integer. 

Although I do think continuous range checking is valuable. Potentially this could be split into two methods, where something like `isInRange` (probably want a better name where) could be as written, and `contains` would require the number to be an integer. But I agree now, the behavior is inconsistent.

mhermier:
@ConfusedSky well situation is more complicated than that. Since there is no requirement for `Range` bounds to be integers, so one has to iterate the range, in case the accumulation brings accumulation errors...

I also guess there is a potential need for `isInRange` functionality, but it can be implemented externally. So unless there is a feasibility issue I didn't see, or a use case where it is proven to be a real performance issue, there is no need for inclusion in core.

ConfusedSky:
@mhermier Ah yea I see. But it does appear to always increment by one. I'll revisit this.
image

mhermier:
See/use `calculateRange` it handles all the corner cases, and in case we add customized step value support, it would be the place to handle the situation.

ConfusedSky:
It appears that the implementation in main doesn't work either
image
Floating point precision error?

That being said also, the specialization for contains for string, isn't consistent with the one for sequence.

mhermier:
Where do you think there is an error? I don't see it...

mhermier:
For fun this statement can be verified in c++ using:
```c++
static_assert((2.2-1)==1.2);
```
which produce in compiler explorer for example:
```
<source>:1:22: error: static assertion failed
    1 | static_assert((2.2-1)==1.2);
      |               ~~~~~~~^~~~~
<source>:1:22: note: the comparison reduces to '(1.2000000000000002e+0 == 1.2e+0)'
```

ConfusedSky:
I tried node and python as well, same deal. Since presumably, it has more to do with IEE-754 than the behavior of wren. 

As written, `contains` doesn't seem to be very useful for ranges that have `from` values that aren't integers.

mhermier:
About `String`, it is consistent, as it produce the expected behavior with parameters of a specific shape. So if you check that the searched string is a one character `String` (even UTF-8 ones), it produce the normal behavior of `contains()`. The fact that it also search do substring is an interesting side effect that augment the behavior. But more importantly the user can act on the parameter to ban the unwanted effect.

On the other hand with the original patch, one has to act on the `Range` to see if the behavior was correct.

The nuance is subtle, I agree. It is a question of "does it work as expected with any valid argument" (as one could trivially guess from the API naming convention), and the original `Range` patch was not.

That said, the lack of a proper `Char` class make `String` a little bit to much complex to apprehend, since it is a ball of mud mixing string, character and byte array data. But that another story, and huge can of worms...

PR #1110, opened 2022-10-07
Add `Object.hash`

mhermier:
The basic rationnale is to offer a fast access to the hash of the interal objects, so users can implement their containers more easily and efficiently.

In addition, it also allow regular objects to be hashed by their reference. It allows to have access to a fast hash for objects meant to be unique instances (where `==` does not need to be overloaded).

`Map` behaviour is preserved by `wrenMapIsValidKey` checking we don't try to pass invalid key values. (Not preserving this behaviour would be to complex, and would require reentrancy...)

### DIFF
--- a/doc/site/modules/core/object.markdown
+++ b/doc/site/modules/core/object.markdown
@@ -29,6 +29,10 @@ Compares two objects using built-in equality. This compares [value
 types](../../values.html) by value, and all other objects are compared by
 identity&mdash;two objects are equal only if they are the exact same object.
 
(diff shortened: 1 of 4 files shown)


PureFox48:
Presumably, exposing the internal hash (a Num) would enable us to use mutable reference types (such as List) as Map keys which is something I've often wished we could do as the alternative of stringifying them is neither nice nor efficient.

If I'm right about that, then I'd be in favor of the change.

PR #1117, opened 2022-10-24
Document metaclasses and static methods

PureFox48:
Thought I'd have a go at this which is currently marked as a TODO.

### DIFF
--- a/doc/site/classes.markdown
+++ b/doc/site/classes.markdown
@@ -469,7 +469,42 @@ to or even should define getters or setters for most of your object's fields.
 
 ## Metaclasses and static members
 
-**TODO**
+### Metaclasses
+
+A class is itself an object and is the only instance of its corresponding *metaclass*. Metaclasses in turn are instances of a class called "Class" which inherits directly from Object.
+
+<pre class="snippet">
+class Foo {}
+System.print(Foo.type)        //> Foo metaclass
+System.print(Foo.type.type)   //> Class
+System.print(Class.supertype) //> Object
+System.print(Foo is Class)    //> true (because Foo metaclass inherits from Class) 
+System.print(Class.type)      //> Class (not Class metaclass)
+System.print(Class is Class)  //> strange but true!
+</pre>
+
+"Class" is unusual in that it has no metaclass of its own and is considered to be an instance of itself.
+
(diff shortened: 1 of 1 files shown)


mhermier:
The wording about the fact that "class is strange/unusual" is bad. Instead the documentation should extract: https://github.com/wren-lang/wren/blob/c2a75f1eaf9b1ba1245d7533a723360863fb012d/src/vm/wren_core.c#L1276-L1296
in some form, and this should be referenced instead, to justify the behavior.

PureFox48:
Well, 'Class' is an unusual class in that it has no metaclass (the only one that doesn't) and  I think people will regard it as  strange that it's an instance of itself - a bit reminiscent of the Russell paradox.

This isn't intended as a criticism as I can understand why it's been done this way and object hierarchies of any programming language tend to have a certain amount of strangeness.

Personally, I'm not in favor of including the object hierarchy diagram in what is intended to be simple, readable documentation and I don't think it highlights the strangeness of the 'Class' class very well in any case. The examples I've chosen hopefully do that.

mhermier:
Well I think, the diagram should belong to the documentation, for advanced explanations of how the object model is implemented. Referring it from that change would seems logical, and would would allow some "not-personal" comments like:
```
System.print(Class is Class) // true, look here for details
```

PureFox48:
Well, if we were to include a link to the diagram in the Class docs, then the best place to do it would be in the 'Inheritance' section which needs looking at anyway as it doesn't integrate too well with the 'Super' section at present.

I'm still not keen on referencing it in the 'Metaclasses' section (I can't include a clickable link to it in a code comment anyway) as I don't think the last two examples are easy to understand without a written explanation.

PureFox48:
OK, I've changed the 'Metaclasses' section so that it no longer refers to 'Class' as being unusual or strange.

PureFox48:
I've added some wording to the 'Static methods' sub-section to make it clear that static getters, setters and operators are supported and that, whilst `this` can be used in a static method,  it refers to the class object itself rather than an instance of it.

PureFox48:
Reading through the whole lot again, I noticed that I'd missed out a word in the 'static methods' section. Fixed now.

PR #1118, opened 2022-11-01
Fix for OrangeC

ruanjiashing:
OrangeC doesn't have computed gotos.


ruby0x1:
fwiw if you `#define WREN_COMPUTED_GOTO 0` before the header is included, it won't use computed gotos, e.g in your build system.

mhermier:
Maybe, the platform check should be removed completely and compiled goto should be opt-in instead. This would make the project less efficient by default, but more C99 compliant. And add a note in the build instructions about the variable definition.

ruby0x1:
I think it's fine as is, it's already configurable and has never been brought up as a problem.

PR #1124, opened 2022-11-29
Populate and use `ObjFn->arity` correctly (#731)

mhermier:
`ObjFn->arity` was only populated for functions. Fix that with wren call handles, and regular methods, so the stack can be properly adjusted when invoking `wrenCallFunction`.

### DIFF
--- a/src/vm/wren_vm.c
+++ b/src/vm/wren_vm.c
@@ -1422,6 +1422,7 @@ WrenHandle* wrenMakeCallHandle(WrenVM* vm, const char* signature)
   // Create a little stub function that assumes the arguments are on the stack
   // and calls the method.
   ObjFn* fn = wrenNewFunction(vm, NULL, numParams + 1);
+  fn->arity = numParams;
   
   // Wrap the function in a closure and then in a handle. Do this here so it
   // doesn't get collected as we fill it in.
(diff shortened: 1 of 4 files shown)


src/vm/wren_vm.c
```diff
@@ -1460,9 +1461,10 @@ WrenInterpretResult wrenCall(WrenVM* vm, WrenHandle* method)
 
   // Discard any extra temporary slots. We take for granted that the stub
   // function has exactly one slot for each argument.
-  vm->fiber->stackTop = &vm->fiber->stack[closure->fn->maxSlots];
+  const int closureNumArgs = closure->fn->arity + 1;
+  vm->fiber->stackTop = &vm->fiber->stack[closureNumArgs];
   
-  wrenCallFunction(vm, vm->fiber, closure, 0);
+  wrenCallFunction(vm, vm->fiber, closure, closureNumArgs);
```
ruby0x1, line 1467: this code might not behave as expected? The closure being called here is the stub function created by the call handle, which only has bytecode to forward the call to the receiver. The closure itself has 0 args marked here specifically to avoid interacting with the stack afaict

mhermier: The old behavior was ONLY correct when passing the correct number of arguments using broken logic. There was a bogus relation between what the last `wrenCallFunction` argument was intended to be and what it was really used for. With that change, the argument is now the number of arguments that are on the stack to perform the execution call. Because of that it is not 0 anymore, since obviously a closure can always have any number arguments.
Because a closure has to accept its arity or more arguments, as per `Fn.call`, there MUST be a stack interaction if there is more arguments than required.

PR #1125, opened 2022-11-30
Add `List` `capacity` and `reserve(_)` methods

mhermier:
Exposing these `List` methods allow to grow the underlying container to a size reducing unnecessary allocations and copying, and can bring performance when the final `size`/required `capacity` of the list is known in advance.

### DIFF
--- a/doc/site/modules/core/list.markdown
+++ b/doc/site/modules/core/list.markdown
@@ -36,6 +36,10 @@ System.print(list) //> [0, 1, 2, 3, 4, 5, 6]
 
 Returns the added items.
 
+### **capacity**
+
+The number of elements that can fit in the underlying container before it needs to be resized.
+
 ### **clear**()
 
 Removes all elements from the list.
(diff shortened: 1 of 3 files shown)


ruby0x1:
`List.filled` is related fwiw

PureFox48:
A problem with ```List.filled``` is that sometimes you don't know how much capacity you'll need when creating the list.

This PR would enable  us to increase the capacity subsequently  in 'one hit'  and, as it doesn't add much extra code,  I think it would be worth doing.

mhermier:
`List.filled(_,_)` as 2 side effect that this patch does not have. It does provides an internal buffer and `List` of the exact size. Even adding a `removeAll()` would be me costly to implement in user than what is provided here.

`List.filled(_,_)` could be extended to have a `capacity` argument, but I'm not sure if it is worth it.

ruby0x1:
(related doesn't mean instead of, I'm aware of the value of the patch!)

Issue #1141, opened 2023-02-25
List.sort() very slow if list is already sorted or nearly so.


PureFox48:
I noticed recently that ```List.sort()```, which is based on the Quicksort algorithm, is very slow indeed when applied to lists which are already sorted or nearly so.

For example, this code takes around 5.7 seconds to run on my Core i7 machine:

```javascript
var list = (1..10000).toList
list.sort()
```

I tracked down the problem to the choice of pivot in the private ```partition_``` method in wren_core.wren:

```javascript
partition_(low, high, comparer) {
    var p = this[high]
    // blah
}
```

If instead of always using the high value we use the mid value:

```javascript
partition_(low, high, comparer) {
    var mid = ((low + high)/2).floor
    a.swap(mid, high)
    var p = this[high]
    // blah
}
```

then the code now runs in around 0.014 seconds. Moreover, the effect of adding these two lines has negligible effect on the time taken to sort unsorted lists.

The implementation also has a problem when the list contains repeated elements. For example this code takes around 2.8 seconds to run whatever the pivot choice:

```javascript
var list = [1] * 10000
list.sort()
```

However, I can't think of a _simple_ way to solve this particular problem which probably occurs much less often than the 'already sorted' one in practice. If anyone has an idea on this, please post it.

These, of course, are known issues with Quicksort generally (see which is also very slow when sorting small lists. The latter could be solved by using an insertion sort for lists under 10 elements say but, frankly, I doubt whether it's worth the effort in a simple scripting language such as Wren.

ruby0x1:
For the last snippet I was wondering. I'm assuming so but did you measure only the sort or both lines? If not what does this version measure at?
```js
var list = List.filled(10000, 1)
list.sort()
```

ruby0x1:
Thanks for digging btw, we can probably add some tests around this kinda thing to give us a baseline to compare with.

PureFox48:
I just measured the sort.

The time to create the list is virtually the same at 0.0183 seconds whether one uses: ```List.filled``` or the ```List.*``` operator so possibly they're using the same code under the hood.

Incidentally, the [Wikipedia article on Quicksort](https://en.wikipedia.org/wiki/Quicksort) suggests that using a 'median of three' may be better than a simple midpoint. It was  actually a shade slower in my own tests possibly because the cost of interpreting the extra code outweighed any efficiency gained.

PureFox48:
Oops, sorry, I messed up the timings for the list creation.

[...]

PR #1146, opened 2023-03-03
Change list.sort to use Hoare partitioning

PureFox48:
#1141 identified two problems with the existing implementation, namely that is was very slow when presented with lists which were already sorted or all the same.

The purpose of this PR is to fix these problems by switching from Lomuto to Hoare partitioning. As far as 'random' lists are concerned, the figures in the above issue indicate that there will be no noticeable change in performance for small lists and possibly a small improvement for large lists.

This change will be invisible to users as only two private methods are affected. No change is required to either the documentation or tests.

### DIFF
--- a/src/vm/wren_core.wren
+++ b/src/vm/wren_core.wren
@@ -334,28 +334,30 @@ class List is Sequence {
   }
 
   quicksort_(low, high, comparer) {
-    if (low < high) {
+    if (low >= 0 && high >= 0 && low < high) {
       var p = partition_(low, high, comparer)
-      quicksort_(low, p - 1, comparer)
-      quicksort_(p + 1, high, comparer)
+      quicksort_(low, p, comparer)
+      quicksort_(p+1, high, comparer)
     }
   }
 
   partition_(low, high, comparer) {
-    var p = this[high]
+    var mid = ((low + high)/2).floor
+    var p = this[mid]
     var i = low - 1
-    for (j in low..(high-1)) {
-      if (comparer.call(this[j], p)) {  
+    var j = high + 1
+    while (true) {
+      while (true) {
         i = i + 1
-        var t = this[i]
-        this[i] = this[j]
-        this[j] = t
+        if (!comparer.call(this[i], p)) break
       }
+      while (true) {
+        j = j - 1
+        if (!comparer.call(p, this[j])) break
+      }
+      if (i >= j) return j
+      swap(i, j)
     }
-    var t = this[i+1]
-    this[i+1] = this[high]
-    this[high] = t
-    return i+1
   }
 
   toString { "[%(join(", "))]" }


PureFox48:
How do I change wren_core.wren.inc to match these changes? It says not to edit it as it will be generated automatically but that doesn't seem to have happened here.

mhermier:
```sh
util/wren_to_c_string.py src/vm/wren_core.wren.inc src/vm/wren_core.wren
```

But wait I have 2 minor style code to submit, before you rebase your change.

PureFox48:
Ah, so there's a Python utility to do this which we need to run manually. OK, thanks I'll do that later after we've agreed any changes.

I've based the implementation here on the Wikipedia pseudo-code which is said to be taken from the  'Introduction to Algorithms' book by T.H.Cormen and others.

As it uses ```do/while```, which we don't have in Wren, to ensure certains lines are run at least once, I've used an infinite loop with a reversed logic condition at the end to simulate that. 

Also, don't forget that we can't change the comparator function here from a simple true/false to three-state as it will break existing code which uses a custom comparator.

It's not necessary to do so in any case as quicksort is an unstable algorithm and so the order of elements comparing equal is not guaranteed to be maintained in any case.

PureFox48:
@mhermier 

Any suggested changes here?

If not, I'll regenerate `wren_core.wren.inc` and add it to the PR.

mhermier:
Well there the minor code changes, in the reviews.

Else there is the proposition in the reviews, that could maybe simplify code a little. Thought I didn't checked if it was safe.

PureFox48:
What reviews are those?

The only comment you made about the code in https://github.com/wren-lang/wren/issues/1141 was:

>  It should optimize a little bit more, because functional is a little bit sub-optimal in wren).

I couldn't see how it could be optimized myself because we have no choice but to use a comparator function.

mhermier:
Look at the pull request log. There are 3 reviews opened. 2 about minor
code style, with missing space around binary operators. And 1 about a
possible improvement, where I wonder if adding an early check, could avoid
to have 3 condition, instead of 1. Last one is an open question.

PureFox48:
Well there must be something wrong somewhere as github is telling me there are `No reviews` for this PR.

Are you sure you've posted them and they're not just in draft?

mhermier:
Seems good enough.
If it was an external library, I would have changed the algorithm to some hybrid, mode efficient with smaller lists, and conserve quick-sort for the large lists .

PureFox48:
Well, back in the day, sorting implementations would typically use insertion sort for small lists (less than 10 elements say) and quicksort for larger arrays. We could still do that here (inserton sort is only about 12 lines) but with modern hardware you'd only notice the difference if, say, you were sorting a large number of small lists in a loop. Frankly, I doubt whether it's worth it for core.

funny-falcon:
There is simpler change possible without change of partitioning: pick the middle element and swap it with last one. And no other changes are needed.

But to be reliable for large arrays, there should be “middle-of-middles” implemented. And fallback to heap sort if quadratic behavior is detected.

(Actually, I like fallback to Shell Sort. With carefully chosen gap scheme, it could be quite efficient. Although, algorithmically it is still worse than Heap Sort.)

Issue #1142, opened 2023-02-26
[RFC] Add comparison operators to the String class


PureFox48:
Currently, if you need to compare two strings or sort a list of strings lexicographically by codepoint, then you need to write your own comparison routine. This is an unsatisfactory state of affairs since you can't add your own comparison operators (or anything else) to the built-in ```String``` class or inherit from it.

I therefore propose that we add a ```compareTo``` method and comparison operators to the ```String``` class so that it will be as easy to compare two strings or sort a list of strings as it is to perform the same operations on numbers.

As we need to work at the codepoint rather than the byte level, it will probably be quick enough to do this from the Wren side and here's my attempt at some code which is essentially what I've been using myself for the last 3 years or so, albeit in the guise of static methods of a proxy class:

```javascript
class String is Sequence {
  /* existing stuff */
  
  // Compares this to another string lexicographically by codepoint.
  // Returns -1, 0 or +1 depending on whether
  // this < other, this == other or this > other respectively.
  compareTo(other) {
     if (!(other is String)) Fiber.abort("Other must be a string")
     if (this == other) return 0
     var cp1 = this.codePoints
     var cp2 = other.codePoints
     var len = (cp1.count <= cp2.count) ? cp1.count : cp2.count
     for (i in 0...len) {
        if (cp1[i] < cp2[i]) return -1
        if (cp1[i] > cp2[i]) return 1
     }
     return (cp1.count < cp2.count) ? -1 : 1
  }

  // Comparison operators.
  <  (other) { compareTo(other) <  0 }
  <= (other) { compareTo(other) <= 0 }
  >  (other) { compareTo(other) >  0 }
  >= (other) { compareTo(other) >= 0 } 
}
```

I look forward to receiving comments on this proposal and any suggestions for improving the code.

mhermier:
You can implement `compareTo` in C for better performance and to handle handle BOM prelude.

But I would go optimize even furzer, and directly use `strcmp`. Since regular UTF-8 encoding is compatible with `strcmp` (ignoring some edge cases like BOM, characters with multiple encoding, sub-optimal encoding...), it should provide a pretty reliable and fast comparison.

mhermier:
Thinking at it again we might need to use `memcmp` instead. Since `String`s known their size and since `\0` is a legal UTF-8 char, I think it is a better solution. That way, we can also compare strings holding binary data.

PureFox48:
[...]

PR #1147, opened 2023-03-04
Add String.compare

PureFox48:
The purpose of this PR is to enable comparisons between strings as discussed in #1142.

This will be particularly useful when sorting a list of strings.

The ```String.compareTo``` instance method is implemented in C for maximum speed and the comparison operators ```<, <=, >, >=```, which call it under the hood, are implemented in Wren.

Tests and documentation are included in the PR.


mhermier:
Implementation is good enough, thought the style is not really my taste.
We could squeeze a little bit more efficiency by making operators primitives, but that can wait till we can prove to be a bottleneck.

PR #1152, opened 2023-03-10
Better Random objects with default seeding

enci:
The no args constructor for Random creates an objects with the same seed within the same second, because `time()` has second resolution. Adding `clock()` to the seed removes this implicit and undocumented behavior.


iwillspeak:
Seems to bring a lot in from that fork. 

Perhaps just cherry pick the src/optional/wren_opt_random.c change? Seems like a sensible idea though.

enci:
Sorry, I forgot about opening this one and just reused the fork.

PR #1161, opened 2023-03-23
Better string interpolation

mhermier:
Currently string interpolation rely on 2 mechanisms:
- `String.+(_)` aka string concatenation.
- Creating a `List` of string and perform `List.join()` on it.

First patch optimize string concatenation, by taking advantage of string immutability. So adding an empty string can trivially optimized by returning the other value.
```
string_interpolation - wren    .......... 2.04s 0.0834 130.18% relative to baseline
string_interpolation_gc - wren .......... 2.93s 0.0724 120.93% relative to baseline
```
It is quite important since the compiler insert them when parsing interpolation strings values happen at the begin ("%(a)..."), in the middle ("...%(a)%(b)...") or at the end ("...%(a)").

The second patch eliminates spurious empty literal string concatenation, and the need of the helper `List` object. It is obsolete the first patch but it is remains an interesting change to have I think.
```
string_interpolation - wren    .......... 0.43s 0.0049 612.68% relative to baseline
string_interpolation_gc - wren .......... 1.34s 0.0149 264.72% relative to baseline
```

Second patch technically produce a little bit more code, but I think in general case interpolated values have a lot of spurious empty strings that are eliminated, reducing code size to a similar size. In addition the removal of the helper `List` shows it has a significant impact in the benchmark.

### DIFF
--- a/src/vm/wren_compiler.c
+++ b/src/vm/wren_compiler.c
@@ -2435,31 +2435,35 @@ static void literal(Compiler* compiler, bool canAssign)
 //     ["a ", b + c, " d"].join()
 static void stringInterpolation(Compiler* compiler, bool canAssign)
 {
-  // Instantiate a new list.
-  loadCoreVariable(compiler, "List");
-  callMethod(compiler, 0, "new()", 5);
-  
+  bool first = true;
   do
   {
     // The opening string part.
-    literal(compiler, false);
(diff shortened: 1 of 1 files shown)


PureFox48:
Can't say I'd ever thought much about the efficiency of string interpolation before but, if can be improved as much as that by making a few simple changes, we might as well have it.

PR #1170, opened 2023-06-24
add unary + operator

tr8dr:
Added a unary '+' operator (`+variable`) to provide symmetry with the unary '-' operator (`-variable`).   These operators are supported by most languages, so makes sense to include.   Although unary '+' is effectively a NOP, it provides clarity in coding situations where want to contrast with the negation, where one may have code applying + and - to some magnitude.

In scripting often have symmetric rules, such as:

```
static rule_type1(alpha, beta, threshold1, threshold2) {
    return (alpha > -threshold) && (beta < +threshold2)
}
static rule_type2(alpha, beta, threshold1, threshold2) {
    return (alpha < +threshold) && (beta > -threshold2)
}
```
Clarifying + and - magnitude clarifies the rules for us.

### DIFF
--- a/src/vm/wren_compiler.c
+++ b/src/vm/wren_compiler.c
@@ -1726,7 +1726,7 @@ typedef enum
   PREC_RANGE,         // .. ...
   PREC_TERM,          // + -
   PREC_FACTOR,        // * / %
-  PREC_UNARY,         // unary - ! ~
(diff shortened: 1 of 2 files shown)


mhermier:
Duplicate of #986

ruby0x1:
Thanks for the PR though @tr8dr , we always appreciate time spent

PR #1177, opened 2023-08-19
Update control-flow.markdown - for statement section

### LINKED ISSUE #1176: Should we document this aspect of for loop control variable behavior?
I was looking through the release notes for Go 1.21 when I came across [LoopVarExperiment](https://github.com/golang/go/wiki/LoopvarExperiment).

The first two examples are irrelevant to Wren as they're using parallel processing which, of course, we don't have.

However, the equivalent Wren code for the third example is:

```javascript
var print123 = Fn.new {
    var prints = []
    for (i in 1..3) {
        prints.add(Fn.new { System.print(i) })
    }
    for (print in prints) print.call()
}

print123.call()

/*
1
2
3
*/
```

which (thankfully) prints 1, 2, 3 because Wren is effectively creating a new variable in each iteration of the loop and it is that variable which is captured by the closure.

However, this behavior is not obvious and anyone who is new to Wren but has experience in other languages which suffer from this problem might think that they have to do something like this to circumvent it:

```javascript
var print123 = Fn.new {
    var prints = []
    for (i in 1..3) {
        var j = i  // create new variable
        prints.add(Fn.new { System.print(j) }) // capture new variable
    }
    for (print in prints) print.call()
}
```

I was wondering therefore whether we should document it somewhere? The most obvious place would be in the [Control Flow - For statement - section](https://wren.io/control-flow.html). Perhaps something on the following lines as a separate paragraph at the end:

"Note that a fresh variable is created for each iteration of the loop and so, if a function in the body captures that variable, its value will not be affected by subsequent iterations."

Any thoughts or suggestions before I launch off on a PR for this?

PureFox48:
Fixes #1176.

### DIFF
--- a/doc/site/control-flow.markdown
+++ b/doc/site/control-flow.markdown
@@ -174,6 +174,16 @@ A `for` loop has three components:
 3. A *body*. This is a curly block or a single statement. It gets executed once
    for each iteration of the loop.
 
+Note that a fresh variable is created for each iteration of the loop and so, if a [function](functions.html) in the body captures that variable, its value will not be affected by subsequent iterations.
+
+<pre class="snippet">
+var prints = []
+for (i in 1..3) {
+  prints.add(Fn.new { System.print(i) })
+}
+for (print in prints) print.call()  //> 1, 2, 3 not 3, 3, 3
+</pre>
+
 ## Break statements
 
 Sometimes, right in the middle of a loop body, you decide you want to bail out


RobLoach (changes requested):
Good call to mention that variable scope it kept for the function.

It may make more sense to document this in the Functions section, as this seems more like a property of [closures](https://github.com/wren-lang/wren/blob/main/doc/site/functions.markdown#closures), rather than loops and control flow.

PureFox48:
Well, the Closures section already makes clear what it means to _capture_ a variable.

This PR is peculiar to the _for_ statement where it looks like there’s only one variable for the entire loop when in fact there’s a separate instance for each iteration. Although this only really manifests itself when there’s a closure involved, I still think that the Control Flow section is the best place to put it.

PR #1187, opened 2024-09-06
Print object tostring

Stephenmccaul:


mhermier:
Hi, thanks for your contribution, but the main change is invalid.

`String` already has the to string method `toString`, so adding 'toString()' does not help and is not supported by other objects. And secondly it does remove the security against rogue objects not returning a `String` from `toString`.
The patch lacks a test case, so my best guess is that you are trying to use the API incorrectly.

For the rest of the change set, it should be split in smaller patches and proposed in a different pull request. 
I'm surprised a compiler would complains that `"%p"` would impose a `(void*)` cast, what compiler and flags are you using ?

Stephenmccaul:
I did not mean to send this.

PR #1233, opened 2026-03-27
Parse full-range hex literals with strtoull

cuiweixie:
## Summary

Hex integer literals are tokenized as `0x...` and parsed with the C library. `strtoll` only covers the signed range; `0xFFFFFFFFFFFFFFFF` (UINT64_MAX) overflows, sets `ERANGE`, and the lexer incorrectly reported "Number literal was too large".

Use `strtoull` so the full 64-bit unsigned range parses, then cast to `double`.

## Regression

`test/core/number/hex_literal_uint64_max.wren`: **fails** with `strtoll` (compile error), **passes** with `strtoull`.

## Testing

```
python3 util/test.py core/number/hex_literal_uint64_max
python3 util/test.py core/number
```

### DIFF
--- a/src/vm/wren_compiler.c
+++ b/src/vm/wren_compiler.c
@@ -750,7 +750,7 @@ static void makeNumber(Parser* parser, bool isHex)
 
   if (isHex)
   {
-    parser->next.value = NUM_VAL((double)strtoll(parser->tokenStart, NULL, 16));
+    parser->next.value = NUM_VAL((double)strtoull(parser->tokenStart, NULL, 16));
   }
   else
   {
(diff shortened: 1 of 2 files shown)


CrazyInfin8:
As an alternative, in PR #984 I've attempted to create an entirely custom number parsing method. Which, should handle this case, but is certainly a much more aggressive change to the language and handles this a completely different way.

It'd consume as many digits that'd fit in the mantissa of a double (about what would fit in 52 bits), consume the rest ignoring the value but incrementing an exponent, then multiplies that mantissa value (as a long double) by the power of the base raised to the exponent.

Which should also prevent the overflow as we only parse enough to fit inside of about 52-bits of a 64-bit integer.

